PR #13, opened 2017-11-02
Add the `verify` command for debugging purpose

mumoshu:
This change adds the `verify` command to verify a token and inspect what identity is included in it to help deciding what was wrong in your setup - an assumed role or iam user used for generating token or anything else.

Usage:

```
$ kubernetes-aws-authenticator verify --cluster-id foo --token=$(./kubernetes-aws-authenticator-osx token --cluster-id foo)
&{ARN:arn:aws:sts::1111111111111:assumed-role/developer/mumoshu@example.com CanonicalARN:arn:aws:iam::11111111111:role/developer AccountID:1111111111111 UserID:AROAIUIYAJDPNR756ILNQ SessionName:mumoshu@example.com}
```

### DIFF
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2017 by the contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


mattmoyer:
Looks great. Thanks!

PR #14, opened 2017-11-02
Add support for federated users and roles

### LINKED ISSUE #12: Idea: Dynamically mapping federated users
@mattmoyer Hi, thanks for sharing the great project 👍 

Please let me share the idea came up to me today - I'd like SAML federated users to be mapped to correspondingly named K8S users created/updated on demand by kubernetes-aws-authenticator so that I could be released from statically mapping many IAM roles/users to k8s users.

I imagine that it would work similar to `mapEC2InstanceRoles` in regard to that [it also maps an EC2 instance id obtained by calling STS' `GetCallerIdentity` to the name of an automatically created k8s service account](https://github.com/heptiolabs/kubernetes-aws-authenticator/blob/ce71e0c60b43d9fec185b6a17cefe1b03941aefe/pkg/server/server.go#L190-L192).

Revisiting what we have so far - calling [GetCallerIdentity](http://docs.aws.amazon.com/cli/latest/reference/sts/get-caller-identity.html) within a session established via AssumeRoleWithSAML gives us:

```json
{
    "Account": "${account_id}",
    "UserId": "${unique_id_of_assumed_iam_role}:${caller_specified_role_name}",
    "Arn": "arn:aws:sts::${aws_account_id}:assumed-role/${assumed_iam_role_name}/${caller_specified_role_name}"
}
```

where:
- `unique_id_of_assumed_iam_role` is something like `AROAIUI*snip*R756ILNQ`, neither a role name nor a role arn.
- `caller_specified_role_name` can be anything specified by the ID provider but could be my e-mail address `mumoshu@example.com`

For example, given a `caller_specified_role_name` of `mumoshu@example.com`, I'd like kubernetes-aws-authenticator to automatically create or update the k8s service account `mumoshu` or `auto-mumoshu`, whichever doesn't collide with k8s users not managed by kubernetes-aws-authenticator.

I'd expect config to look like:

```yaml
mapFederatedUsers:
  - roleARN: arn:aws:iam::000000000000:role/Developer
    usernameFormat: adm-{{CallerSpecifiedRoleName}}
    groups:
    - developers
  - roleARN: arn:aws:iam::000000000000:role/Admin
    usernameFormat: dev-{{CallerSpecifiedRoleName}}
    groups:
    - admins
```

WDYT?

mumoshu:
Resolves #12

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -49,7 +49,7 @@ type handler struct {
 	http.ServeMux
 	clusterID        string
 	lowercaseRoleMap map[string]config.RoleMapping
-	lowercaseUserMap map[string]config.StaticUserMapping
+	lowercaseUserMap map[string]config.UserMapping
 }
 
 // New creates a new server from a config
@@ -67,14 +67,14 @@ func (c *Server) Run() {
 			"username":       mapping.Username,
 			"usernameFormat": mapping.UsernameFormat,
 			"groups":         mapping.Groups,
-		}).Infof("statically mapping IAM role")
+		}).Infof("mapping IAM role")
 	}
-	for _, mapping := range c.StaticUserMappings {
+	for _, mapping := range c.UserMappings {
 		logrus.WithFields(logrus.Fields{
 			"user":     mapping.UserARN,
 			"username": mapping.Username,
 			"groups":   mapping.Groups,
-		}).Infof("statically mapping IAM user")
+		}).Infof("mapping IAM user")
 	}
 
 	// we always listen on localhost (and run with host networking)
@@ -117,12 +117,12 @@ func (c *Server) getHandler() *handler {
 	h := &handler{
 		clusterID:        c.ClusterID,
 		lowercaseRoleMap: make(map[string]config.RoleMapping),
-		lowercaseUserMap: make(map[string]config.StaticUserMapping),
+		lowercaseUserMap: make(map[string]config.UserMapping),
 	}
 	for _, m := range c.RoleMappings {
 		h.lowercaseRoleMap[strings.ToLower(m.RoleARN)] = m
 	}
-	for _, m := range c.StaticUserMappings {
+	for _, m := range c.UserMappings {
 		h.lowercaseUserMap[strings.ToLower(m.UserARN)] = m
 	}
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -47,10 +47,9 @@ var tokenReviewDenyJSON = func() []byte {
 // server state (internal)
 type handler struct {
 	http.ServeMux
-	clusterID           string
-	lowercaseRoleMap    map[string]config.StaticRoleMapping
-	lowercaseEC2RoleMap map[string]config.EC2InstanceRoleMapping
-	lowercaseUserMap    map[string]config.StaticUserMapping
+	clusterID        string
+	lowercaseRoleMap map[string]config.RoleMapping
+	lowercaseUserMap map[string]config.StaticUserMapping
 }
 
 // New creates a new server from a config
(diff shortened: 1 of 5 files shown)


mattmoyer:
Thanks @mumoshu, this looks great.

I'm +1 on combining the EC2 and federated cases. Maybe we could even combine `mapRole` with `mapAssumedRole` also and just have a generic `{SessionName}` placeholder?

In the EC2 case you could trust it to be an instance ID, in the federated case you could trust it to be the external identity, and otherwise (for plain `sts:AssumeRole`) you'd be free to use it for informational purposes if you wanted.

pkg/server/server.go
```diff
@@ -187,9 +187,12 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
 	var username string
 	var groups []string
 	if ec2RoleMapping, exists := h.lowercaseEC2RoleMap[arnLower]; exists {
+		// username must be a DNS-1123 hostname matches the regex "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
+		specifiedRole := strings.Replace(identity.SessionName, "@", "-", -1)
 		username = ec2RoleMapping.UsernameFormat
 		username = strings.Replace(username, "{{AccountID}}", identity.AccountID, -1)
 		username = strings.Replace(username, "{{InstanceID}}", identity.SessionName, -1)
+		username = strings.Replace(username, "{{CallerSpecifiedRoleName}}", specifiedRole, -1)
```
mumoshu: @mattmoyer Thanks for the review! I'm ok with merging the `InstanceID` and `CallerSpecifiedRoleName` into just `SessionName` as you've suggested.
However then, would you mind if I blindly replace every occurrence of `@` in a session name to `-` like I did in the line 191? (Would be ok as it won't affect instance ids - just confirming assuming it would look a bit dirty?)

mumoshu:
> Maybe we could even combine mapRole with mapAssumedRole

I guess we may want `mapRoles` for static mapping whereas `mapAssumedRoles` is for dynamic mapping? 🤔 
Would it be a matter of naming? Then, I'd rather suggest renaming `mapAssumedRoles` to `mapRolesDynamically`(hmm) while leaving `mapRoles` implying `mapRolesStatically`.

mumoshu:
Ah, did you mean to merge the two like this?

```yaml
  # Each mapRoles entry maps an IAM role to username(s) and group(s)
  mapRoles:
  # e.g., map arn:aws:iam::000000000000:role/KubernetesAdmin to a cluster admin
  - roleARN: arn:aws:iam::000000000000:role/KubernetesAdmin
    username: kubernetes-admin
    groups:
    - system:masters
  # e.g., map EC2 instances in my "KubernetesNode" role to users like
  # "aws:000000000000:instance:i-0123456789abcdef0"
  - roleARN: arn:aws:iam::000000000000:role/KubernetesNode
    usernameFormat: aws:{{AccountID}}:instance:{{InstanceID}}
    groups:
    - system:bootstrappers
    - aws:instances
  # e.g., map federated users in my "KubernetesAdmin" role to users like
  # "admin:alice-example.com"
  - roleARN: arn:aws:iam::000000000000:role/KubernetesAdmin
    usernameFormat: admin:{{CallerSpecifiedRoleName}}
    groups:
    - system:masters
```

mattmoyer:
> Ah, did you mean to merge the two like this?

Yeah exactly. You could still have a static mapping if you didn't use any of the template parameters in your format.

I can take a quick shot at this. One sec...

mumoshu:
Oh, I'm afraid if we had done the same things at different times!
Anyway, I've just pushed additional commits about that.
Would it be something you had intended?

PR #20, opened 2017-11-27
Fix `go get` command to use new cmd subdir.

mattmoyer:
Fallout from https://github.com/heptio/authenticator/pull/19.


Pea1594 (approved):

Pea1594:
**n localhost:5000. L**

PR #27, opened 2017-12-08
[feature] Run the authenticator with an unprivileged user

stefancocora:
Use alpine as the base image because it brings unprivileged user and ca
certificate tooling.
Use the alpine upstream updated and maintained ca-certificates rather than
pulling ca-certificates.crt out of an alpine:latest to bake into this image.
Software running in containers should default to run as unprivileged user rather
than running as the root user inside the container.

Fixes Github issue https://github.com/heptio/authenticator/issues/26


mattmoyer:
Thanks for this contribution! I tweaked this a little and included it in https://github.com/heptio/authenticator/pull/41.

Rather than start using Alpine for the runtime image, I used an Alpine build stage to set up the unprivileged user, then copied the `/etc/passwd` and CA certificates into a final scratch image. Do you think that's an okay approach?

stefancocora:
Everything is a compromise/trade-off ...
To answer your question I'd have to know the advantages and the disadvantages of the build-on-alpine-then-copy-to-scratch decision.

**Advantages ?**
I'm unsure what the additional benefits the scratch image bring to the table, maybe you can enumerate them.
The alpine image is small enough as it is.

**Disadvantages**
- keep in mind the maintenance overhead of having to deal with 2 layers of abstraction before setting up your `rootfs` filesystem for the `authenticator` , debugging issues when something goes wrong
- the lack of debug-ability when the `authenticator` `rootfs` doesn't have a shell built-in

I'm not sure what is the best approach, personally I'd opt for using just the alpine image due to the simplicity and reduced layers of abstraction to build the `rootfs` needed by `authenticator`

What do you think about this approach ?

mattlandis:
@mattmoyer Since this is in #41 this can be closed.

PR #33, opened 2017-12-11
Build container for binary build

RaphaelHerd:
I added a simple script to easy build the heptio binary. It's useful if you don't want to wast you laptop. In addition, it prevents versioning conflicts with existing developments.


nckturner:
I think we'd like to stick with the Makefile and goreleaser.

PR #41, opened 2018-01-12
Start using goreleaser, run as non-root user

mattmoyer:
This moves the build process over to https://github.com/goreleaser/goreleaser, which is a small tool for building and releasing Go binaries.

It also switches the Docker image build to make sure the `heptio-authenticator-aws` process is running as a non-root user by default.

Fixes https://github.com/heptio/authenticator/issues/39
Fixes https://github.com/heptio/authenticator/issues/28
Fixes https://github.com/heptio/authenticator/issues/26


Makefile
```diff
@@ -1,26 +1,17 @@
...
-VERSION ?= v0.1.0
+GORELEASER := $(shell command -v goreleaser 2> /dev/null)
 
-.PHONY: build push build-container
+.PHONY: build test format
 
-build: build-container heptio-authenticator-aws-osx
+build:
+ifndef GORELEASER
+	$(error "goreleaser not found (`go get-u -v github.com/goreleaser/goreleaser` to fix)")
```
nckturner: Should be `go get -u` ?

mattmoyer: Ah, thanks. Fixed.

PR #42, opened 2018-01-16
Support getting tokens from an IAM Role requiring MFA

gwkunze:
Adding the AssumeRoleTokenProvider allows accounts with MFA to also request a token. Instead of getting an error:
```
heptio-authenticator-aws token -i my-cluster -r arn:aws:iam::123456789012:role/KubernetesAdmin
could not get token: could not create session: AssumeRoleTokenProviderNotSetError: assume role with MFA enabled, but AssumeRoleTokenProvider session option not set.
```

The app asks the user for an mfa code:
```
heptio-authenticator-aws token -i my-cluster -r arn:aws:iam::123456789012:role/KubernetesAdmin
Assume Role MFA token code: 123456
k8s-aws-v1.abcdefg......
```


nckturner:
Great! LGTM

gwkunze:
I should note that since `stscreds.StdinTokenProvider` prompts on stdout and reads from stdin the `kubectl --kubeconfig /path/to/kubeconfig --token "$(heptio-authenticator-aws token -i CLUSTER_ID -r ROLE_ARN)" [...]`-command as described doesn't work as the prompt will be part of the `--token` argument. Switching to custom prompt written to stderr would probably resolve this.

Another issue is that this will mean that the user will need to enter an MFA code for every kubectl command they want to execute. Including having to wait 30 seconds between commands for a new code to be generated as you can't use the same one twice. This is of course not the best user experience. It might be better if there was some caching of the pre-signed url on the client side machine, only generating a new token when the url is about to expire...

nckturner:
@gwkunze Noted, the implementation of this proposal https://github.com/kubernetes/features/issues/541 which is how heptio authenticator will be exec'd by kubectl, will have to make sure the exec'd process properly inherits stdin from kubectl, so I don't think we should switch to a custom prompt written to stderr.  Its a similar situation for binaries that have a `login` functionality.  

> We should take care that interactive stderr and stdin are correctly inherited by the sub-process to enable this kind of interaction. The plugin will still be responsible for prompting the user, receiving user feedback, and timeouts.

Caching is also be handled by that proposal, but perhaps we can wait until the proposal is finalized before moving forward with this.  

> Plugins will be called on client-go initialization, and again when the API server returns a 401 HTTP status code indicating expired credentials. Plugins can indicate their credentials explicit expiry using the Expiry field on the returned ExecCredentials object, otherwise credentials will be cached hroughout the lifetime of a program.

mattmoyer:
@gwkunze thanks for this! We run with MFA-restricted IAM roles here too so this is very welcome.

I'd still like to merge this, but I agree with @nckturner re:
> wait until the proposal is finalized before moving forward with this.

If it's helpful, I've been using https://github.com/99designs/aws-vault to do the MFA dance.

mattmoyer:
If you don't mind, do a `git --amend --signoff` to add a `Signed-Off-By:` line and make our DCO bot happy.

gwkunze:
Apologies for the late update.

mumoshu:
@mattlandis @mattmoyer @nckturner Hi, this is a friendly reminder towards merging this great work.

Given the commit is now signed-off and it is already LGTMed, probably you want to merge this?

mattlandis:
I want to check that this works properly with the new 1.10 auth provider changes in client-go before we merge this change in.

nckturner:
So, turns out this doesn't quite work with the 1.10 auth provider because [it prints the prompt to enter your MFA code on stdout](https://github.com/aws/aws-sdk-go/blob/fe4c33d5e201f595b450d08d9d7e36fedb2cbcf8/aws/credentials/stscreds/assume_role_provider.go#L105), which in the 1.10 exec based system is reserved for the formatted json token output.  

I propose we merge this change, tag and branch for kubernetes 1.9, then merge 1.10 support and create an issue in this repository for the broken 1.10 MFA functionality.  We can then look into fixing that with a custom [AssumeRoleProvider](https://github.com/aws/aws-sdk-go/blob/fe4c33d5e201f595b450d08d9d7e36fedb2cbcf8/aws/credentials/stscreds/assume_role_provider.go#L133) or something similar.

nckturner:
And prompting every kubectl invocation would be annoying, you're right, we should look into caching client side beyond the lifetime of the authenticator.

mattmoyer:
It looks like providing a custom TokenProvider function for stderr vs. stdout would be pretty easy:
https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/stscreds/assume_role_provider.go#L103-L109

It looks like if we just replace the `fmt.Printf("...")` with `fmt.Fprint(os.Stderr, "...")` we'll be good to go:

```go
func StdinStderrTokenProvider() (string, error) {
	var v string
	fmt.Fprint(os.Stderr, "Assume Role MFA token code: ")
	_, err := fmt.Scanln(&v)
	return v, err
}
```

nckturner:
Great, let's merge this PR and get a fix in.

PR #43, opened 2018-01-16
Record and export metrics with prometheus.

mattlandis:
Add metrics for the /authenticate call for latency, number of requests,
malformed requests, invalid tokens and unknown users.

Expose the metrics via the /metrics path.

Note that this currently uses prometheus/go-client at 0.9.0-pre1.  It provides some nice additional features that clean up the latency metrics.  If we want to go back to 0.8.x we can.


pkg/server/server.go
```diff
@@ -126,10 +145,56 @@ func (c *Server) getHandler() *handler {
 	}
 
 	h.HandleFunc("/authenticate", h.authenticateEndpoint)
+	h.Handle("/metrics", promhttp.Handler())
 	return h
 }
 
+func createMetrics() metrics {
+	m := metrics{
+		latency: prometheus.NewHistogram(prometheus.HistogramOpts{
```
jrnt30: Nothing of serious substance, but had you considered adding in some of these additional attributes as a label on the histogram instead of discrete metrics?  Correlating the latent requests with a failure vs. a success would be potentially useful in debugging issues.

Also, does the fact that the histogram provides the `_count` and `_sum` metrics make a few of these redundant?

mattlandis: I'll give it a try.  I was looking at adding a second latency to capture the sts call latency so it could be separated from the others.  Under normal load while you are not receiving malformed requests it wouldn't cause an issue, but a bunch of malformed requests that terminate quickly could mask the actual latency of normal calls.

jrnt30: Gotcha, I think those could be differentiated pretty cleanly as different labels instead of different metrics, but that is more of a design decision on how you write your alerting/dash-boarding logic.

jrnt30: Gotcha, I think those could be differentiated pretty cleanly as different labels instead of different metrics, but that is more of a design decision on how you write your alerting/dash-boarding logic.

mattlandis:
Rewrote the metrics using just one histogram for latency.

mattlandis:
@mattmoyer Bump

PR #44, opened 2018-01-22
Add tests target to the Makefile

mattlandis:
Tests are not currently run as part of the make.

Should be noted that -race takes the test time from ~0.1 seconds to ~3 seconds.  I am open to removing it if we feel running in all cases is too long and we can make a release target that runs it.

Signed-off-by: Matt Landis <matlan@amazon.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,21 @@
 default: build
 
 CMD_IMPORT ?= github.com/heptio/authenticator/cmd/heptio-authenticator-aws
+GITHUB_REPO ?= github.com/heptio/authenticator
 REPO ?= gcr.io/heptio-images/authenticator
 VERSION ?= v0.1.0
 
-.PHONY: build push build-container
+.PHONY: test build push build-container
 
-build: build-container heptio-authenticator-aws-osx
+build: test build-container heptio-authenticator-aws-osx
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,17 +1,21 @@
...
 CMD_IMPORT ?= github.com/heptio/authenticator/cmd/heptio-authenticator-aws
+GITHUB_REPO ?= github.com/heptio/authenticator
 REPO ?= gcr.io/heptio-images/authenticator
 VERSION ?= v0.1.0
 
-.PHONY: build push build-container
+.PHONY: test build push build-container
 
-build: build-container heptio-authenticator-aws-osx
+build: test build-container heptio-authenticator-aws-osx
```
nckturner: `test` could probably just be a dependency to `heptio-authenticator-aws-osx`.

mattlandis: If it is only a target for osx then it will not run when build-container is run explicitly.

mattlandis:
@mattmoyer Bump.

mattmoyer:
I think this is covered by https://github.com/heptio/authenticator/pull/41, which I just bumped to include `-race` and `-cover`. Take a look over there and see if that works for you?

mattlandis:
Yea.  That works for me.

Issue #31, opened 2017-12-11
Race Condition in K8S Start


RaphaelHerd:
Hi,

I successfully tested the auth plugin follwing your instructions. Now I would like to automate the deployment. What I found out is, that currently I have a race condition between the DaemonSet and KubeApiServer.

While KubeApiServer is starting (after a new master instance is launched by the AWS Autoscaling Group), the KubeAPIServer is not able to find the kubeconfig.yaml. This is because it's not present on the machine. It's not because the DS is nor started before and therefore heptio auth was not able to generate the config file on the local disc of this machine.

Do you have any ideas how to solve this issue? Using K8s 1.7.10 and current version of Kops for installing.

Thx
Raphael

bbck:
I just ran into this and got around it using `fileAssets` in kops to place a dummy file. You also need to edit the example deployment to drop the webhook configuration into `/srv/kubernetes` since it's mounted in the apiserver pod. 

```  yaml
fileAssets:
  - content: |
      clusters:
        - name: heptio-authenticator-aws
          cluster:
            certificate-authority: /srv/kubernetes/ca.crt
            server: https://127.0.0.1:21362/authenticate
      users:
        - name: apiserver
      current-context: webhook
      contexts:
      - name: webhook
        context:
          cluster: heptio-authenticator-aws
          user: apiserver
    name: heptio-authenticator-aws
    path: /srv/kubernetes/webhook-authentication.yaml
    roles:
    - Master
```

RaphaelHerd:
Funny to hear that. I did exactly the same :-). But instead of file Assets I'm using a hook with an execContainer and mounted the path in it.

`
  - execContainer:
      command:
      - sh
      - -c
      - "apk -Uuv add groff less python py-pip && pip install awscli && mkdir -p /rootfs/srv/kubernetes/heptio-authenticator-aws/ && aws s3 cp 's3://..../kubeconfig.yaml' '/rootfs/srv/kubernetes/heptio-authenticator-aws/kubeconfig.yaml' --region eu-central-1"
      image: alpine
`

bbck:
This was using kops 1.8.

mirthy:
@bbck Did you have any issues with certificates?  I'm running into an issue where the apiserver doesn't like the self-signed cert that the authenticator is using.

mattmoyer:
[...]

Issue #25, opened 2017-12-05
Bad certificate error


hugespoon:
**Kubernetes Version**: 1.8.4

I'm attempting to give this a try in a testing cluster I've spun up with `kops 1.8.0`. I think I'm quite close, but it looks like I've got a certificate problem somewhere. When attempting to do the following, I get a few errors:

`
kubectl --kubeconfig=/path/to/my/kubeconfig --token="$(heptio-authenticator-aws token -i mycluster.local)" get nodes
`

The first is immediate on my local workstation:
`error: You must be logged in to the server (Unauthorized)`

The second is from the container itself:
`time="2017-12-05T20:14:47Z" level=info msg="http: TLS handshake error from 127.0.0.1:41980: remote error: tls: bad certificate" http=error`

I'm assuming I have the wrong certificate configured somewhere, but from the docs it is not clear where that might be. Does the authenticator need to use the cluster certificates (in this case generated by kops) or are the certs it generates on its own correct?

mirthy:
I think it's the API server not liking the fact that the heptio service is using a self-signed cert.

From the API server logs:
E0110 15:57:04.102999       6 authentication.go:64] Unable to authenticate the request due to an error: [invalid bearer token, [invalid bearer token, Post https://127.0.0.1:21362/authenticate: x509: certificate signed by unknown authority]]

From the service:
time="2018-01-10T15:57:04Z" level=info msg="http: TLS handshake error from 127.0.0.1:56138: remote error: tls: bad certificate" http=error

baniol:
@hugespoon 
you're not assuming role (with `-r`) and trying to authenticate as IAM user, which is probably not mapped in ConfigMap (https://github.com/heptio/authenticator/blob/master/example.yaml#L68)

I was able to fix it after examining the logs of the heptio-authenticator pod, where I spotted:
```
time="2018-01-20T22:44:05Z" level=warning msg="access denied because ARN is not mapped" arn="arn:aws:iam::000000000000:user/my-user" client="127.0.0.1:38836" method=POST path=/authenticate
```

KaustubhKhati:
Any Update on the issue?
We too are facing the same problem only thing that is different for us is that after all the kops rolling update if we restart the docker container named

KaustubhKhati:
I have seen that after doing kops rolling update restarting the docker pod 'k8s_kube-apiserver-*' in Master worked for me rather than kubelet service. And we could automate this by adding a file asset 

[...]

PR #45, opened 2018-01-26
Add documentation on using Authenticator with Kops

rifelpet:
This addresses a comment in #31.

I have successfully added Authenticator with kops using hooks and a systemd service. I wasn't sure where you want this, If it should go in its own docs file I'm happy to move it there.

I'm also not positive that this is the "best" approach but it definitely works 😄  `fileAssets` or `execContainer` are also viable solutions as discussed in the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,11 @@ Perform the following steps to setup Authenticator on a Kops cluster:
          ExecStart=/bin/mkdir -p /srv/kubernetes/heptio-authenticator-aws
          ExecStart=/usr/local/bin/aws s3 cp --recursive s3://KOPS_STATE_STORE/CLUSTER_NAME/addons/authenticator /srv/kubernetes/heptio-authenticator-aws/
    ```
+  If using a non-default AMI that does not have the AWS CLI, replace the second ExecStart statement with
+
+  ```
+  ExecStart=/usr/bin/docker run --net=host --rm -v /srv/kubernetes/heptio-authenticator-aws:/srv/kubernetes/heptio-authenticator-aws quay.io/coreos/awscli@sha256:7b893bfb22ac582587798b011024f40871cd7424b9026595fd99c2b69492791d aws s3 cp --recursive s3://KOPS_STATE_STORE/CLUSTER_NAME/addons/authenticator /srv/kubernetes/heptio-authenticator-aws/
+  ```
 3. Apply the changes with `kops update cluster ${CLUSTER_NAME}`.
    If the cluster already exists, roll the cluster with `kops rolling-update cluster ${CLUSTER_NAME}` in order to recreate the master nodes.
 4. Update the Authenticator DaemonSet's state and output volumes to both use `/srv/kubernetes/heptio-authenticator-aws/` for their `hostPath`s.
--- a/README.md
+++ b/README.md
@@ -105,6 +105,48 @@ The token is valid for 15 minutes (the shortest value AWS permits) and can be re
 You can also omit `-r ROLE_ARN` to sign the token with your existing credentials without assuming a dedicated role.
 This is useful if you want to authenticate as an IAM user directly or if you want to authenticate using an EC2 instance role or a federated role.
 
+## Kops Usage
+Clusters managed by [Kops](https://github.com/kubernetes/kops) can be configured to use Authenticator.
+Both single and HA master cluster configurations are supported.
+Perform the following steps to setup Authenticator on a Kops cluster:
+1. Pre-generate the certificate, key, and kubeconfig and upload them to the kops state store.
+   ```
+   heptio-authenticator-aws init -i $CLUSTER_NAME
+   aws s3 cp cert.pem ${KOPS_STATE_STORE}/${CLUSTER_NAME}/addons/authenticator/cert.pem;
+   aws s3 cp key.pem ${KOPS_STATE_STORE}/${CLUSTER_NAME}/addons/authenticator/key.pem;
+   aws s3 cp heptio-authenticator-aws.kubeconfig ${KOPS_STATE_STORE}/${CLUSTER_NAME}/addons/authenticator/kubeconfig.yaml;
+   ```
+2. Add the following sections to the cluster spec, either using `kops edit cluster ${CLUSTER_NAME}` or editing the manifest yaml file.
+   Be sure to replace `KOPS_STATE_STORE` and `CLUSTER_NAME` with their appropriate values since those environment variables are not available at runtime.
+   This downloads the files from the state store on masters to a directory that is volume mounted by kube-apiserver.
+   Kops does not support adding additional volumes to kube-apiserver so we must reuse the existing `/srv/kubernetes` hostPath volume.
+   ```
+   apiVersion: kops/v1alpha2
+   kind: Cluster
+   spec:
+     kubeAPIServer:
+       authenticationTokenWebhookConfigFile: /srv/kubernetes/heptio-authenticator-aws/kubeconfig.yaml
+     hooks:
+     - name: kops-hook-authenticator-config.service
+       before:
+         - kubelet.service
+       roles: [Master]
+       manifest: |
+         [Unit]
+         Description=Download Heptio AWS Authenticator configs from S3
+         [Service]
+         Type=oneshot
+         ExecStart=/bin/mkdir -p /srv/kubernetes/heptio-authenticator-aws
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,6 +105,45 @@ The token is valid for 15 minutes (the shortest value AWS permits) and can be re
...
+       before:
+         - kubelet.service
+       roles: [Master]
+       manifest: |
+         [Unit]
+         Description=Download Heptio AWS Authenticator configs from S3
+         [Service]
+         Type=oneshot
+         ExecStart=/bin/mkdir -p /srv/kubernetes/heptio-authenticator-aws
+         ExecStart=/usr/local/bin/aws s3 cp --recursive s3://KOPS_STATE_STORE/CLUSTER_NAME/addons/authenticator /srv/kubernetes/heptio-authenticator-aws/
```
phillipj: I had to add an explicit `--region` argument to this command.

phillipj:
Really appreciated this example, helped me a lot to get the authenticator up-n-running in our clusters with kops 👍

kdillane:
How did you boostrap the Authenticator DaemonSet and ConfigMap?  When creating a cluster using kops and using the authenticationTokenWebhookConfigFile kubeApiServer argument, there's no clear path to creating these resources.

phillipj:
I started off with the [example.yaml](https://github.com/heptio/authenticator/blob/master/example.yaml) referenced in the [Run the server section](https://github.com/heptio/authenticator#2-run-the-server) of the `README.md`. Did some changes to the ConfigMap by trial-n-error.

Important to remember step 4 of the `README.md` changes proposed in this PR: 

> Update the Authenticator DaemonSet's state and output volumes to both use `/srv/kubernetes/heptio-authenticator-aws/` for their `hostPath`s.

HagaiBarel:
Doesn't work while using coreos container linux. Turns out the ami doesn't have aws cli installed and therefore the `aws s3 cp` command fails

kdillane:
Did you script the cluster creation end to end?  How did you create the ConfigMap and DaemonSet if the api server is protected by the Authenticator?

rifelpet:
@kdillane other authentication methods still work, so having the certificate data in your kubeconfig works fine without any aws credentials. We deploy in a CI pipeline:
```
kops export kubecfg ${CLUSTER_NAME}
kubectl apply -f authenticator.yml
```

kdillane:
Makes sense. Okay, I must have been running into a separate issue from authentication not working. I’ll dig in. Thanks for the feedback!

sp-joseluis-ledesma:
I've used a slightly different approch, In the kops hook:
```
  - manifest: |
      Type=oneshot
      ExecStart=/bin/mkdir -p /etc/kubernetes/manifests /etc/heptio-authenticator-aws
      ExecStart=/bin/sh -c '/bin/echo <BASE64_CONFIG.YAML> |/usr/bin/base64 -d >/etc/heptio-authenticator-aws/config.yaml'
      ExecStart=/bin/sh -c '/bin/echo  <BASE64_POD.YAMLK>|/usr/bin/base64 -d >/etc/kubernetes/manifests/heptio-authenticator-aws.manifest'
      ExecStop=/bin/rm -f /etc/kubernetes/manifests/heptio-authenticator-aws.manifest
      RemainAfterExit=true
    name: heptio-authenticator-aws.service
    requires:
    - kubelet.service
    - protokube.service
    roles:
    - Master
```
where POD.YAML is a modified version of the original DaemonSet:
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    scheduler.alpha.kubernetes.io/critical-pod: ""
  creationTimestamp: null
  labels:
    k8s-app: heptio-authenticator-aws
  name: heptio-authenticator-aws
  namespace: kube-system
spec:
  # run on the host network (don't depend on CNI)
  hostNetwork: true
  containers:
  - name: heptio-authenticator-aws
    image: gcr.io/heptio-images/authenticator:v0.1.0
    args:
    - server
    - --config=/etc/heptio-authenticator-aws/config.yaml
    - --state-dir=/var/heptio-authenticator-aws
    - --generate-kubeconfig=/srv/kubernetes/heptio-authenticator-aws/kubeconfig.yaml
    resources:
      requests:
        memory: 20Mi
        cpu: 10m
      limits:
        memory: 20Mi
        cpu: 100m

    volumeMounts:
    - name: config
      mountPath: /etc/heptio-authenticator-aws/
    - name: state
      mountPath: /var/heptio-authenticator-aws/
    - name: output
      mountPath: /srv/kubernetes/heptio-authenticator-aws/

  volumes:
  - name: config
    hostPath:
      path: /etc/heptio-authenticator-aws
  - name: output
    hostPath:
      path: /srv/kubernetes/heptio-authenticator-aws/
  - name: state
    hostPath:
      path: /var/heptio-authenticator-aws/
```

This would work for cases like using CoreOS AMI (@HagaiBarel), and allows using kops rolling-update cluster to update the config.yaml (without being necessary to force it)

README.md
willejs: I am using 
```
ExecStart=/usr/bin/docker run --net=host --rm -v /srv/kubernetes/heptio-authenticator-aws:/srv/kubernetes/heptio-authenticator-aws quay.io/coreos/awscli@sha256:7b893bfb22ac582587798b011024f40871cd7424b9026595fd99c2b69492791d aws s3 cp --recursive {{.config_base}}/addons/authenticator /srv/kubernetes/heptio-authenticator-aws/
``` 
instead on core os

README.md
```diff
@@ -105,6 +105,45 @@ The token is valid for 15 minutes (the shortest value AWS permits) and can be re
...
+   ```
+3. Apply the changes with `kops update cluster ${CLUSTER_NAME}`.
+   If the cluster already exists, roll the cluster with `kops rolling-update cluster ${CLUSTER_NAME}` in order to recreate the master nodes.
+4. Update the Authenticator DaemonSet's state and output volumes to both use `/srv/kubernetes/heptio-authenticator-aws/` for their `hostPath`s.
```
christopherhein: Probably want a 5th step or an addition to this that says to `apply` the `ConfigMap` and `DaemonSet` to be pedantic.

README.md
christopherhein: I like the addition of the MasterMind templating in this, might be nice to pull that into the readme.

christopherhein:
Another note for this as of right now is if you use the new `ExecCredential` configuration in your `kubecfg` it will get mangled when you create or delete a cluster using `kops`

rifelpet:
@christopherhein Addressed both of your comments

christopherhein:
@rifelpet Can you do one last thing, and add a note about using docker for non-default AMI deploys of `kops` like @willejs said…

```
ExecStart=/usr/bin/docker run --net=host --rm -v /srv/kubernetes/heptio-authenticator-aws:/srv/kubernetes/heptio-authenticator-aws quay.io/coreos/awscli@sha256:7b893bfb22ac582587798b011024f40871cd7424b9026595fd99c2b69492791d aws s3 cp --recursive {{.config_base}}/addons/authenticator /srv/kubernetes/heptio-authenticator-aws/
```

christopherhein:
Thanks @rifelpet, @nckturner lgtm! :shipit:

Raffo:
Been using the same suggestions, works for me... more or less. Unfortunately, I'm hitting #25 which makes the setup definitely not perfect, so a workaround should be mentioned in this guide or just integrated.

christopherhein:
I've ran through this a number of more times. I would say we're good to go and if we need to make updates after when @Raffo gets a chance to retest that shouldn't be that big of a deal. Also we have this live which has been working well - https://aws.amazon.com/blogs/opensource/deploying-heptio-authenticator-kops/

Raffo:
I agree, when I'll have the change to test it again I will definitely open an issue if I get another issue. Thanks for the great work!

christopherhein:
@nckturner can we get this merged?

StevenACoffman:
@rifelpet Do you have suggestions for how to secure the dashboard in kops using heptio-authenticator? `heptio-authenticator-aws  token -i $cluster -r $arn | jq -r '.status.token'` maybe?

christopherhein:
@StevenACoffman check - https://github.com/heptio/authenticator/issues/84

j00p34:
For anyone coming here from the AWS page, like me: The main readme https://github.com/kubernetes-sigs/aws-iam-authenticator has more up to date info.

PR #50, opened 2018-02-05
Automap ARNs to user/role for allowed accounts.

Sushant:
Signed-off-by: Sushant Bhadkamkar <sushant@lyft.com>

- Extend config to list AWS accounts from which users/roles can be mapped automatically to user/role ARNs. e.g.
```
  mapAccounts:
  - "123456789012"
  - "456789012345"
```


pkg/config/types.go
```diff
@@ -83,4 +83,8 @@ type Config struct {
 	// UserMappings is a list of mappings from AWS IAM User to
 	// Kubernetes username + groups.
 	UserMappings []UserMapping
+
+	// AWS Accounts that are allowed without an explicit user/role mapping
```
mattlandis: We need to figure out what a good name for this is.  Names we have discussed internally include `AWSAccounts`, `MappedAccounts`, `AWSAccountsToMapToUser`, `AutoMappedAWSAccounts`. 

I open to suggestions on this.  I've been having trouble trying to find something that conveys the meaning and is concise at the same time.

Comment should start with name of struct member `MappedAccounts`

pkg/server/server.go
```diff
@@ -224,6 +233,9 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
 	} else if userMapping, exists := h.lowercaseUserMap[arnLower]; exists {
 		username = userMapping.Username
 		groups = userMapping.Groups
+	} else if _, exists := h.accountMap[identity.AccountID]; exists {
```
mattlandis: What do we want to do for the use case where the AccountID is mapped and the AWS User/Group is mapped to a list of groups but no username?  Should the username be required if there is a Role/User mapping?  Should we add an automapped username if it is not provided and automap accounts is present?

pkg/server/server.go
```diff
@@ -224,6 +233,9 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+		groups = []string{}
+		username = identity.Name
```
mattlandis: I think this proposal is to map the ARN to the username (with some normalization to restrict to allowed characters if needed).  Otherwise a role and user with the same name would potentially have the same permissions based on RBAC.

pkg/token/token.go
```diff
@@ -52,6 +52,9 @@ type Identity struct {
 	// UserID is the unique user/role ID (e.g., "AROAAAAAAAAAAAAAAAAAA").
 	UserID string
 
+	// IAM user or role name
```
mattlandis: If we end up keeping this Go comments start with the name of the thing being commented, in this case `Name` and are complete sentences.

pkg/server/server.go
Sushant: I didn't want to handle that in this PR. As an explicit Role/User mapping exists, having a mapped AccountID doesn't really make a difference. Handling that case would mean we're changing the current behavior.

pkg/server/server.go
Sushant: You mean, username should be something like `role:Alice` or `user:Alice`?

pkg/config/types.go
Sushant: I wasn't too sure about this either. Can change the name and comment when we decide on it.

pkg/server/server.go
mattlandis: I was thinking the full ARN of the user/role.  `arn:aws:iam::012345678901:user/Alice`.  Otherwise an automapped user `Alice` in account `555555555555` would have the same access as `Alice`.

Sushant:
@mattlandis Have addressed your comments. The only thing pending now is the naming of the config field for accounts. Let me know your thoughts.

mattlandis:
I'll try and take another look today if I can get time.

pkg/config/types.go
mattlandis: I think I like AutoMappedAWSAccounts.  It identifies what accounts we are talking about and what we are doing with them.  I would like either @mattmoyer or @nckturner to offer their input as well.  Naming things is hard.

mattlandis:
Everything else looks good to me.  Once we have the name hammered down update it, add an example to the README and I'll take one last look.

For the example include an account number starting with 0 and quote the value in the list.  The yaml parser being used interprets it as a number during parsing before converting to a []string and the leading 0's get dropped.  This was a bug we found with the version we had written.

PR #53, opened 2018-02-08
Add errors types to Verify to differentiate between token and STS errors

mattlandis:
It is useful to track latency for cases where STS is called differently
than when the token passed in does not meet the requirements.  This
splits the errors returned by token.Verify into two groups
TokenFormatError and STSError to allow users to differentiate between
the two.

Signed-off-by: Matt Landis <matlan@amazon.com>


pkg/token/token_test.go
```diff
@@ -13,16 +13,25 @@ import (
...
 }
 
 func errorContains(t *testing.T, err error, expectedErr string) {
+	t.Helper()
 	if err == nil || !strings.Contains(err.Error(), expectedErr) {
 		t.Errorf("err should have contained '%s' was '%s'", expectedErr, err)
 	}
 }
 
+func isSTSError(t *testing.T, err error) {
```
nckturner: Nit: I'd prefer validateSTSError or assertSTSError because you aren't returning a bool, your asserting and erroring on failure.

mattlandis: Done.

mattlandis:
@nckturner  updated name and rebased.

Issue #47, opened 2018-01-29
Root and Federated IAM users may fail to authenticate


rtripat:
If the token is for a Root or Federated IAM user, then GetCallerIdentity returns ARNs in the following formats respectively
1. arn:aws:iam::123456789012:root
2. arn:aws:sts::123456789012:federated-user/Bob

Canonicalizing these ARNs will fail as they don't match user or assumed-role ARN patterns resulting in a failed authentication. See [here](https://github.com/heptio/authenticator/blob/master/pkg/token/token.go#L309)

Note: I discovered these while reading the code but haven't tested the scenario myself.

Reference
1) Federated IAM user: https://docs.aws.amazon.com/STS/latest/APIReference/API_GetFederationToken.html
2) IAM ARN formats: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-arns

mattlandis:
I am working on PR for this.

jonmoter:
Is this still an issue? We're having trouble authenticating as a federated IAM user (through Okta). It works as long as we create a separate IAM user.

nckturner:
@jonmoter Can you detail your issue a little more?

devkid:
If that helps, we got it to work like this:
```yaml
    - rolearn: arn:aws:iam::1234567890:role/some-role
      username: prefix:{{SessionName}}
      groups:
        - some-group
```

With the `prefix:{{SessionName}}` you can add single users to (cluster) role bindings.

Issue #46, opened 2018-01-29
Role ARNs are normalized on evaluation but not in the config


mattlandis:
The authenticator normalizes the assumed role that is returned from STS to match the IAM role, but if the RoleARN in the config is defined as a STS assumed role (arn:aws:sts::ACCOUNT:/assumed-role/RoleName/Session) instead of the IAM role (arn:aws:iam::ACCOUNT:/role/RoleName) it will never match.

I am proposing that we normalize the roles defined in the config file when it is loaded to reduce user error.

I am working on a fix if we agree that this is a valid solution.

mattmoyer:
+1, that sounds reasonable.

PR #54, opened 2018-02-08
Canonicalize ARNs in the look up map.

mattlandis:
The ARNs from the token being validated are canonicalized before lookup.
This can cause a mismatch between what is sent by the client and what
has been configured in the mapping.  This canonicalizes the map as well.

This addresses issues #47 and #46

### DIFF
--- a/pkg/arn/arn.go
+++ b/pkg/arn/arn.go
@@ -1,0 +1,68 @@
+package arn
+
+import (
+	"fmt"
+	"strings"
+
+	awsarn "github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+)
+
+// Canonicalize validates IAM resources are appropriate for the authenticator
+// and converts STS assumed roles into the IAM role resource.
+//
+// Supported IAM resources are:
+//   * AWS account: arn:aws:iam::123456789012:root
+//   * IAM user: arn:aws:iam::123456789012:user/Bob
+//   * IAM role: arn:aws:iam::123456789012:role/S3Access
+//   * IAM Assumed role: arn:aws:sts::123456789012:assumed-role/Accounting-Role/Mary (converted to IAM role)
+//   * Federated user: arn:aws:sts::123456789012:federated-user/Bob
+func Canonicalize(arn string) (string, error) {
+	parsed, err := awsarn.Parse(arn)
+	if err != nil {
+		return "", fmt.Errorf("arn '%s' is invalid: '%v'", arn, err)
+	}
+
+	if err := checkPartition(parsed.Partition); err != nil {
+		return "", fmt.Errorf("arn '%s' does not have a recognized partition", arn)
+	}
+
+	parts := strings.Split(parsed.Resource, "/")
+	resource := parts[0]
+
+	switch parsed.Service {
+	case "sts":
+		switch resource {
+		case "federated-user":
+			return arn, nil
+		case "assumed-role":
+			if len(parts) < 2 {
+				return "", fmt.Errorf("assumed-role arn '%s' does not have a role", arn)
+			}
+			role := parts[1]
+			return fmt.Sprintf("arn:%s:iam::%s:role/%s", parsed.Partition, parsed.AccountID, role), nil
+		default:
+			return "", fmt.Errorf("unrecognized resource %s for service sts", parsed.Resource)
+		}
+	case "iam":
+		switch resource {
+		case "role", "user", "root":
+			return arn, nil
+		default:
+			return "", fmt.Errorf("unrecognized resource %s for service iam", parsed.Resource)
+		}
+	}
+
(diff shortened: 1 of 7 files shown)


pkg/arn/arn.go
```diff
@@ -0,0 +1,72 @@
...
+		return "", fmt.Errorf("arn '%s' does not have a recognized partition", arn)
+	}
+
+	parts := strings.Split(parsed.Resource, "/")
+	resource := parts[0]
+
+	switch parsed.Service {
+	case "sts":
+		switch resource {
+		case "federated-user":
```
nckturner: I think we discussed this somewhat, but I just wanted to make sure that we are ok that federated-user is left as an `sts` identity even though assumed-role is translated into an `iam` identity.  Thoughts?  It seems confusing from a UX standpoint.

mattlandis: Federated user only appears as an sts value.  The role is an IAM identity when it is assumed it has the sts value with assumed role.

So these are legal:
* `arn:aws:iam::ACCOUNT_NUMBER:role/RoleName`
* `arn:aws:sts::ACCOUNT_NUMBER:assumed-role/RoleName/Session`
* `arn:aws:sts::ACCOUNT_NUMBER:federated-user/User`

These are not:
* `arn:aws:sts::ACCOUNT_NUMBER:role/RoleName`
* `arn:aws:iam::ACCOUNT_NUMBER:assumed-role/RoleName/Session`
* `arn:aws:iam::ACCOUNT_NUMBER:federated-user/User`

Esentially what we are saying is that with this change you never give a specific assumed-role permissions in the cluster, only users, roles, federated users and root accounts.

mattlandis:
@mattmoyer + anyone else.  Does anyone have opinions on this?

pkg/arn/arn.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+			role := parts[1]
+			return fmt.Sprintf("arn:%s:iam::%s:role/%s", parsed.Partition, parsed.AccountID, role), nil
+		default:
+			return "", fmt.Errorf("unrecognized resource %s for service sts", parsed.Resource)
+		}
+	case "iam":
+		switch resource {
+		case "role":
+			fallthrough
```
mumoshu: Wow, I have never realized that we could write it such way in golang!
But anyway, I guess you can safely write `case "role", "user", "root":` if you prefer it.

pkg/arn/arn.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return "", fmt.Errorf("service %s in arn %s is not a valid service for identities", parsed.Service, arn)
+}
+
+func checkPartition(partition string) error {
+	switch partition {
+	case endpoints.AwsPartitionID:
+	case endpoints.AwsCnPartitionID:
+	case endpoints.AwsUsGovPartitionID:
```
mumoshu: Ah, so nice. I personally maintain a project called kube-aws where I occasionally overlook supports for gov and cn partitions like this.
Good job!

pkg/arn/arn_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+var arnTests = []struct {
+	arn      string // input arn
+	expected string // canonacalized arn
+	err      error  // expected error value
+}{
+	{"NOT AN ARN", "", fmt.Errorf("Not an arn")},
+	{"arn:aws:iam::123456789012:user/Alice", "arn:aws:iam::123456789012:user/Alice", nil},
+	{"arn:aws:iam::123456789012:role/Users", "arn:aws:iam::123456789012:role/Users", nil},
+	{"arn:aws:sts::123456789012:assumed-role/Admin/Session", "arn:aws:iam::123456789012:role/Admin", nil},
```
mumoshu: 👍for supporting my use-case. Thx.

mattlandis:
Rebased to cover merge conflicts.

pkg/arn/arn.go
nckturner: I agree, this is a little cleaner.

pkg/arn/arn.go
rtripat: Ideally, I'd prefer if the mappings contain ARN of IAM identities only and not the other IAM resources i.e. those that are returned by GetCallerIdentity response. Namely
- arn:aws:sts::ACCOUNT_NUMBER:assumed-role/RoleName/Session
- arn:aws:sts::ACCOUNT_NUMBER:assumed-role/RoleName/*
- arn:aws:sts::ACCOUNT_NUMBER:federated-user/User
- arn:aws:iam::123456789012:root
- arn:aws:iam::123456789012:user/Bob

'arn:aws:iam::123456789012:role/RoleName' is not an IAM identity ARN instead it's a resource ARN

nckturner: Since this would change the current config format for users (they specify the role ARN), I'd prefer to change this in a V2.

mattlandis:
Flattend the switch statement with fallthroughs to a single case with multiple values.

PR #61, opened 2018-03-22
Add a NodeMapping that automatically maps a role to the name used by kubelet

nckturner:
- kubelet reports itself as `system:node:<aws-private-dns>`
- modifies authenticator to be able to discover an ec2 instance's private dns
   and use it for the node name when the role is part of a NodeMapping
- allows a role to be specified in the config file to assume for the describe
   instances call, in case it is in a separate account.

Addresses [issues/57](https://github.com/heptio/authenticator/issues/57)

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,4 +219,17 @@ server:
   mapAccounts:
   - "012345678901"
   - "456789012345"
+
+  # map nodes that should conform to the username "system:node:<private-DNS>".  This
+  # requires the authenticator to query the EC2 API in order to discover the private
+  # DNS of the EC2 instance originating the authentication request.  Optionally, you
+  # may specify a role that should be assumed before querying the EC2 API with the
+  # top level key "defaultEC2DescribeInstancesRoleARN"
+  defaultEC2DescribeInstancesRoleARN: arn:aws:iam::000000000000:role/DescribeInstancesRole
+  mapNodes:
+  - roleARN: arn:aws:iam::000000000000:role/KubernetesNode
+    groups:
+    - system:nodes
+    - system:bootstrappers
+
 ```
--- a/cmd/heptio-authenticator-aws/root.go
+++ b/cmd/heptio-authenticator-aws/root.go
@@ -75,17 +75,21 @@ func initConfig() {
 
 func getConfig() (config.Config, error) {
 	config := config.Config{
-		ClusterID:              viper.GetString("clusterID"),
-		LocalhostPort:          viper.GetInt("server.port"),
-		GenerateKubeconfigPath: viper.GetString("server.generateKubeconfig"),
-		KubeconfigPregenerated: viper.GetBool("server.kubeconfigPregenerated"),
-		StateDir:               viper.GetString("server.stateDir"),
+		ClusterID:                          viper.GetString("clusterID"),
+		DefaultEC2DescribeInstancesRoleARN: viper.GetString("defaultEC2DescribeInstancesRoleARN"),
+		LocalhostPort:                      viper.GetInt("server.port"),
+		GenerateKubeconfigPath:             viper.GetString("server.generateKubeconfig"),
+		KubeconfigPregenerated:             viper.GetBool("server.kubeconfigPregenerated"),
+		StateDir:                           viper.GetString("server.stateDir"),
 	}
 	if err := viper.UnmarshalKey("server.mapRoles", &config.RoleMappings); err != nil {
 		return config, fmt.Errorf("invalid server role mappings: %v", err)
 	}
 	if err := viper.UnmarshalKey("server.mapUsers", &config.UserMappings); err != nil {
 		logrus.WithError(err).Fatal("invalid server user mappings")
+	}
+	if err := viper.UnmarshalKey("server.mapNodes", &config.NodeMappings); err != nil {
+		logrus.WithError(err).Fatal("invalid server node mappings")
 	}
 	if err := viper.UnmarshalKey("server.mapAccounts", &config.AutoMappedAWSAccounts); err != nil {
 		logrus.WithError(err).Fatal("invalid server account mappings")
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -56,6 +56,17 @@ type UserMapping struct {
 	Groups []string
 }
 
+// NodeMapping is a mapping of an AWS Role ARN that is used by a Kubernetes node
+// to a generated Username that matches the string "system:node:NodeName"
+// where NodeName is the private DNS of the calling EC2 instance.
+type NodeMapping struct {
+	// RoleARN is the AWS Resource Name of the EC2 Instance Role. (e.g., "arn:aws:iam::000000000000:role/Foo").
+	RoleARN string
+
+	// Groups is a list of Kubernetes groups this role will authenticate as (e.g., `system:nodes`)
+	Groups []string
+}
+
 // Config specifies the configuration for a heptio-authenticator-aws server
 type Config struct {
 	// ClusterID is a unique-per-cluster identifier for your
@@ -89,7 +100,17 @@ type Config struct {
 	// Kubernetes username + groups.
 	UserMappings []UserMapping
 
+	// NodeMappings is a list of mappings from AWS IAM Role to
+	// Kubernetes node name (i.e. system:node:NodeName).
+	NodeMappings []NodeMapping
+
 	// AutoMappedAWSAccounts is a list of AWS accounts that are allowed without an explicit user/role mapping.
 	// IAM ARN from these accounts automatically maps to the Kubernetes username.
 	AutoMappedAWSAccounts []string
+
+	// DefaultEC2DescribeInstancesRoleARN is an optional AWS Resource Name for an IAM Role to be assumed
+	// before calling ec2:DescribeInstances to determine the private DNS of the calling kubelet (EC2 Instance).
+	// If nil, defaults to using the IAM Role attached to the instance where heptio-authenticator-aws is
+	// running.
+	DefaultEC2DescribeInstancesRoleARN string
 }
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -19,16 +19,25 @@ package server
 import (
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/heptio/authenticator/pkg/arn"
 	"github.com/heptio/authenticator/pkg/config"
 	"github.com/heptio/authenticator/pkg/token"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
@@ -53,9 +62,11 @@ type handler struct {
 	http.ServeMux
 	lowercaseRoleMap map[string]config.RoleMapping
 	lowercaseUserMap map[string]config.UserMapping
+	lowercaseNodeMap map[string]config.NodeMapping
 	accountMap       map[string]bool
 	verifier         token.Verifier
 	metrics          metrics
+	nodeNameProvider NodeNameProvider
 }
 
 // metrics are handles to the collectors for prometheous for the various metrics we are tracking.
@@ -95,6 +106,12 @@ func (c *Server) Run() {
 			"username": mapping.Username,
 			"groups":   mapping.Groups,
 		}).Infof("mapping IAM user")
+	}
+	for _, mapping := range c.NodeMappings {
+		logrus.WithFields(logrus.Fields{
+			"role":   mapping.RoleARN,
+			"groups": mapping.Groups,
+		}).Infof("mapping Node role")
 	}
 	for _, account := range c.AutoMappedAWSAccounts {
 		logrus.WithField("accountID", account).Infof("mapping IAM Account")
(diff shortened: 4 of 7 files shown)


nckturner:
A couple fixes in progress: 
- Get region from metadata if not specified in config.
- Allow assume role per `mapNodes` in case of multiple node groups in separate accounts.

pkg/config/types.go
```diff
@@ -56,6 +56,24 @@ type UserMapping struct {
...
+// NodeMapping is a mapping of an AWS Role ARN that is used by a Kubernetes node
+// to a generated Username that matches the string "system:node:NodeName"
+// where NodeName is the private DNS of the calling EC2 instance.
+// If EC2DescribeInstancesRoleARN is not provided, then the
+//
+type NodeMapping struct {
+	// RoleARN is the AWS Resource Name of the EC2 Instance Role. (e.g., "arn:aws:iam::000000000000:role/Foo").
+	RoleARN string
+
+	// Groups is a list of Kubernetes groups this role will authenticate as (e.g., `system:masters`)
```
mattlandis: Use groups nodes are likely to get in the example instead of system:masters.

pkg/server/server.go
```diff
@@ -274,6 +328,48 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+func (h *handler) setPrivateDNSName(id string, privateDNSName string) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+	h.nodeNameCache[id] = privateDNSName
+}
+
+func (h *handler) getPrivateDNSFromEC2(id string) (string, error) {
+	privateDNSName, err := h.getPrivateDNSName(id)
+	if err == nil {
+		return fmt.Sprintf("%s%s", config.NodeNamePrefix, privateDNSName), nil
```
mattlandis: For this string concatenation will probably work best and is easy to ready `config.NodeNamePrefix + privateDNSName`.

pkg/server/server.go
```diff
@@ -274,6 +328,48 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+			if aws.StringValue(instance.InstanceId) == id {
+				privateDNSName = aws.StringValue(instance.PrivateDnsName)
+				h.setPrivateDNSName(id, privateDNSName)
+			}
+		}
+	}
+	if privateDNSName == "" {
+		return "", errors.New(fmt.Sprintf("failed to find private DNS Name for %s", id))
+	}
+	return fmt.Sprintf("%s%s", config.NodeNamePrefix, privateDNSName), nil
```
mattlandis: String concat.

cmd/heptio-authenticator-aws/root.go
```diff
@@ -75,17 +75,22 @@ func initConfig() {
 
 func getConfig() (config.Config, error) {
 	config := config.Config{
-		ClusterID:              viper.GetString("clusterID"),
-		LocalhostPort:          viper.GetInt("server.port"),
-		GenerateKubeconfigPath: viper.GetString("server.generateKubeconfig"),
-		StateDir:               viper.GetString("server.stateDir"),
+		Region: viper.GetString("region"),
```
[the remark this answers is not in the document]
nckturner: Actually this is correct gofmt.  Not sure why, but I verified it corrects it to be this if I indent it.  Moving the entire line down to be last in the block, forces it to line up with the other lines.

nckturner:
Not going to add a fix for node groups in multiple separate accounts unless someone asks for it, since the aws cloudprovider can only currently create loadbalancers in a single separate account.

pkg/server/server.go
```diff
@@ -159,6 +183,34 @@ func (c *Server) getHandler() *handler {
...
+			logrus.WithError(err).Fatal("Region not found in config and failed to discover from ec2 instance metadata")
+		}
+		cfg.Region = aws.String(region)
+	}
+	sess = session.Must(session.NewSession(&cfg))
+	if c.DefaultEC2DescribeInstancesRoleARN != "" {
+		logrus.WithFields(logrus.Fields{
+			"roleARN": c.DefaultEC2DescribeInstancesRoleARN,
+		}).Infof("Using assumed role for EC2 API")
+		cfg.Credentials = stscreds.NewCredentials(sess, c.DefaultEC2DescribeInstancesRoleARN)
```
rtripat: STS credentials duration can be increased up to maximum of 1 hour to reduce API calls

pkg/server/server.go
```diff
@@ -238,6 +290,16 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
 	} else if userMapping, exists := h.lowercaseUserMap[arnLower]; exists {
 		username = userMapping.Username
 		groups = userMapping.Groups
+	} else if nodeMapping, exists := h.lowercaseNodeMap[arnLower]; exists {
```
rtripat: If the config contains Node's IAM role as a RoleMapping then this block won't be executed right? Do we want that to happen?

pkg/server/server.go
```diff
@@ -238,6 +290,16 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+		privateDNS, err := h.getPrivateDNSFromEC2(identity.SessionName)
```
rtripat: either the function should return privateDNS or the function name should be getUsernameForNode

pkg/server/server.go
```diff
@@ -274,6 +336,51 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+	privateDNSName, err := h.getPrivateDNSName(id)
+	if err == nil {
+		return config.NodeNamePrefix + privateDNSName, nil
+	}
+
+	// Look up instance from EC2 API
+	instanceIds := []*string{&id}
+	ec2Service := ec2.New(h.sess)
+	output, err := ec2Service.DescribeInstances(&ec2.DescribeInstancesInput{
+		InstanceIds: instanceIds,
```
rtripat: Consider using privateDnsName filter 

https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#DescribeInstancesInput

nckturner: All I know is the instanceId when making the request, so I wouldn't be able to use the filter, unless I don't understand how it works?

pkg/server/server.go
nckturner: Right.  I'm not sure if we care which takes precedence, but if you have an argument one way or the other I'm all ears.

pkg/server/server.go
nckturner: Fixed.

pkg/server/server.go
nckturner: Changed to 1 hr.

pkg/server/server.go
rtripat: If the authenticated user is an EC2 instance, then NodeMapping should override.

You can determine whether the authenticated user is an EC2 instance or not by inspecting the UserId field in GetCallerIdentity response. It should be in the format "role-id:ec2-instance-id" as per https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable

pkg/server/server.go
rtripat: I made a mistake while reading through the API documentation and presumed it's an output filter. Ignore this comment.

nckturner:
Per conversation with @rtripat, we think if the mapping includes a `system:node`, `system:bootstrappers` or `system:proxier` group, or a username of "system:nodes:*",  the authenticator should validate at least that the InstanceId [passed in the UserId of the sts GetCallerIdentity request](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable), which is settable by the user, is an InstanceId that belongs to an instance in the node account by calling `ec2 DescribeInstances`.  This will help to prevent users from impersonating nodes

pkg/server/server.go
```diff
@@ -276,6 +307,111 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+	}
+
+	// Look up instance from EC2 API
+	instanceIds := []*string{&id}
+	ec2Service := ec2.New(p.sess)
+	output, err := ec2Service.DescribeInstances(&ec2.DescribeInstancesInput{
+		InstanceIds: instanceIds,
+	})
+	if err != nil {
+		return "", err
```
rtripat: nit: add a more explicit message if the api returns an error because instance doesn't exist.

pkg/config/types.go
```diff
@@ -89,7 +100,20 @@ type Config struct {
...
 	AutoMappedAWSAccounts []string
+
+	// DefaultEC2DescribeInstancesRoleARN is an optional AWS Resource Name for an IAM Role to be assumed
+	// before calling ec2:DescribeInstances to determine the private DNS of the calling kubelet (EC2 Instance).
+	// If nil, defaults to using the IAM Role attached to the instance where heptio-authenticator-aws is
+	// running.
+	DefaultEC2DescribeInstancesRoleARN string
+
+	// Region is the AWS region, such as us-east-1.  If nil, discovered by calling ec2 metadata.
+	Region string
```
rtripat: This can be removed an inferred from EC2 instance metadata. If it's running on a non-EC2 instance like for development then it can be inferred from environment variables along with credentials.

pkg/server/server.go
nckturner: That should be caught by the next return, but I made both errors more explicit.

nckturner:
Squashed commits and addressed @rtripat last feedback.

pkg/server/server.go
```diff
@@ -276,6 +307,103 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
 	})
 }
 
+type NodeNameProvider interface {
```
mattlandis: This could be renamed to be local as it is not used outside this package.

pkg/server/server.go
```diff
@@ -276,6 +307,103 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+	GetNodeName(string) (string, error)
+}
+
+type nodeNameEC2PrivateDNSProvider struct {
+	sess          *session.Session
+	nodeNameCache map[string]string
+	lock          sync.Mutex
+}
+
+func NewNodeNameEC2PrivateDNSProvider(roleARN string) NodeNameProvider {
```
mattlandis: This could be renamed to be local as it is not used outside this package.

pkg/server/server.go
```diff
@@ -276,6 +307,103 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+
+		sess.Config.Credentials = credentials.NewCredentials(ap)
+	}
+	return sess
+}
+
+func (p *nodeNameEC2PrivateDNSProvider) getPrivateDNSName(id string) (string, error) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	name, ok := p.nodeNameCache[id]
```
mattmoyer: Do we potentially need to worry about flushing this cache? Assuming this daemon might be running for 90+ days without a restart?

I'm guessing not anymore since EC2 instance IDs are so much longer.

pkg/server/server.go
```diff
@@ -276,6 +307,103 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+	}
+
+	// Look up instance from EC2 API
+	instanceIds := []*string{&id}
+	ec2Service := ec2.New(p.sess)
+	output, err := ec2Service.DescribeInstances(&ec2.DescribeInstancesInput{
+		InstanceIds: instanceIds,
+	})
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("failed querying private DNS from EC2 API for node %s: %s", id, err.Error()))
```
mattmoyer: Nit: this could be:
```go
return "", fmt.Errorf("failed querying private DNS from EC2 API for node %s: %s", id, err.Error())
```

README.md
```diff
@@ -219,4 +219,17 @@ server:
...
   - "012345678901"
   - "456789012345"
+
+  # map nodes that should conform to the username "system:node:<private-DNS>".  This
+  # requires the authenticator to query the EC2 API in order to discover the private
+  # DNS of the EC2 instance originating the authentication request.  Optionally, you
+  # may specify a role that should be assumed before querying the EC2 API with the
+  # top level key "defaultEC2DescribeInstancesRoleARN"
+  defaultEC2DescribeInstancesRoleARN: arn:aws:iam::000000000000:role/DescribeInstancesRole
+  mapNodes:
```
mattmoyer: I'm fine with this approach as written, but but what if instead of a special `mapNodes` config, we added a new template variable in `mapRoles` like `{{SessionName}}` but for the instance private DNS name.

Something like:
```yaml
   mapRoles:
   - roleARN: arn:aws:iam::000000000000:role/KubernetesNode
     username: system:node:{{EC2InstancePrivateDNSName}}
     groups:
     - system:nodes
     - system:bootstrappers
```

If that variable was present in the template, we'd do the EC2 DescribeInstances lookup to get the DNS name.

The main improvement would be that it could also be used with other EC2 instance attributes like `{{EC2InstanceVPCID}}`, `{{EC2InstanceSubnetID}}`, `{{EC2InstanceRegion}}`, etc..

pkg/server/server.go
```diff
@@ -231,7 +252,17 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
 	log = log.WithField("arn", identity.CanonicalARN)
 	var username string
 	var groups []string
-	if roleMapping, exists := h.lowercaseRoleMap[arnLower]; exists {
+	if nodeMapping, exists := h.lowercaseNodeMap[arnLower]; exists {
+		nodeName, err := h.nodeNameProvider.GetNodeName(identity.SessionName)
```
SaranBalaji90: If I understand this correctly, I can use instance-id as session name and gain access to my kubernetes cluster right?

mattlandis: Only if you have the ability to assume the role that is in the nodeMapping.  Which is why it is critical that only ec2 be able to assume the roles you use for instances.

README.md
nckturner: Gave that approach a go in https://github.com/heptio/authenticator/pull/65

pkg/server/server.go
nckturner: I think it should be ok esp with longer instance ids.

mattmoyer: I think you're right, I just wanted to make sure we'd thought of this. In any case even if there was a collision I couldn't imagine it being exploitable in any practical way.

nckturner:
Closed in favor of https://github.com/heptio/authenticator/pull/65

PR #65, opened 2018-03-27
Add a new template variable {{EC2PrivateDNSName}} 

nckturner:
- kubelet reports itself as `system:node:<aws-private-dns>`
- modifies authenticator to be able to discover an ec2 instance's private dns
  and use it for the node name when {{EC2PrivateDNSName}} is used in the
  template username
- allows a role to be specified in the config file to assume for the describe
  instances call, in case it is in a separate account.

Fixes: https://github.com/heptio/authenticator/issues/57
Alternative to: https://github.com/heptio/authenticator/pull/61

Signed-off-by: Nick Turner <nic@amazon.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,9 @@ server:
   # output `path` where a generated webhook kubeconfig will be stored.
   generateKubeconfig: /etc/kubernetes/heptio-authenticator-aws.kubeconfig # (default)
 
+  # role to assume before querying EC2 API in order to discover metadata like EC2 private DNS Name
+  ec2DescribeInstancesRoleARN: arn:aws:iam::000000000000:role/DescribeInstancesRole
+
   # each mapRoles entry maps an IAM role to a username and set of groups
   # Each username and group can optionally contain template parameters:
   #  1) "{{AccountID}}" is the 12 digit AWS ID.
@@ -194,6 +197,17 @@ server:
     groups:
     - system:bootstrappers
     - aws:instances
+
+  # map nodes that should conform to the username "system:node:<private-DNS>".  This
+  # requires the authenticator to query the EC2 API in order to discover the private
+  # DNS of the EC2 instance originating the authentication request.  Optionally, you
+  # may specify a role that should be assumed before querying the EC2 API with the
+  # top level key "defaultEC2DescribeInstancesRoleARN".
+  - roleARN: arn:aws:iam::000000000000:role/KubernetesNode
+    username: system:node:{{EC2PrivateDNSName}}
+    groups:
+    - system:nodes
+    - system:bootstrappers
 
   # map federated users in my "KubernetesAdmin" role to users like
   # "admin:alice-example.com". The SessionName is an arbitrary role name
@@ -219,4 +233,5 @@ server:
   mapAccounts:
   - "012345678901"
   - "456789012345"
+
 ```
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -19,16 +19,26 @@ package server
 import (
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
+	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/heptio/authenticator/pkg/arn"
 	"github.com/heptio/authenticator/pkg/config"
 	"github.com/heptio/authenticator/pkg/token"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
@@ -48,6 +58,12 @@ var tokenReviewDenyJSON = func() []byte {
 	return res
 }()
 
+// Pattern to match EC2 instance IDs
+var (
+	instanceIDPattern = regexp.MustCompile("^i-(\\w{8}|\\w{17})$")
+	dns1123Pattern    = regexp.MustCompile("[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*")
+)
+
 // server state (internal)
 type handler struct {
 	http.ServeMux
@@ -56,6 +72,7 @@ type handler struct {
 	accountMap       map[string]bool
 	verifier         token.Verifier
 	metrics          metrics
+	ec2Provider      EC2Provider
 }
 
 // metrics are handles to the collectors for prometheous for the various metrics we are tracking.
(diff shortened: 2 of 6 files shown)


pkg/server/server.go
```diff
@@ -287,12 +284,146 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+	template = strings.Replace(template, "{{AccountID}}", identity.AccountID, -1)
+
 	// usernames and groups must be a DNS-1123 hostname matching the regex
 	// "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
 	sessionName := strings.Replace(identity.SessionName, "@", "-", -1)
-
-	template = strings.Replace(template, "{{AccountID}}", identity.AccountID, -1)
 	template = strings.Replace(template, "{{SessionName}}", sessionName, -1)
-	return template
+
```
mattlandis: Since usernames must be DNS-1123 compliant should we have a final check to ensure the rendered template is valid and return an error if it is not?  For instance if I accidentally put in `{{AccountId}}` instead of `{{AccountID}}` it will not get replaced and we will return an invalid username to k8s.

pkg/server/server.go
```diff
@@ -287,12 +284,146 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+	for _, reservation := range output.Reservations {
+		for _, instance := range reservation.Instances {
+			if aws.StringValue(instance.InstanceId) == id {
+				privateDNSName = aws.StringValue(instance.PrivateDnsName)
+				p.setPrivateDNSNameCache(id, privateDNSName)
+			}
+		}
+	}
+	if privateDNSName == "" {
+		return "", errors.New(fmt.Sprintf("failed to find node %s", id))
```
mattlandis: You can use [fmt.Errorf](https://golang.org/pkg/fmt/#Errorf) here.

pkg/server/server.go
```diff
@@ -287,12 +284,146 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
 	sessionName := strings.Replace(identity.SessionName, "@", "-", -1)
-
-	template = strings.Replace(template, "{{AccountID}}", identity.AccountID, -1)
 	template = strings.Replace(template, "{{SessionName}}", sessionName, -1)
-	return template
+
+	return template, nil
+}
+
+type EC2Provider interface {
```
mattlandis: This can be local scope.

mattlandis:
Also can we open two issues
1) add metrics for this since it is a place latency could be introduced
2) Examine managing cache lifetime so that it won't be an issue for long running clusters with frequent node changes

pkg/server/server.go
```diff
@@ -287,12 +284,149 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+		return identity.CanonicalARN, []string{}, nil
+	}
+	return "", nil, fmt.Errorf("ARN is not mapped: %s", arn)
+}
+
+func (h *handler) renderTemplate(template string, identity *token.Identity) (string, error) {
+
+	// Private DNS requires EC2 API call
+	if strings.Contains(template, "{{EC2PrivateDNSName}}") {
+		re := regexp.MustCompile("^i-(\\w{8}|\\w{17})$")
```
dbenhur: I've seen `(^i-(\w{8}|\w{17})$)|(^mi-\w{17}$)` used elsewhere.  EC2 apparently also has [managed instances](https://docs.aws.amazon.com/systems-manager/latest/userguide/managed_instances.html) for hybrid environments with "mi-" id prefix.  Not sure if these should be allowed or not (are they covered by getPrivateDNSName?).

mattlandis: @dbenhur Let's open a issue for this.  I think it requires more investigation to make sure there are no other side effects we aren't thinking about.

@nckturner Add the sign-off and I think this is good to go.

README.md
```diff
@@ -173,6 +173,9 @@ server:
   # output `path` where a generated webhook kubeconfig will be stored.
   generateKubeconfig: /etc/kubernetes/heptio-authenticator-aws.kubeconfig # (default)
 
+  # role to assume before querying EC2 API in order to discover metadata like EC2 private DNS Name
+  defaultEC2DescribeInstancesRoleARN: arn:aws:iam::000000000000:role/DescribeInstancesRole
```
mattmoyer: Nit: the "default" part here is slightly confusing to me. Maybe something like `serverAssumeRoleARN`?

pkg/server/server.go
nckturner: Cool, will rebase and sign off.

pkg/config/constants.go
```diff
@@ -29,4 +29,7 @@ const (
 
 	// certLifetime is the lifetime of the CA certificate (100 years)
 	certLifetime = time.Hour * 24 * 365 * 100
+
+	// nodeNamePrefix is the username prefix that the apiserver expects of kubelet
+	NodeNamePrefix = "system:node:"
```
mattmoyer: This is no longer needed as far as I can tell.

pkg/server/server.go
```diff
@@ -287,12 +284,149 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+		groups := []string{}
+		for _, groupPattern := range roleMapping.Groups {
+			group, err := h.renderTemplate(groupPattern, identity)
+			if err != nil {
+				return "", nil, fmt.Errorf("failed mapping group: %s", err.Error())
+			}
+			groups = append(groups, group)
+		}
+		return username, groups, nil
+	} else if userMapping, exists := h.lowercaseUserMap[arn]; exists {
```
mattmoyer: Nit: these `else if` clauses could now be plain `if` clauses since we return early at the end of each one.

pkg/server/server.go
mattmoyer: I think it'd also be fine to do minimal validation here and just trust that `ec2:DescribeInstances` is going to return an appropriate error if the instance ID is invalid.

We could also stick this regex in a global `var instanceIDPattern = regexp.MustCompile(...)` so we only compile once at init.

pkg/server/server.go
```diff
@@ -287,12 +284,149 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
+	template = strings.Replace(template, "{{AccountID}}", identity.AccountID, -1)
+
 	// usernames and groups must be a DNS-1123 hostname matching the regex
 	// "[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
 	sessionName := strings.Replace(identity.SessionName, "@", "-", -1)
-
-	template = strings.Replace(template, "{{AccountID}}", identity.AccountID, -1)
 	template = strings.Replace(template, "{{SessionName}}", sessionName, -1)
-	return template
+	re := regexp.MustCompile("[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*")
```
mattmoyer: Same note here re: using a global `var dns1123Pattern = regexp.MustCompile(...)` compiled at init.

README.md
nckturner: Ah, default is incorrect now.  I like having EC2DescribeInstances in the name though, as I think it helps differentiate it from the other references to roles.  How about `serverEC2DescribeInstancesRoleARN`?

pkg/config/constants.go
nckturner: Removed.

pkg/server/server.go
nckturner: Done.

README.md
mattmoyer: Yeah I like that.

pkg/server/server.go
nckturner: Moved to global.

nckturner:
Squashed review commit.

PR #72, opened 2018-04-07
Adding support for kubectl ExecCredential output

### LINKED ISSUE #64: Support kubectl external credential provider
Hi,
we have integrated the heptio authenticator in our workflow and it has been working great. The Login process with kubectl is still a little complicated though. kubectl in kubernetes 1.10 added an interesting option for external auth providers https://github.com/kubernetes/features/issues/541
It would be great if the authenticator could support this.

Regards,
Sebastian

christopherhein:
**Why:**

* Allows you to use the `heptio-authenticator-aws` binary in your `~/.kube/config` as an external `ExecCredential`

**Usage:**

```yaml
users:
- name: user
  user:
    exec:
      command: heptio-authenticator-aws
      args:
        - "token"
        - "--cluster-id"
        - "cluster-id"
```

* closes #64

**Todos:**

- [x] Documentation for usage on the `readme.md`
- [x] Add error handling to the token response (not sure if this is possible or needed)

**Questions:**

* Do we need to implement a check for `KUBERNETES_EXEC_INFO`?
* Do we need the `expirationTimestamp` it's optional per the documentation, if so could you provide me a little guidance on where to make or get that from?

https://kubernetes.io/docs/admin/authentication/#client-go-credential-plugins


nckturner:
Can you separate vendor into a separate commit to make this more reviewable?

christopherhein:
@nckturner Updated with commits separated, I should have error handling done soon to review that as well.

christopherhein:
I was mistaken about the flow for the authenticator, I thought it passed a token to the API server that it validated after learning a little more it's passing a signed URL that the server uses to in essence call`sts:getCallerIdentity` with that being the flow I'm not sure how I can do error handling like defined - https://kubernetes.io/docs/admin/authentication/#input-and-output-formats  

Any thoughts or suggestions?

cmd/heptio-authenticator-aws/token.go
```diff
@@ -35,7 +35,7 @@ var tokenCmd = &cobra.Command{
 		clusterID := viper.GetString("clusterID")
 
 		if clusterID == "" {
-			fmt.Fprintf(os.Stderr, "error: cluster ID not specified\n")
+			fmt.Fprintf(os.Stderr, "Error: cluster ID not specified\n")
```
christopherhein: Updated to be cohesive with the default Cobra error responses. `heptio-authenticator-aws token -i` for example returns:

```sh
Error: flag needs an argument: 'i' in -i
```

pkg/token/token.go
```diff
@@ -180,6 +184,21 @@ func (g generator) GetWithRole(clusterID string, roleARN string) (string, error)
 	return v1Prefix + base64.RawURLEncoding.EncodeToString([]byte(presignedURLString)), nil
 }
 
+// FormatJSON formats the json to support 1.10 external authProvider
```
christopherhein: Updated in a later commit.

christopherhein:
@nckturner updated based on your review. For the MFA, assuming we can track that separately through https://github.com/heptio/authenticator/pull/42 right?

nckturner:
I'd like to merge this after https://github.com/heptio/authenticator/pull/42, I've proposed how we do that on that PR.

PR #73, opened 2018-04-10
Add --log-format server flag to configure output format

phillipj:
These changes allows the logger format to be specified when starting the
server, either to `text` (default) or `json`.

I'm also interested in adding support for the fluentd format, though that would mean adding a 3rd party dependency: https://github.com/joonix/log.

Any thoughts about the changes as is and/or possible new dependency?

Refs https://github.com/heptio/authenticator/issues/52

### DIFF
--- a/cmd/aws-iam-authenticator/root.go
+++ b/cmd/aws-iam-authenticator/root.go
@@ -48,9 +48,10 @@ func Execute() {
 }
 
 func init() {
-	logrus.SetFormatter(&logrus.TextFormatter{FullTimestamp: true})
 	cobra.OnInitialize(initConfig)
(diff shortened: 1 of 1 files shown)


phillipj:
Thanks for reviewing! You not wanting to add a new dependency for this is totally understandable, no problem for me dropping that idea.

Pretty sure I got that CLA signed now, at least that's what I interpret from the verification email titled `You just signed CNCF Individual Contributor License Agreement`. The inline status in this PR hasn't updated though, but that might take a little while? 🤷‍♀️

nckturner:
Can you rebase @phillipj?

phillipj:
Sure thing, rebased and force pushed -- conflicts were trivial enough that a 3-way merge did the trick.

Issue #63, opened 2018-03-26
Make 15-minute token limit configurable


chrisleck:
15 minutes is annoying short for my dev team's experience.  We use Okta federated AWS authentication, so having to go through the password + 2FA dance every 15 minutes is really asking too much.

jsco2t:
Our team has a similar setup - and have been using the generated `tokens` to perform auth into the `k8s dashboard`. This same 15 minute limitation makes it a serious pain to do any real work with the dashboard.

nckturner:
Increasing the pre-signed URL duration is not something we can do at this point, but let me know if #193 addressed the difficulties around MFA.  It caches the temporary credentials instead of the token.

norganna:
For anyone else affected by this 15 minute limitation that is interested in a sideways hack around this issue, I have created a proxy that performs the regular aws-iam-authentication every 10 minutes for you and injects the tokens into the requests as it forwards them on.

Hopefully one day there will be the ability to authenticate for more than 15 minutes at a time, but for now this solution improves my productivity:

[kdash authenticating dashboard proxy](https://github.com/norganna/kdash)

PR #75, opened 2018-04-13
[WIP] Add "duration" option for token command.

kamatama41:
This addresses #63 
It would make users happy who feel like the 15 mins is too short :)

### DIFF
--- a/cmd/heptio-authenticator-aws/token.go
+++ b/cmd/heptio-authenticator-aws/token.go
@@ -33,6 +33,7 @@ var tokenCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		roleARN := viper.GetString("role")
 		clusterID := viper.GetString("clusterID")
+		duration := viper.GetInt("duration")
 
 		if clusterID == "" {
 			fmt.Fprintf(os.Stderr, "Error: cluster ID not specified\n")
@@ -49,10 +50,10 @@ var tokenCmd = &cobra.Command{
 		}
 		if roleARN != "" {
 			// if a role was provided, assume that role for the token
-			tok, err = gen.GetWithRole(clusterID, roleARN)
+			tok, err = gen.GetWithRole(clusterID, roleARN, duration)
 		} else {
 			// otherwise sign the token with immediately available credentials
-			tok, err = gen.Get(clusterID)
+			tok, err = gen.Get(clusterID, duration)
 		}
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "could not get token: %v\n", err)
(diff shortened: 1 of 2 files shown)


cmd/heptio-authenticator-aws/token.go
```diff
@@ -67,4 +68,6 @@ func init() {
 	tokenCmd.Flags().StringP("role", "r", "", "Assume an IAM Role ARN before signing this token")
 	viper.BindPFlag("role", tokenCmd.Flags().Lookup("role"))
 	viper.BindEnv("role", "DEFAULT_ROLE")
+	tokenCmd.Flags().IntP("duration", "d", 15, "Seconds that the token is valid for.")
```
mattmoyer: I think this flag could be a [`DurationP`](https://godoc.org/github.com/spf13/pflag#FlagSet.DurationP) which should be a bit cleaner.

pkg/token/token.go
```diff
@@ -172,8 +172,11 @@ func (g generator) GetWithRole(clusterID string, roleARN string) (string, error)
 	// if a roleARN was specified, replace the STS client with one that uses
 	// temporary credentials from that role.
 	if roleARN != "" {
+		opt := func(p *stscreds.AssumeRoleProvider) {
+			p.Duration = time.Duration(duration) * time.Minute
```
mattmoyer: The usage message above says "Seconds" but this looks like the argument is interpreted as minutes? I think passing a `time.Duration` all the way through (rather than `int`) might be better to avoid this type of confusion.

kamatama41: The usage is wrong. I intended "Minutes" 🙇

kamatama41:
@mattlandis @mattmoyer  
Thank you for reviewing my PR!
I misunderstood that the 15 mins limitation just comes from the duration for the AssumeRoleProvider, but seems to depend on the limitation for signed request as well. (Sorry I had not tested this PR before created.. :bow:) So I think this PR should be closed and the design should be reconsidered. What do you think?

mhobotpplnet:
This is VERY annoying when using eks with iam roles and logging into k8 dashboard. 
Literally every 15min one would need to re-login.

bobhenkel:
Yeah, we use the command below to give users a token to log into the kube dashboard with a token. Works great minus that every 15 minutes their browser session closes on them and they have to get a new token.  For kubectl people don't notice that sometimes a command takes longer, but they sure do notice in the web ui when it says session expired.

```
heptio-authenticator-aws token -i clustername | jq .status.token -r
```

okgolove:
Are there any news? It's really very annoying.

PR #81, opened 2018-05-15
Pull the response body so we can log it if there's an error

jbell-shibumi:
Change when the response body is read so that we can log it in the output if there's an error.

Makes issue diagnosis easier, for instance, if you have a mistaken cluster ID


jbell-shibumi:
Missing sign-off and I can't seem to figure out how to add it.. will open a new pr

PR #82, opened 2018-05-15
Pull the response body so we can log it if there's an error

jbell-shibumi:
Signed-off-by: Johnathan Bell <johnathan.bell@shibumi.com>

Added response body logging for better diagnosis of issues.  Based on v0.1.0 tag

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -212,13 +212,13 @@ func Verify(token string, serverID string) (*Identity, error) {
 	}
 	defer response.Body.Close()
 
-	if response.StatusCode != 200 {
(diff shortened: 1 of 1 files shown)


christopherhein:
Seems like a reasonable addition.

nckturner:
I think it could be helpful to have response body as an optional field to print to logs, but it would be too much if we are unable to turn it off.  Would you like to make that fix and rebase on master @jbell-shibumi?

jbell-shibumi:
Let me see what i can do @nckturner -- I'm not that familiar with golang...

mattlandis:
Closing due to inactivity.  Feel free to reopen or open new PR with the issues addressed.

PR #103, opened 2018-06-20
Add support for ARNs with paths.

### LINKED ISSUE #98: Error matching iam role when role uses a path
I'm having a problem getting the iam roles to match when the assumed role contains a path.

```
kubeconfig settings:
apiVersion: v1
clusters:
- cluster:
    server: <endpoint>
    certificate-authority-data: <cert>
  name: kubernetes
contexts:
- context:
    cluster: kubernetes
    user: aws
  name: aws
current-context: aws
kind: Config
preferences: {}
users:
- name: aws
  user:
    exec:
      command: heptio-authenticator-aws
      args:
        - "token"
        - "-i"
        - "eks-cluster"
        - "auth-test"
        - "-r"
        - "arn:aws:iam::000000000000:role/roles/SomeRole"

```
Non Working ConfigMap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: aws-auth
  namespace: kube-system
data:
  mapRoles: |
    - rolearn: arn:aws:iam::000000000000:role/worker-node
      username: system:node:{{EC2PrivateDNSName}}
      groups:
        - system:bootstrappers
        - system:nodes
    - rolearn: arn:aws:iam::000000000000:role/roles/SomeRole
      username: admin:{{SessionName}}
      groups:
        - system:masters
```

Working ConfigMap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: aws-auth
  namespace: kube-system
data:
  mapRoles: |
    - rolearn: arn:aws:iam::000000000000:role/worker-node
      username: system:node:{{EC2PrivateDNSName}}
      groups:
        - system:bootstrappers
        - system:nodes
    - rolearn: arn:aws:iam::000000000000:role/SomeRole
      username: admin:{{SessionName}}
      groups:
        - system:masters
```

As you can see from the non-working and working comparison I have to trim out the iam path to get the authentication to work.  (arn:aws:iam::000000000000:role/roles/SomeRole vs  arn:aws:iam::000000000000:role/SomeRole) Technically on AWS side the arn with the trimmed path is not the same role.

mattlandis:
IAM ARNs can have paths (ex.
arn:aws:iam::123456789012:assumed-role/org/team/rolename/session).  When
normalizing the assumed role the output only contained the first part of
the path after assumed-role.

This should fix #98 .

### DIFF
--- a/pkg/arn/arn.go
+++ b/pkg/arn/arn.go
@@ -36,10 +36,11 @@ func Canonicalize(arn string) (string, error) {
 		case "federated-user":
 			return arn, nil
 		case "assumed-role":
-			if len(parts) < 2 {
+			if len(parts) < 3 {
 				return "", fmt.Errorf("assumed-role arn '%s' does not have a role", arn)
 			}
-			role := parts[1]
+			// IAM ARNs can contain paths, part[0] is resource, parts[len(parts)] is the SessionName.
+			role := strings.Join(parts[1:len(parts)-1], "/")
 			return fmt.Sprintf("arn:%s:iam::%s:role/%s", parsed.Partition, parsed.AccountID, role), nil
 		default:
 			return "", fmt.Errorf("unrecognized resource %s for service sts", parsed.Resource)
--- a/pkg/arn/arn_test.go
+++ b/pkg/arn/arn_test.go
@@ -16,6 +16,7 @@ var arnTests = []struct {
 	{"arn:aws:sts::123456789012:assumed-role/Admin/Session", "arn:aws:iam::123456789012:role/Admin", nil},
 	{"arn:aws:sts::123456789012:federated-user/Bob", "arn:aws:sts::123456789012:federated-user/Bob", nil},
 	{"arn:aws:iam::123456789012:root", "arn:aws:iam::123456789012:root", nil},
+	{"arn:aws:sts::123456789012:assumed-role/Org/Team/Admin/Session", "arn:aws:iam::123456789012:role/Org/Team/Admin", nil},
 }
 
 func TestUserARN(t *testing.T) {


christopherhein:
:clap: :clap: :clap: Nice work @mattlandis

nckturner:
Looks good to me.

jpb:
Has anyone had success using IAM roles with paths with this fix? From what I've seen anecdotally, assumed roles do not encode the role's path in their ARN, and [AWS documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_identityandaccessmanagement.html) appears to confirm this:

Assumed role ARN format:

```
arn:${Partition}:iam::${Account}:assumed-role/${RoleName}/${RoleSessionName}
```

compared to the role ARN format:

```
arn:${Partition}:iam::${Account}:role/${RoleNameWithPath}
```

M00nF1sh:
I have tested it works with some caveats:
suppose i have an role named `arn:aws:iam::283511030707:role/dev/product/m00nf1sh`, I need to configrue `mapRoles` with `arn:aws:iam::283511030707:role/m00nf1sh` instead of `arn:aws:iam::283511030707:role/dev/product/m00nf1sh` to make it work. (**this can cause potential privilege escalation**)

BTW, in the pr, `parts[len(parts)] is the SessionName.` should be` parts[len(parts)-1] is the SessionName`. and `arn:aws:sts::123456789012:assumed-role/Org/Team/Admin/Session` should be `arn:aws:sts::123456789012:assumed-role/Admin/Session`(since assumed-role doesn't contain path)

ch3lo:
I agree with @MarkPare , I had inverted 2 days deploying a EKS cluster in Landing Zone without success and the problem was my instances roles had a path in the ARN. In my comnpany this is a requirement because we group our IAM Roles in path for security reasons.

Regards

michael-burt:
I think this is still broken. I spent a ton of time finding this bug, and I can confirm that it still there.

PR #108, opened 2018-06-27
Renamed heptio-authenticator-aws -> aws-iam-authenticator

nckturner:
Now that the repository has moved, we can rename to aws-iam-authenticator.

Signed-off-by: Nick Turner <nic@amazon.com>

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -2,8 +2,8 @@
 project_name: authenticator
 
 builds:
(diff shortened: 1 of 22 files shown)


mattmoyer:
/lgtm

> Do you want to update the go paths in this commit as well? Or should it be a separate commit?

Either way works for me.

Thanks for doing this!

nckturner:
I can add that in as well.

PR #114, opened 2018-07-03
userarn should be the key instead of userARN

srinivasreddych:
Under mapUsers, userARN should be replaced with userarn. kubectl doesnt add the user if the key is userARN

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ server:
   # each mapUsers entry maps an IAM role to a static username and set of groups
   mapUsers:
   # map user IAM user Alice in 000000000000 to user "alice" in group "system:masters"
-  - userARN: arn:aws:iam::000000000000:user/Alice
+  - userarn: arn:aws:iam::000000000000:user/Alice
     username: alice
     groups:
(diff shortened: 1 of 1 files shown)


niclic:
Having integrated with KOPS without any issues since v0.1.0, I just ran into this very problem when deploying the same `ConfigMap` that we were using with KOPS to an EKS cluster (v0.3.0). Took a long time to figure out. Has the configuration changed?

christopherhein:
It's actually a configuration difference between EKS and the OSS `aws-iam-authenticator`. We're in conversations about changing the loading of these roles using CRDs instead check out - https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/116 for more information on that.

For now the OSS distro still uses the `userARN` & `roleARN` and EKS uses `userarn` & `rolearn`.

nckturner:
Closing this, as @christopherhein said, its a (confusing) difference between OSS and EKS authenticator. Were working on moving to the OSS configuration, but for now, EKS users should refer to the EKS docs.

PR #121, opened 2018-07-12
Mapped existing sts caller specified role name

### LINKED ISSUE #120: Using the RoleAssumption option of the `token` cli does not persist underlying Session Name
Thanks for the great tool!  We are working on leveraging this for a client and overall this has gone very smoothly.  

One facet of our deployment environment is that we use federation via ActiveDirectory to create a user's STS session which embeds their AD username as the Session Name.  This allows very easy analysis of our CloudTrail logs to quickly assess the individual who executed those actions. 

When using the `token -i <cluster name>` facility this SessionName is fully mapped onto the audit logs of Kubernetes beautifully, providing us a similarly easy way to look directly at the logs to determine "who did that?". 

What we wanted to provide though is a bit more flexibility to our users on how they authenticate to the K8 cluster.  What that means in practice is setting up a set of IAM roles that can be assumed from those "seed" IAM roles that ADFS provided.  This design allows us the ability to have a user authenticate a bit more flexibility
Ex: 
- map two different IAM roles -> K8 that are assumable by the same group of individuals to allow them a Read Only access vs. Admin access to the prod cluster so they have to be deliberate about how they auth
- Decouple the ADFS mapped roles to allow different mappings per environment based on additional factors 

What we found is that when using the `token -i <cluster name> -r <k8-contextual-iam-role>` we loose a bit of the fidelity of that audit record when we use the `SessionName` attribute.  When doing so, our initial `SessionName` is no longer our email address rather a generalized timestamp.  

We *can* still trace back the user by parsing the [`role id`](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable) across calls, but it requires significantly more digging and "joining" of CloudTrail and Audit logs. 

The proposed solution I came up with (couldn't find a more straightforward way of doing this) is, if the Role ARN is passed into the Token Generation process to check if our current caller identity already has a more explicit `caller-specified-role-name`.  

If so, we can set that as the SessionName of the STS session we create.

jrnt30:
Fixes #120 

When using a federated user session in conjunction with
the -r role assumption option, some of the fidelity of that
initial session is lost (ex: caller-specified-name from
                         https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable)

This will conditionally check, if a new role is asked to be assumed,
if there is already an existing role specified and if so, it will
pass it through to the new session to allow for easier auditing
of users as they traverse more complex IAM role delegation structures.

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -182,8 +182,28 @@ func (g generator) GetWithRoleForSession(clusterID string, roleARN string, sess 
 	// if a roleARN was specified, replace the STS client with one that uses
 	// temporary credentials from that role.
 	if roleARN != "" {
+		// Determine if the current session is already a federated identity.  If so,
+		// we carry through this session name onto the new session to provide better
+		// aduiting capabilities
+		//
+		// NOTE: I couldn't figure out a more `direct` way of indicating if the
+		// current Caller Identity's PrincipalType is `Federated`
+		// (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable)
+		// If there is a more direct approach I would be interested in understanding
+		resp, err := stsAPI.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+		if err != nil {
+			return "", err
+		}
+
+		userIDParts := strings.Split(*resp.UserId, ":")
+		sessionSetter := func(provider *stscreds.AssumeRoleProvider) {
+			if len(userIDParts) == 2 {
+				provider.RoleSessionName = userIDParts[1]
+			}
+		}
+
 		// create STS-based credentials that will assume the given role
-		creds := stscreds.NewCredentials(sess, roleARN)
+		creds := stscreds.NewCredentials(sess, roleARN, sessionSetter)
 
 		// create an STS API interface that uses the assumed role's temporary credentials
 		stsAPI = sts.New(sess, &aws.Config{Credentials: creds})


mattlandis:
I like the idea of providing this for convenience.  Should this be done by default or should it be an option?  Should we also provide a way to override the session name?

You call out being able to use it for auditing but I would be careful with that.  Ultimately the session name is still controlled by the client so unless where this is running is considered trusted they can always set their own session name with a custom client.

jrnt30:
I would say default is acceptable since the alternative is just a random
string.

Re the trust issue, understood and thanks. The initial session for all of
our roles only come from ADFS which controls that. There are some
additional attributes of the audit log that allow *proving* that they match
up.

micahhausler:
Hey thanks for this PR. We're not going to merge this PR because the session can be an arbitrary string, this makes that audit not trustworthy. While it could be helpful for debugging, its not something we want to suggest should be trusted for an audit.

/close

jrnt30:
The session string can be arbitrary but the other component can not.  With a completely opaque token as it is now, you *ALWAYS* have to go back to something like CloudTrail to try and associate a user with a given attribute.  I personally don't think that carrying forward the same session name attribute that you get if you aren't federated is that much less secure than the alternative.  

Would it be acceptable to map it to a configuration flag that a user has to opt into?

micahhausler:
I think an off-by-default option would be a good solution. The flag could be named something like `--forward-session-name` and stating in the help text that this forwards the caller-specified session name. Also a README note also specifying that this should not be used for access control. What we do not want to happen is for someone to rely on this for auditing or ACLs based on the session name.

/reopen

jrnt30:
Sounds fair, thanks. I'll make some adjustments

PR #122, opened 2018-07-17
Adding a flag to only output the token

### LINKED ISSUE #85: Flag for returning just the token attribute
Would anyone else find it useful to have a flag to just return the token from the `kubectl` `execcredential` flow? I've found that I'm using the `heptio-authenticator-aws token -r <role> -i <name> | jq -r ".status.token"` to generate tokens for things like the Kubernetes dashboard auth.

Curious if anyone else would use something like `--token-only` or some other name?

christopherhein:
**Why:**

* This enables you to use the authenticator binary to generate
  short-lived access to tools like the Kubernetes Dashboard using the
  token field

* closes #85

**Usecase:**

```sh
aws-iam-authenticator -i <cluster-name> --only-token 
```

### DIFF
--- a/cmd/aws-iam-authenticator/token.go
+++ b/cmd/aws-iam-authenticator/token.go
@@ -33,6 +33,7 @@ var tokenCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		roleARN := viper.GetString("role")
 		clusterID := viper.GetString("clusterID")
+		tokenOnly := viper.GetBool("tokenOnly")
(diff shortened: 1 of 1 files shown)


mattlandis:
Nice.  This will save a bunch of `| jq -r '.status.token'`.

keithlayne:
Just wanted to point out that the actual flag is `--token-only`.  Also, it would be awesome if this (and maybe k8s dashboard usage with this project) were mentioned in the readme.

PR #124, opened 2018-07-23
Adding configs to allow you to run from local

### LINKED ISSUE #123: Allow Configuration of the Authenticator from a non-localhost ENV
Currently, you cannot use `init` to process files based on a remote location, the certificates tat are generated have hardcoded values for the IP and hostname to respond to.

This makes development more difficult requiring you to rebuild the docker image and restart the pod each time there is an update.

Would be nice to have a way to set up where the server binds and what the SSL configuration is.

christopherhein:
**Why:**

* This makes development much easier allowing you to pass `--kubeconfig`
  to a local directory and have it listen to the remote apiserver
* Caveat! This needs to also be configured with the auth webhook so
  development requires an addressable server

**Issue:** closes #123

### DIFF
--- a/cmd/aws-iam-authenticator/init.go
+++ b/cmd/aws-iam-authenticator/init.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 var initCmd = &cobra.Command{
@@ -53,5 +54,17 @@ var initCmd = &cobra.Command{
 }
 
 func init() {
+	initCmd.Flags().String(
+		"hostname",
+		"localhost",
+		"Hostname that should be used for writing into the selfsigned certificates")
+	viper.BindPFlag("server.hostname", initCmd.Flags().Lookup("hostname"))
+
+	initCmd.Flags().String(
+		"address",
+		"127.0.0.1",
+		"IP Address to bind the server to listen to. (should be a 127.0.0.1 or 0.0.0.0)")
+	viper.BindPFlag("server.address", initCmd.Flags().Lookup("address"))
+
 	rootCmd.AddCommand(initCmd)
 }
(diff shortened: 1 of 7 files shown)


cmd/aws-iam-authenticator/root.go
```diff
@@ -80,9 +80,11 @@ func getConfig() (config.Config, error) {
 		ClusterID:                         viper.GetString("clusterID"),
 		ServerEC2DescribeInstancesRoleARN: viper.GetString("server.ec2DescribeInstancesRoleARN"),
 		LocalhostPort:                     viper.GetInt("server.port"),
+		Localhostname:                     viper.GetString("server.hostname"),
```
mattlandis: Does it have to be a localhost?  Can I provide another hostname?

cmd/aws-iam-authenticator/init.go
```diff
@@ -53,5 +54,18 @@ var initCmd = &cobra.Command{
...
 
 func init() {
+	initCmd.Flags().String(
+		"hostname",
+		"localhost",
+		"Hostname that should be used for writing into the selfsigned certificates")
+	viper.BindPFlag("server.hostname", initCmd.Flags().Lookup("hostname"))
+
+	initCmd.Flags().StringP(
+		"bind",
```
nckturner: Nit: call the flag `addr` or `address`.

nckturner:
Can you update your commit message?  `--kubeconfig` was part of the larger change, correct?

cmd/aws-iam-authenticator/root.go
christopherhein: No, I was following the `LocalhostPort` reference, I can change both of those if you want.

christopherhein:
@nckturner yeah, changed both the config and renamed all the components to not have a localhost specificity.

christopherhein:
Does `/lgtm` trigger a merge on this or does @mattlandis need to `/lgtm` too?

PR #126, opened 2018-08-06
feat(verify): creating output format option for verify command

joshkurz:
Wanted to be able to use the verify command and not have to parse out data from it in an unusual way.

Not sure if this is needed or not. I know this command is listed as, "Verify a token for debugging purpose" but it's actually pretty useful for determining who a token belongs too.

### DIFF
--- a/cmd/aws-iam-authenticator/verify.go
+++ b/cmd/aws-iam-authenticator/verify.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


mattlandis:
Thanks for the contribution.

We may want to look at making this a top level flag for other commands as a follow up if makes sense.

joshkurz:
Yeah I agree. Would be nicer if we had an output function that would manage all the logic. Then every command could just send to that, and based on the output flags, it would determine what to do.

PR #128, opened 2018-08-07
#125 Adding `version` subcommand

### LINKED ISSUE #125: --version flag
On a support chat with AWS they asked for the version of `heptio-authenticator-aws` and it doesn't seem to have a flag to report the version of itself. It would be helpful to have that.

christopherhein:
**Why:**

* allows you to echo the version of the build, uses built in `goreleaser` `ldflags` to get the latest release information
    
* closes #125
    
```bash
→ ./dist/darwin_amd64/aws-iam-authenticator version -s
Version: v0.3.0
Commit: 01dd27d
→ 
→ ./dist/darwin_amd64/aws-iam-authenticator version
{"Version":"v0.3.0","Commit":"01dd27d"}

```

We can customize what params are enabled by using the `build` section in the `.goreleaser.yaml` file like so:

```yaml
  build:
    # ....
    ldflags:
    - "-s -w -X main.version={{.Tag}} -X main.commit={{.Commit}}"
```

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,6 +1,9 @@
 # Configuration for https://goreleaser.com/
 project_name: authenticator
 
+git:
+  short_hash: true
+
 builds:
   - binary: aws-iam-authenticator
     main: ./cmd/aws-iam-authenticator/
@@ -12,6 +15,8 @@ builds:
       - amd64
     env:
       - CGO_ENABLED=0
+    ldflags:
+      - "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}"
 
 dockers:
   - image: gcr.io/heptio-images/authenticator
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ services:
   - docker
 
 install:
-  - curl -s -L --retry 8 -o /tmp/goreleaser.tgz https://github.com/goreleaser/goreleaser/releases/download/v0.45.1/goreleaser_Linux_x86_64.tar.gz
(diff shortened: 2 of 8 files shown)


christopherhein:
@nckturner It looks like any time I mutate the `.goreleaser.yaml` file it breaks the Travis job, even for something simple like:

```yaml
git:
  short_hash: true
```

or

```yaml
    ldflags:
      - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}}
```

Which _famous last words_ works fine locally building. Do you know if we are doing something special to that file in the travis process? Could it be version updates, I re-installed `goreleaser` and it still works normally only difference I see is I'm running:

```shell
→ go version
go version go1.10.1 darwin/amd64
```

nckturner:
Hmm, it seems like as long as these two commands still work, then the build should still work: 

```
	$(GORELEASER) --skip-publish --rm-dist --snapshot
	go test -v -cover -race $(GITHUB_REPO)/...
```

christopherhein:
That's what I expected as well @nckturner but check out this test run. Just for visibility I'm making the `goreleaser.yaml` change in a separate commit. If we can get this solved I'll squash it back down.

nckturner:
Looks like I am getting the same error: 

```
/usr/local/bin/goreleaser --skip-publish --rm-dist --snapshot

   • releasing using goreleaser 0.63.0...
   • loading config file       file=.goreleaser.yaml
   ⨯ release failed after 0.00s error=yaml: unmarshal errors:
  line 19: cannot unmarshal !!seq into string
make: *** [build] Error 1
```

Fixed with: 
```
    ldflags: "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}"
```

christopherhein:
thanks @nckturner this is all updated!

PR #134, opened 2018-08-17
Role assumption names

### LINKED ISSUE #120: Using the RoleAssumption option of the `token` cli does not persist underlying Session Name

jrnt30:
Fixes #120
See conversation from #121 for additional context/discussion

When using a federated user session in conjunction with
the -r role assumption option, some of the fidelity of that
initial session is lost (ex: caller-specified-name from
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable)

This will conditionally check, if a new role is asked to be assumed,
if there is already an existing role specified and if so, it will
pass it through to the new session to allow for easier auditing
of users as they traverse more complex IAM role delegation structures.

- [ ] Ensure proper fallback for role assumption without session forwarding

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -184,24 +184,27 @@ func (g generator) GetWithRoleForSession(clusterID string, roleARN string, sess 
 
 	// if a roleARN was specified, replace the STS client with one that uses
 	// temporary credentials from that role.
-	if roleARN != "" && g.forwardSessionName {
-		// Determine if the current session is already a federated identity.  If so,
-		// we carry through this session name onto the new session to provide better
-		// aduiting capabilities
-		//
-		// NOTE: I couldn't figure out a more `direct` way of indicating if the
-		// current Caller Identity's PrincipalType is `Federated`
-		// (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable)
-		// If there is a more direct approach I would be interested in understanding
-		resp, err := stsAPI.GetCallerIdentity(&sts.GetCallerIdentityInput{})
-		if err != nil {
-			return "", err
-		}
+	if roleARN != "" {
+		sessionSetter := func(provider *stscreds.AssumeRoleProvider) {}
+		if g.forwardSessionName {
+			// Determine if the current session is already a federated identity.  If so,
+			// we carry through this session name onto the new session to provide better
+			// aduiting capabilities
+			//
+			// NOTE: I couldn't figure out a more `direct` way of indicating if the
+			// current Caller Identity's PrincipalType is `Federated`
+			// (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable)
+			// If there is a more direct approach I would be interested in understanding
+			resp, err := stsAPI.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+			if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/token/token.go
```diff
@@ -181,9 +184,29 @@ func (g generator) GetWithRoleForSession(clusterID string, roleARN string, sess
 
 	// if a roleARN was specified, replace the STS client with one that uses
 	// temporary credentials from that role.
-	if roleARN != "" {
+	if roleARN != "" && g.forwardSessionName {
```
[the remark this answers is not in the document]
jrnt30: Great point!  Thanks.

pkg/token/token.go
```diff
@@ -182,8 +185,31 @@ func (g generator) GetWithRoleForSession(clusterID string, roleARN string, sess
 	// if a roleARN was specified, replace the STS client with one that uses
 	// temporary credentials from that role.
 	if roleARN != "" {
+		sessionSetter := func(provider *stscreds.AssumeRoleProvider) {}
+		if g.forwardSessionName {
+			// Determine if the current session is already a federated identity.  If so,
+			// we carry through this session name onto the new session to provide better
+			// aduiting capabilities
```
micahhausler: Can you change this from 'auditing' to 'debugging'?

pkg/token/token.go
```diff
@@ -182,8 +185,31 @@ func (g generator) GetWithRoleForSession(clusterID string, roleARN string, sess
...
+			//
+			// NOTE: I couldn't figure out a more `direct` way of indicating if the
```
micahhausler: Can you remove this note?

PR #135, opened 2018-08-17
Update Readme for simplified Kops 1.10 instructions

rifelpet:
[Kops 1.10](https://github.com/kubernetes/kops/releases/tag/1.10.0) was released 2 days ago which [adds support for managing the AWS IAM Authenticator](https://github.com/kubernetes/kops/pull/5197) via the cluster spec. This greatly simplifies the install instructions.

It does require Kops 1.10 and Kubernetes 1.10, so if theres a desire to maintain both sets of instructions I can update this PR to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,51 +120,7 @@ You can also omit `-r ROLE_ARN` to sign the token with your existing credentials
 This is useful if you want to authenticate as an IAM user directly or if you want to authenticate using an EC2 instance role or a federated role.
 
 ## Kops Usage
-Clusters managed by [Kops](https://github.com/kubernetes/kops) can be configured to use Authenticator.
-Both single and HA master cluster configurations are supported.
-Perform the following steps to setup Authenticator on a Kops cluster:
-1. Pre-generate the certificate, key, and kubeconfig and upload them to the kops state store.
-   ```
-   aws-iam-authenticator init -i $CLUSTER_NAME
-   aws s3 cp cert.pem ${KOPS_STATE_STORE}/${CLUSTER_NAME}/addons/authenticator/cert.pem;
-   aws s3 cp key.pem ${KOPS_STATE_STORE}/${CLUSTER_NAME}/addons/authenticator/key.pem;
-   aws s3 cp aws-iam-authenticator.kubeconfig ${KOPS_STATE_STORE}/${CLUSTER_NAME}/addons/authenticator/kubeconfig.yaml;
-   ```
-2. Add the following sections to the cluster spec, either using `kops edit cluster ${CLUSTER_NAME}` or editing the manifest yaml file.
-   Be sure to replace `KOPS_STATE_STORE` and `CLUSTER_NAME` with their appropriate values since those environment variables are not available at runtime.
-   This downloads the files from the state store on masters to a directory that is volume mounted by kube-apiserver.
-   Kops does not support adding additional volumes to kube-apiserver so we must reuse the existing `/srv/kubernetes` hostPath volume.
-   ```
-   apiVersion: kops/v1alpha2
-   kind: Cluster
-   spec:
-     kubeAPIServer:
-       authenticationTokenWebhookConfigFile: /srv/kubernetes/aws-iam-authenticator/kubeconfig.yaml
-     hooks:
-     - name: kops-hook-authenticator-config.service
-       before:
-         - kubelet.service
-       roles: [Master]
-       manifest: |
-         [Unit]
-         Description=Download AWS Authenticator configs from S3
-         [Service]
-         Type=oneshot
-         ExecStart=/bin/mkdir -p /srv/kubernetes/aws-iam-authenticator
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -122,46 +122,20 @@ This is useful if you want to authenticate as an IAM user directly or if you wan
...
-  ```
-  ExecStart=/usr/bin/docker run --net=host --rm -v /srv/kubernetes/aws-iam-authenticator:/srv/kubernetes/aws-iam-authenticator quay.io/coreos/awscli@sha256:7b893bfb22ac582587798b011024f40871cd7424b9026595fd99c2b69492791d aws s3 cp --recursive s3://KOPS_STATE_STORE/CLUSTER_NAME/addons/authenticator /srv/kubernetes/aws-iam-authenticator/
-  ```
-3. Apply the changes with `kops update cluster ${CLUSTER_NAME}`.
+2. Apply the changes with `kops update cluster ${CLUSTER_NAME}`.
    If the cluster already exists, roll the cluster with `kops rolling-update cluster ${CLUSTER_NAME}` in order to recreate the master nodes.
-4. Update the Authenticator DaemonSet's state and output volumes to both use `/srv/kubernetes/aws-iam-authenticator/` for their `hostPath`s.
-5. Apply the DaemonSet and ConfigMap resource manifests to launch the Authenticator server on the cluster.
 
 *Note:* Certain Kops commands will overwrite the `ExecCredential` in kubeconfig so it may need to be restored manually. See [kubernetes/kops#5051](https://github.com/kubernetes/kops/issues/5051) for more information.
```
christopherhein: We should probably change this to say something like "If you are using Kops <1.10 …" instead of "Certain Kops" since the 1.10 Kops fixes this bug by supporting `client-go` 7.0

rifelpet: It's kind of strange to provide instructions that require >=1.10 and then give a warning about behavior when using <1.10. Maybe we just remove the warning altogether if we're not going to provide instructions for <1.10 ?

christopherhein: That's a fair point. Removing is probably the better option.

rifelpet:
Updated to remove outdated note

dhemeier:
Hey @rifelpet ,

I had currently tested to change to aws-authenticator on one of my clusters. You don't need to do a `rolling-update` on existing clusters anymore. The `aws-iam-authenticator` deployed directly after the `kops update cluster --yes`. So you could also remove the rolling-update task from the description.

dhemeier:
Sorry this is not correct, 

all deployments and folders are created on the master, but the certificates are only inserted on instance launch. So you need the `rolling-update`. 

But then Kops have a problem, they don't display a "needs update" message, so you can only deploy the master again with `kops rolling-update cluster --instance-group master-XX-XX-XX --force --yes`

Sorry for my mistake.

rifelpet:
@dhemeier Thanks for the heads up, I've updated the `rolling-update` command to include `--force`, and added `--yes` to both `update` and `rolling-update`.

README.md
```diff
@@ -122,49 +122,20 @@ This is useful if you want to authenticate as an IAM user directly or if you wan
...
-  ```
-3. Apply the changes with `kops update cluster ${CLUSTER_NAME}`.
-   If the cluster already exists, roll the cluster with `kops rolling-update cluster ${CLUSTER_NAME}` in order to recreate the master nodes.
-4. Update the Authenticator DaemonSet's state and output volumes to both use `/srv/kubernetes/aws-iam-authenticator/` for their `hostPath`s.
-5. Apply the DaemonSet and ConfigMap resource manifests to launch the Authenticator server on the cluster.
-
-*Note:* Certain Kops commands will overwrite the `ExecCredential` in kubeconfig so it may need to be restored manually. See [kubernetes/kops#5051](https://github.com/kubernetes/kops/issues/5051) for more information.
 
+2. Apply the changes with `kops update cluster ${CLUSTER_NAME} --yes`.
+   If the cluster already exists, roll the cluster with `kops rolling-update cluster ${CLUSTER_NAME} --force --yes` in order to recreate the master nodes.
```
somcsel: If only the master nodes need to be rolled, then use `kops rolling-update cluster ${CLUSTER_NAME} --instance-group ${MASTER_INSTANCE_GROUP_NAME} --force --yes` to prevent unnecessary recycling of the nodes.

somcsel:
According to the [kops docs](https://github.com/kubernetes/kops/blob/master/docs/authentication.md#aws-iam-authenticator) a ConfigMap also needs to be created.  Without this, the cluster will not validate.

somcsel:
Perhaps it would be better to maintain one set of authoritative instructions in the kops git and just link to them from this README?

README.md
Globegitter: More generically one could also just do `kops rolling-update cluster ${CLUSTER_NAME} --instance-group-roles=Master --force --yes`

rifelpet:
@somcsel Agreed. I've updated this PR to just point to the kops documentation. I added some additional instructions over there so hopefully it should be sufficient on its own.

nckturner:
@christopherhein is this ok to merge?

christopherhein:
/lgtm

@nckturner I'm good with these updates.

rifelpet:
Is there any way we can get this merged? CC @nckturner

PR #138, opened 2018-08-24
Fix format string of ListenUrl in config

bfrydl:
This currently causes the `init` command to output e.g. `https://%!d(string=localhost):21362/authenticate`.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -28,7 +28,7 @@ func (c *Config) ListenURL() string {
 }
 
 func (c *Config) ListenAddr() string {
-	return fmt.Sprintf("%d:%d", c.Hostname, c.HostPort)
+	return fmt.Sprintf("%s:%d", c.Hostname, c.HostPort)
 }
 
 func (c *Config) GenerateFiles() error {


bfrydl:
Oh geez there's a whole process to contributing, but it's only one letter.

christopherhein:
Thanks for filing and fixing this @bfrydl are you able to sign that CLA?

bfrydl:
I don't really want to sign up for an account on that site and sign a CLA just to fix one letter. Maybe an issue would have been a more appropriate way of addressing this.

To be clear though, as minor as this change is, if someone follows the directions in the README to install aws-iam-authenticator via `go get`, pregenerate the config, and then add it to the kops cluster spec hooks, everything will seem fine but the masters will not come online because of the invalid config. With a new cluster you're left with nothing, not even an API server, and must debug this manually on one of the failing masters.

christopherhein:
@bfrydl alright, I think this was a bug that came out of one of my additions. I'm happy to make the change since I've already signed the CLA. Just wanted to check with you before doing so.

Thank you for finding the bug if you'd like to close this PR I will reopen one.

bfrydl:
Sure, I'll close this one.

christopherhein:
Actually I just noticed this is already fixed on `master` the last commit to that file fixed this - https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/7b75f057bf4f23fe77f62c02cc5a6290ea6f5e23#diff-091ace02b49a46f55f2586cfec89a160

christopherhein:
Thanks again for noticing and trying to fix this @bfrydl

PR #143, opened 2018-09-04
Automatically reload configuration if changed

bailantaotao:
HI there,

This PR major feature is reload configuration.

I provide the following flag for `server`
`--reload-enabled` enables dynamic config reload.
`--reload-period` specifies how often the files are checked for changes. 

It's not supported by below config
```
+ defaultRole
+ server.generateKubeconfig
+ server.stateDir
```

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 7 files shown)


pkg/token/token.go
```diff
@@ -226,25 +228,37 @@ func (g generator) FormatJSON(token string) string {
 // Verifier validates tokens by calling STS and returning the associated identity.
 type Verifier interface {
 	Verify(token string) (*Identity, error)
+	ReloadConfig(c config.Config)
 }
 
 type tokenVerifier struct {
+	lock      sync.Mutex
```
mattlandis: Could we use a RWMutext here?  I think there will be a lot more reads than writes.

bailantaotao: No problem.

bailantaotao: And i see more lock is using `sync.Mutex`, do we need to change? or create another PR to fix it?

https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/143/files#diff-09856fe9becddf0199651f451409356aR398

PR #144, opened 2018-09-04
Lookup assumed role ARN with IAM API

jpb:
Previously, `arn.Canonicalize` assumed that the full IAM role ARN could be extracted from an assumed-role ARN. This is not the case for IAM roles with a path, in which case only the role name is encoded in the assumed-role ARN. This change uses the IAM API to determine the correct IAM role ARN for a given assumed-role ARN.

The `aws` package was (initially) added to resolve circular dependencies between `server` and `arn`, but it turns out more than `server` needs use of the newly introduced `iamProvider`. `ec2Provider` was moved to this package to keep things symmetrical.

This is a potential proper fix for https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/98 / https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/103 (see https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/103#issuecomment-417124170)


jpb:
Hey @nckturner are you able to take a look at this? Or is there someone else you would suggest to have review this PR? Thanks!

nckturner:
@jpb Apologies, I have been out of the office for a bit.  Taking a look!

nckturner:
@jpb Quick update--I engaged some folks in the AWS security org to discuss the best way to handle paths in roles moving forward.  I will keep this issue updated.  

I totally agree that the current behavior leaves something to be desired.  The role ARN _does_ have the path in it, whereas the assumed-role ARN does not.  The current behavior basically forces you to construct an ARN for the role that does not include the path to put in your map.

jpb:
Thanks for following up @nckturner. 

If it helps, [the documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_identityandaccessmanagement.html#identityandaccessmanagement-resources-for-iam-policies) specifies that IAM roles include the path in their ARN (`arn:${Partition}:iam::${Account}:role/${RoleNameWithPath}`) but assumed roles do not (`arn:${Partition}:iam::${Account}:assumed-role/${RoleName}/${RoleSessionName}`). This works out because roles are uniquely identified by their name, not by their name and path. The best documentation I could find on that is that [GetRole does not accept the path as a parameter](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetRole.html) when fetching a role (a path will always have a slash in it, which is not valid in the `RoleName` parameter, nor is there a "path" parameter that can be specified for that API endpoint).

alex-berger:
Any progress on this PR, @nckturner any update about this from the AWS security folks?

romil-punetha:
Can someone suggest me whats wrong with the following aws-auth configmap entry?
 ```mapRoles: | 
   - groups:
      - curefit:eks-admin
      rolearn: arn:aws:iam::000000000:role/eks-admins
      username: admin:{{SessionName}}
    - groups:
      - curefit:eks-viewer
      rolearn: arn:aws:iam::000000000:role/eks-all-cluster-viewers-role
      username: viewer:{{SessionName}}```
```
This role is associated with a policy, that is attached to a group, and no one from that group is able to access the cluster.

alex-berger:
@romil-punetha I guess this is due to https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/114.

romil-punetha:
@alex-berger Im using `rolearn` which is what the fix says.

PR #160, opened 2018-10-13
Add expiration to token creation

### LINKED ISSUE #133: heptio-authenticator occassionally returns Unauthorized when used in a kubernetes go client
Issue:

I have a sample HTTP server that listens on `/`. When `/` is called, the handler simply authenticates with eks cluster using the Heptio-authenticator-module and returns to the user the number of pods running in that cluster.For authentication I use the kubernetes go client.
Also note that for every new request on `/` I create a new kuberentes config so there is no config reuse between requests.

I observed that after ~15 min the eks authentication fails with a `Unauthorized` error which is unexpected as I am creating generating a new kuberenets config object for every request which should result in a token generated for every request.
Also the immediate request to `/` (after unauthorized request) seems to be successful again which is more confusing.

Attached is a snippet of my code and logs

1. First request: success
curl -v -s localhost:9000/
log output: 
```
There are 2 pods in the cluster
```

2. Second request: success
curl -v -s localhost:9000/
log output: 
```
There are 2 pods in the cluster
```

3. Third request after ~15 min: failed
curl -v -s localhost:9000/
log output: 
```
2018/08/10 14:10:27 http: panic serving [::1]:63678: Unauthorized
goroutine 11 [running]:
```

4. Fourth request: success
curl -v -s localhost:9000/
log output: 
```
There are 2 pods in the cluster
```
```
package main

import (
  //"flag"
  "fmt"
  //"path/filepath"
  //"time"
  metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

  //"k8s.io/apimachinery/pkg/api/errors"
  "k8s.io/client-go/kubernetes"
  //_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
  restclient "k8s.io/client-go/rest"
  capi "k8s.io/client-go/tools/clientcmd/api"
  "net/http"
)

var (
  endpoint string = "REDACTED"
)

func getClientSet() *kubernetes.Clientset {
  awsenv := capi.ExecEnvVar{
    Name:  "AWS_PROFILE",
    Value: "dv",
  }
  config := &restclient.Config{
    Host: endpoint,
    TLSClientConfig: restclient.TLSClientConfig{
      //CAData: clusterCert,
      CAFile: "REDACTED",
    },
    ExecProvider: &capi.ExecConfig{
      Command:    "heptio-authenticator-aws",
      Args:       []string{"token", "-i", "dev-v1"},
      APIVersion: "client.authentication.k8s.io/v1alpha1",
      Env:        []capi.ExecEnvVar{awsenv},
    },
  }
  clientset, err := kubernetes.NewForConfig(config)
  if err != nil {
    panic(err.Error())
  }
  return clientset
}

func getClusterCredentials(w http.ResponseWriter, r *http.Request) {
  clientset := getClientSet()
  fmt.Println(clientset)
  pods, err := clientset.CoreV1().Pods("local").List(metav1.ListOptions{})
  if err != nil {
    panic(err.Error())
  }
  fmt.Printf("There are %d pods in the cluster\n", len(pods.Items))
  fmt.Fprintf(w, "There are %d pods in the cluster\n", len(pods.Items))
}

func main() {
  http.HandleFunc("/", getClusterCredentials)
  http.ListenAndServe(":9000", nil)

  
}
```

nckturner:
STS presigned urls expire after 15 minutes, and are not configurable,
so we refresh tokens after 14 minutes to prevent getting occasional
unauthorized responses.

Fixes #133

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -69,18 +69,22 @@ const (
 	// signed, but we set this unused parameter to 60 for legacy reasons (we check for a value between 0 and 60 on the
 	// server side in 0.3.0 or earlier).  IT IS IGNORED.  If we can get STS to support x-amz-expires, then we should
 	// set this parameter to the actual expiration, and make it configurable.
-	tokenExpirationParam = 60
+	requestPresignParam = 60
 	// The actual token expiration (presigned STS urls are valid for 15 minutes after timestamp in x-amz-date).
-	tokenExpiration = 15 * time.Minute
-	// Refresh token after 14 minutes for some cushion.
-	refreshTokenAfter = 14 * time.Minute
-	v1Prefix          = "k8s-aws-v1."
-	maxTokenLenBytes  = 1024 * 4
-	clusterIDHeader   = "x-k8s-aws-id"
+	presignedURLExpiration = 15 * time.Minute
+	v1Prefix               = "k8s-aws-v1."
+	maxTokenLenBytes       = 1024 * 4
+	clusterIDHeader        = "x-k8s-aws-id"
 	// Format of the X-Amz-Date header used for expiration
 	// https://golang.org/pkg/time/#pkg-constants
-	dateHeaderFormat = "20060102T030405Z"
+	dateHeaderFormat = "20060102T150405Z"
 )
+
+// Token is generated and used by Kubernetes client-go to authenticate with a Kubernetes cluster.
+type Token struct {
+	Token      string
+	Expiration time.Time
+}
 
 // FormatError is returned when there is a problem with token that is
 // an encoded sts request.  This can include the url, data, action or anything
@@ -137,15 +141,15 @@ type getCallerIdentityWrapper struct {
 // Generator provides new tokens for the heptio authenticator.
 type Generator interface {
 	// Get a token using credentials in the default credentials chain.
-	Get(string) (string, error)
+	Get(string) (Token, error)
 	// GetWithRole creates a token by assuming the provided role, using the credentials in the default chain.
-	GetWithRole(clusterID, roleARN string) (string, error)
+	GetWithRole(clusterID, roleARN string) (Token, error)
 	// GetWithRoleForSession creates a token by assuming the provided role, using the provided session.
-	GetWithRoleForSession(clusterID string, roleARN string, sess *session.Session) (string, error)
+	GetWithRoleForSession(clusterID string, roleARN string, sess *session.Session) (Token, error)
 	// GetWithSTS returns a token valid for clusterID using the given STS client.
-	GetWithSTS(clusterID string, stsAPI *sts.STS) (string, error)
+	GetWithSTS(clusterID string, stsAPI *sts.STS) (Token, error)
 	// FormatJSON returns the client auth formatted json for the ExecCredential auth
-	FormatJSON(string) string
+	FormatJSON(Token) string
 }
 
 type generator struct {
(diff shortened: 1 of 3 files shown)


nckturner:
/hold

Going to change the value of expiration parameter to Presign() back to 60, since changing it to 900 (15 minutes) breaks backwards compatibility of tokens, and it doesn't have any effect :D I think its reasonable to increase the range of accepted values for the x-amz-expires header, in case we want to push STS to support it in the future.

nckturner:
I set the value of the Presign() parameter back to 60, but left the accepted values on the server side from 0-15 minutes.  Based on what I've heard, people would use a longer token expiration than 15 minutes, so if STS is willing to support x-amz-expires at some point in the future, then we should decide what the longest expiration we are willing to allow.  That can be done in a future change.

pkg/token/token.go
```diff
@@ -231,13 +248,15 @@ func (g generator) GetWithSTS(clusterID string, stsAPI *sts.STS) (string, error)
...
 func (g generator) FormatJSON(token string) string {
+	expirationTimestamp := metav1.NewTime(time.Now().Local().Add(refreshTokenAfter))
 	execInput := &clientauthv1alpha1.ExecCredential{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "client.authentication.k8s.io/v1alpha1",
 			Kind:       "ExecCredential",
 		},
 		Status: &clientauthv1alpha1.ExecCredentialStatus{
-			Token: token,
+			ExpirationTimestamp: &expirationTimestamp,
```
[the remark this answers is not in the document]
nckturner: What happens is the FormatJSON is executed each time a token is generated, and the response is cached in client-go.  This will solve the issue where client-go is running continuously, like in the sample program provided in the issue, or in kubelet, etc.  For kubectl, we need client caching in a file, for which there is currently another PR open.  I did refactor a little based on your comment (no functionality change) where its now generating the expiration next to the place where the token is generated.  The expiration of the token is based on the `x-amz-date` header, so I think I'll try deriving the expiration from that instead, which is technically correct, though in practice I doubt there would be a noticeable difference from the two timestamps.

pkg/token/token.go
```diff
@@ -65,11 +65,26 @@ type Identity struct {
...
+	presignedURLExpiration = 15 * time.Minute
+	v1Prefix               = "k8s-aws-v1."
+	maxTokenLenBytes       = 1024 * 4
+	clusterIDHeader        = "x-k8s-aws-id"
+	// Format of the X-Amz-Date header used for expiration
+	// https://golang.org/pkg/time/#pkg-constants
+	dateHeaderFormat = "20060102T030405Z"
 )
 
+type Token struct {
```
mattlandis: Comment describing the exported struct.

pkg/token/token.go
```diff
@@ -125,15 +140,15 @@ type getCallerIdentityWrapper struct {
 // Generator provides new tokens for the heptio authenticator.
 type Generator interface {
 	// Get a token using credentials in the default credentials chain.
-	Get(string) (string, error)
+	Get(string) (*Token, error)
```
mattlandis: I would prefer returning an empty struct to nil since we don't use it as a pointer (modify contents etc.)

nakulpathak3:
hey I know this went in but I'm getting `403 Forbidden "could not get token: AccessDenied: Access denied" error` at random times still. I'm using the Kubernetes Python Client which implemented the usage of aws-iam-authenticator here - https://github.com/kubernetes-client/python-base/blob/master/config/exec_provider.py#L77. This is happening after creating a role and using the `-r` flag not with normal user accounts.

mlafeldt:
Verified that this is working with `kubectl proxy`. Good job!

PR #169, opened 2018-10-23
Update example.yaml for Kops Usage (#1)

goodfoo:
* Update example.yaml for Kops Usage

Add a comment for Kops usage
- --kubeconfig-pregenerated=true


christopherhein:
thanks @goodfoo great to have these updated!

PR #173, opened 2018-11-08
Support Global Different Region STS Endpoints

sunfuze:
* using regexp to test if host is valid

[sts endpoints of different regions](http://docs.amazonaws.cn/en_us/general/latest/gr/rande.html#sts_region)


andromedarabbit:
Any further news ?

nckturner:
The regex looks good to me, but I'd like to see some added tests for it.

PR #178, opened 2018-12-01
Include aws request ID when logging errors

doublerainbo:
A small change here to the error logging for the authenticator. When we were triaging some errors (turned out to be ratelimit, but that's a tangent), our AWS account managers and the support staff insisted on having request IDs to go with it so that they can look into it on their end. With this small change we were able to give some data points for them to investigate more for us and I hope that if others are in the same boat, this can be of use

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -35,6 +35,7 @@ import (
 	"github.com/kubernetes-sigs/aws-iam-authenticator/pkg/arn"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientauthv1alpha1 "k8s.io/client-go/pkg/apis/clientauthentication/v1alpha1"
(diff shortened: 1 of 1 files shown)


nckturner:
Tested, output looks good:

```
WARN[2019-04-27T20:00:09Z] access denied                                 client="127.0.0.1:55340" error="sts getCallerIdentity failed: error from AWS (expected 200, got 403). Body: {\"Error\":{\"Code\":\"SignatureDoesNotMatch\",\"Message\":\"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.\",\"Type\":\"Sender\"},\"RequestId\":\"0f87f54b-6927-11e9-852b-637b5ba836b4\"}" method=POST path=/authenticate
```

/lgtm
/approve

PR #184, opened 2018-12-22
Add support for multiple regions

monder:
When resolving `{{EC2PrivateDNSName}}`, authenticator looks for instances only in the current region. This PR adds an optional setting allowing to specify other regions to query.

This allows to have single cluster spanning multiple AWS regions that are connected using peered VPC.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/aws-iam-authenticator/pkg/token"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
@@ -82,6 +83,7 @@ type handler struct {
 	verifier         token.Verifier
 	metrics          metrics
 	ec2Provider      EC2Provider
+	regions          []string
 	featureGates     featuregate.MutableFeatureGate
 }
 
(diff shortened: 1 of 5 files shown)


christopherhein:
@monder can you update the conflicts?

monder:
@christopherhein done. Rebased and tested.
I've added additional [wrapper](https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/184/files#diff-09856fe9becddf0199651f451409356aR431) to keep aws abstraction introduced in https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/226

nckturner:
This allows ec2 instances to pass validation if they originate from different regions, another related consideration is whether what STS endpoint the tokens they craft use (in the presigned URL).  We should be moving away from the global STS endpoint, so the question is how does the client decide which region to choose.  

/cc @wongma7

pkg/server/server.go
```diff
@@ -484,29 +489,66 @@ func (p *ec2ProviderImpl) setPrivateDNSNameCache(id string, privateDNSName strin
...
+	return nil, nil
+}
+
+func getEC2Service(roleARN string, region *string) ec2iface.EC2API {
+	config := aws.NewConfig()
+	if region != nil {
+		config = config.WithRegion(aws.StringValue(region))
 	}
-	return privateDNSName, nil
+	return ec2.New(newSession(roleARN), config)
```
wongma7: Now we will create a new session every time we do a lookup. is there any downside to this?

pkg/server/server.go
```diff
@@ -484,29 +489,66 @@ func (p *ec2ProviderImpl) setPrivateDNSNameCache(id string, privateDNSName strin
...
+}
+
+func (p *ec2ProviderImpl) findInstanceByID(id string, region *string) (*ec2.Instance, error) {
+	output, err := p.getEC2Service(p.roleARN, region).DescribeInstances(&ec2.DescribeInstancesInput{
 		InstanceIds: aws.StringSlice([]string{id}),
 	})
 	if err != nil {
-		return "", fmt.Errorf("failed querying private DNS from EC2 API for node %s: %s", id, err.Error())
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "InvalidInstanceID.NotFound" {
+			// "not found" is not an error in this case
```
wongma7: Don't like this pattern of returning `nil, nil`, would prefer if it defined an error string/type and had the caller check it. `findInstanceByID` sounds to me like it should either find the instance or give me an error, not neither.

PR #185, opened 2018-12-25
Fix UTC time conversion bug.

llamahunter:
Golang has a weird time formatting system.  It does not do time
zone conversion for you.  If you want to output UTC, you must first
convert your time to UTC.  Previously, this test would only work on
a machine that happened to be in UTC.


llamahunter:
Could this get pulled? thx.

Issue #183, opened 2018-12-21
Cache credentials from credentials_process


llamahunter:
Once pull #182 is accepted and issue #141 is closed, the next problem is to have aws-iam-authenticator cache temporary SSO credentials.  Per https://github.com/aws/aws-sdk-go/issues/1329, this needs to be implemented by the *user* of aws-sdk-go (for reasons not entirely clear to me).

llamahunter:
I have a working implementation of this.  pull request coming soon.

llamahunter:
Waiting on https://github.com/aws/aws-sdk-go/pull/2375 getting accepted, so that I can create my pull request based on it for aws-iam-authenticator.

Issue #151, opened 2018-09-24
EKS Connections using role switching with MFA ask everytime for MFA token


daraacca:
I wish to connect to an EKS Instance using kubectl with the authenticator. To connect to this instance I want to assume a role in the necessary AWS Account where EKS Lives.    The config map of the EKS Instance has also been updated to specify the same role -

I have the following kube config 

```
apiVersion: v1
clusters:
- cluster:
    server: https://xxxxxxx.sk1.us-east-1.eks.amazonaws.com
    certificate-authority-data: xxxxxxxxxxxxxxx
  name: kubernetes
contexts:
- context:
    cluster: kubernetes
    user: aws
  name: aws
current-context: aws
kind: Config
preferences: {}
users:
- name: aws
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      command: aws-iam-authenticator
      env:
      - name: "AWS_PROFILE"
        value: "playground-power-user"
      args:
        - "token"
        - "-i"
        - "test-cluster"
```

my AWS config file is as follows

```
[profile playground-power-user]
region = us-east-1
source_profile = main
role_arn = arn:aws:iam::12345678:role/power-user
mfa_serial = arn:aws:iam::112233445566:mfa/user@company.com
```

my aws-auth-cm.yaml  config map is as follows

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: aws-auth
  namespace: kube-system
data:
  mapRoles: |
    - rolearn: arn:aws:iam::12345678:role/eks-cluster-role
      username: system:node:{{EC2PrivateDNSName}}
      groups:
        - system:bootstrappers
        - system:nodes
    - rolearn: arn:aws:iam::12345678:role/power-user
      username: admin:{{AccountID}}:{{SessionName}}
      groups:
        - system:masters
```

This is all good. When i connect to EKS, I am asked for my MFA token, which is accepted and command is executed

>> kubectl get clusterrolebindings
Assume Role MFA token code: 275230
NAME                                                   AGE
aws-node                                               7d
cluster-admin                                          7d
eks:kube-proxy                                         7d
eks:node-bootstrapper                                  7d

HOWEVER, for every kubectl command, I have to give the MFA token again.  This is obviously unworkable. Any ideas to avoid this???

jukkafor:
[...]

Issue #99, opened 2018-06-15
Reuse tokens for up to 15 minutes - Wrapper script for client


roffe:
Hi gang and thanks for the hard work!

Got the authenticator running today and everything seems really nice!

However i noticed that all my kubectl commands now takes a little longer to complete. Not much but enough for me to hack together this little proof of concept "wrapper/cache" for the authenticator that stores the last response and would cat it directly from disk if the file is <800 seconds old.

```
#!/usr/bin/env bash
# Only tested on OS X
CACHE_FILE=${HOME}/.kube/heptio-authenticator-${AWS_PROFILE}.cache
MAXTIME=800
if [ -f $CACHE_FILE ]; then
    TS_DB=$(stat -t "%s" ${CACHE_FILE} | cut -d' ' -f10 | tr -d '"')
    AGE=$(( `date +%s` - $TS_DB ))
    if [[ $AGE -le $MAXTIME ]]; then
        cat ${CACHE_FILE}
    else
        heptio-authenticator-aws "$@" | tee $CACHE_FILE
    fi
else
    heptio-authenticator-aws "$@" | tee $CACHE_FILE
fi
```

Would something similar be doable directly in the client and let a user pass for example  `--cache` as a runtime opt to use the disk cache for a slighter faster response

christopherhein:
Totally feasible, it'd still need the file to recall for the cache but we could probably add a function to check for this file and its age after https://github.com/heptio/authenticator/blob/master/cmd/heptio-authenticator-aws/token.go#L45 and if it passed add a func to `token{}` interface to `GetFromCache()`

christopherhein:
@roffe would you like to give this a go and work on a PR?

roffe:
Sure, i'll take a crack at it and see what i can come up with

jrnt30:
I think we would also need to adjust the [pre-signed TTL that the `token` creation generates](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go#L202) as well for it to work on the server side as well, wouldn't we?

solacens:
A improved script is written in bash

Check [here](https://gist.github.com/solacens/d4eaca06c3c0a4f7912ec1ec47175395).

PR #193, opened 2019-01-11
Cache aws credentials

llamahunter:
Fixes https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/183

Possibly fixes or addresses the following in an alternate way: https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/151, https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/99, https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/63, https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/140

### DIFF
--- a/cmd/aws-iam-authenticator/token.go
+++ b/cmd/aws-iam-authenticator/token.go
@@ -35,6 +35,7 @@ var tokenCmd = &cobra.Command{
 		clusterID := viper.GetString("clusterID")
 		tokenOnly := viper.GetBool("tokenOnly")
 		forwardSessionName := viper.GetBool("forwardSessionName")
+		cache := viper.GetBool("cache")
 
 		if clusterID == "" {
 			fmt.Fprintf(os.Stderr, "Error: cluster ID not specified\n")
@@ -45,7 +46,7 @@ var tokenCmd = &cobra.Command{
 		var tok token.Token
 		var out string
 		var err error
-		gen, err := token.NewGenerator(forwardSessionName)
+		gen, err := token.NewGenerator(forwardSessionName, cache)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "could not get token: %v\n", err)
 			os.Exit(1)
@@ -77,8 +78,10 @@ func init() {
 	tokenCmd.Flags().Bool("forward-session-name",
 		false,
 		"Enable mapping a federated sessions caller-specified-role-name attribute onto newly assumed sessions. NOTE: Only applicable when a new role is requested via --role")
+	tokenCmd.Flags().Bool("cache", false, "Cache the token on disk until it expires")
 	viper.BindPFlag("role", tokenCmd.Flags().Lookup("role"))
 	viper.BindPFlag("tokenOnly", tokenCmd.Flags().Lookup("token-only"))
 	viper.BindPFlag("forwardSessionName", tokenCmd.Flags().Lookup("forward-session-name"))
+	viper.BindPFlag("cache", tokenCmd.Flags().Lookup("cache"))
 	viper.BindEnv("role", "DEFAULT_ROLE")
 }
--- a/cmd/aws-iam-authenticator/token.go
+++ b/cmd/aws-iam-authenticator/token.go
@@ -78,7 +78,7 @@ func init() {
 	tokenCmd.Flags().Bool("forward-session-name",
 		false,
 		"Enable mapping a federated sessions caller-specified-role-name attribute onto newly assumed sessions. NOTE: Only applicable when a new role is requested via --role")
-	tokenCmd.Flags().Bool("cache", false, "Cache the token on disk until it expires")
+	tokenCmd.Flags().Bool("cache", false, "Cache the credential on disk until it expires. Uses the aws profile specified by AWS_PROFILE or the default profile.")
 	viper.BindPFlag("role", tokenCmd.Flags().Lookup("role"))
 	viper.BindPFlag("tokenOnly", tokenCmd.Flags().Lookup("token-only"))
 	viper.BindPFlag("forwardSessionName", tokenCmd.Flags().Lookup("forward-session-name"))
--- a/pkg/token/filecache.go
+++ b/pkg/token/filecache.go
@@ -1,0 +1,277 @@
+package token
+
+import (
+	"context"
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/gofrs/flock"
+	"gopkg.in/yaml.v2"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"time"
+)
+
+// A mockable filesystem interface
+var f filesystem = osFS{}
+
+type filesystem interface {
+	Stat(filename string) (os.FileInfo, error)
+	ReadFile(filename string) ([]byte, error)
+	WriteFile(filename string, data []byte, perm os.FileMode) error
+}
+
+// default os based implementation
+type osFS struct {}
+
+func (osFS) Stat(filename string) (os.FileInfo, error) {
+	return os.Stat(filename)
+}
+
+func (osFS) ReadFile(filename string) ([]byte, error) {
+	return ioutil.ReadFile(filename)
+}
+
+func (osFS) WriteFile(filename string, data []byte, perm os.FileMode) error {
+	return ioutil.WriteFile(filename, data, perm)
+}
+
+// A mockable environment interface
+var e environment = osEnv{}
+
+type environment interface {
+	Getenv(key string) string
+	LookupEnv(key string) (string, bool)
+}
+
+// default os based implementation
+type osEnv struct {}
+
+func (osEnv) Getenv(key string) string {
+	return os.Getenv(key)
+}
+
+func (osEnv) LookupEnv(key string) (string, bool) {
+	return os.LookupEnv(key)
+}
+
+// A mockable flock interface
+type filelock interface {
+	Unlock() error
+	TryLockContext(ctx context.Context, retryDelay time.Duration) (bool, error)
+	TryRLockContext(ctx context.Context, retryDelay time.Duration) (bool, error)
+}
+
+var newFlock = func(filename string) filelock {
+	return flock.New(filename)
+}
+
+// cacheFile is a map of clusterID/roleARNs to cached tokens
+type cacheFile struct {
+	// a map of clusterIDs/profiles/roleARNs to cachedTokens
+	ClusterMap map[string]map[string]map[string]cacheToken `yaml:"clusters"`
+}
+
+// cacheToken is a single cached token entry, along with expiration time
+type cacheToken struct {
+	Token      credentials.Value
+	Expiration time.Time
+	// If set will be used by IsExpired to determine the current time.
+	// Defaults to time.Now if CurrentTime is not set.  Available for testing
+	// to be able to mock out the current time.
+	currentTime func() time.Time
+}
+
+// IsExpired determines if the cached token has expired
+func (t *cacheToken) IsExpired() bool {
+	curTime := t.currentTime
+	if curTime == nil {
+		curTime = time.Now
+	}
+	return t.Expiration.Before(curTime())
+}
+
+// readCacheWhileLocked reads the contents of the token cache and returns the
+// parsed yaml as a cacheFile object.  This method must be called while a shared
+// lock is held on the filename.
+func readCacheWhileLocked(filename string) (cache cacheFile, err error) {
+	cache = cacheFile{
+		map[string]map[string]map[string]cacheToken{},
+	}
+	data, err := f.ReadFile(filename)
+	if err != nil {
+		err = fmt.Errorf("unable to open file %s: %v", filename, err)
+		return
+	}
+
+	err = yaml.Unmarshal(data, &cache)
+	if err != nil {
+		err = fmt.Errorf("unable to parse file %s: %v", filename, err)
+	}
+	return
+}
+
+// writeCacheWhileLocked writes the contents of the token cache using the
+// yaml marshaled form of the passed cacheFile object.  This method must be
+// called while an exclusive lock is held on the filename.
+func writeCacheWhileLocked(filename string, cache cacheFile) error {
+	data, err := yaml.Marshal(cache)
+	if err == nil {
+		// write privately owned by the user
+		err = f.WriteFile(filename, data, 0600)
+	}
+	return err
+}
+
+// FileCacheProvider is a Provider implementation that wraps an underlying Provider
+// (contained in Credentials) and provides caching support for tokens for the
+// specified clusterID and roleARN
+type FileCacheProvider struct {
+	credentials                 *credentials.Credentials // the underlying implementation that has the *real* Provider
+	clusterID, profile, roleARN string                   // parameters used to create Provider
+	cachedToken                 cacheToken               // the cached token, if it exists
+}
+
+// NewFileCacheProvider creates a new Provider implementation that wraps a provided Credentials,
+// and works with an on disk cache to speed up token usage when the cached copy is not expired.
+// If there are any problems accessing or initializing the cache, an error will be returned, and
+// callers should just use the existing credentials provider.
+func NewFileCacheProvider(clusterID, profile, roleARN string, creds *credentials.Credentials) (FileCacheProvider, error) {
+	filename := CacheFilename()
+	cachedToken := cacheToken{}
+	if info, err := f.Stat(filename); !os.IsNotExist(err) {
+		if info.Mode() & 0077 != 0 {
+			// cache file has secret credentials and should only be accessible to the user, refuse to use it.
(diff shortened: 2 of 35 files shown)


pkg/token/token.go
```diff
@@ -188,6 +191,21 @@ func (g generator) GetWithRole(clusterID string, roleARN string) (Token, error)
...
+	if g.cache {
+		// figure out what profile we're using
+		var profile string;
+		if v := os.Getenv("AWS_PROFILE"); len(v) > 0 {
+			profile = v
+		} else {
+			profile = session.DefaultSharedConfigProfile
+		}
+		// create a cacheing Provider wrapper around the Credentials
+		if cacheProvider, err := NewFileCacheProvider(clusterID, profile, roleARN, sess.Config.Credentials); err == nil {
```
llamahunter: If caching is enabled, wrap the existing Credentials (and its contained Provider) in a new caching provider.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+	credentials                 *credentials.Credentials // the underlying implementation that has the *real* Provider
+	clusterID, profile, roleARN string                   // parameters used to create Provider
+	cachedToken                 cacheToken               // the cached token, if it exists
+}
+
+// NewFileCacheProvider creates a new Provider implementation that wraps a provided Credentials,
+// and works with an on disk cache to speed up token usage when the cached copy is not expired.
+// If there are any problems accessing or initializing the cache, an error will be returned, and
+// callers should just use the existing credentials provider.
+func NewFileCacheProvider(clusterID, profile, roleARN string, creds *credentials.Credentials) (FileCacheProvider, error) {
```
llamahunter: does a lot of check to make sure things look good with the cache.  If any don't look right, it bails on the cache and reverts to existing behavior.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+	}, nil
+}
+
+// Retrieve() implements the Provider interface, returning the cached token if is not expired,
+// otherwise fetching the token from the underlying Provider and caching the results on disk
+// with an expiration time.
+func (f *FileCacheProvider) Retrieve() (credentials.Value, error) {
+	if !f.cachedToken.IsExpired() {
+		// use the cached token
+		return f.cachedToken.Token, nil
```
llamahunter: The main point... if we can use the cached credentials, don't fetch them again, since that can send you through a lengthy and annoying SSO process.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+	} else {
+		_, _ = fmt.Fprintf(os.Stderr, "No cached token available.  Refreshing...\n")
+		// fetch the token from the underlying Provider
+		token, err := f.credentials.Get()
```
llamahunter: Expired credential, so fetch from underlying provider.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+					if _, ok := cache.ClusterMap[f.clusterID]; !ok {
+						// first use of this cluster id
+						cache.ClusterMap[f.clusterID] = map[string]map[string]cacheToken{}
+					}
+					if _, ok := cache.ClusterMap[f.clusterID][f.profile]; !ok {
+						// first use of this profile
+						cache.ClusterMap[f.clusterID][f.profile] = map[string]cacheToken{}
+					}
+					cache.ClusterMap[f.clusterID][f.profile][f.roleARN] = f.cachedToken
+					err = writeCacheWhileLocked(filename, cache)
```
llamahunter: Update the cache with the new credential and expiration time.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+	return f.cachedToken.Expiration
+}
+
+// CacheFilename returns the name of the token cache file, which can either be
+// set by environment variable, or use the default of ~/.kube/cache/token.yaml
+func CacheFilename() string {
+	if filename, ok := e.LookupEnv("AWS_IAM_AUTHENTICATOR_CACHE_FILE"); ok {
+		return filename
+	} else {
+		return filepath.Join(UserHomeDir(), ".kube", "cache", "token.yaml")
```
llamahunter: default cache location, but overridable via AWS_IAM_AUTHENTICATOR_CACHE_FILE env variable.

llamahunter:
Hmm... I think I might be using confusing terminology.  This is really caching the AWS credentials, not the kubernetes access token.  Would you like me to fix up the identifiers and comments to be clearer about what's going on?  Also, probably should reference https://github.com/aws/aws-sdk-go/issues/1329 somewhere in the code, so that people understand why this needs to be done, as the aws-sdk-go people say it's not their responsibility to cache credentials, and any app credential caching should *not* share a cache with the awscli for compatibility reasons.

llamahunter:
@nckturner @mattmoyer any thoughts on this?  Any rework needed?  Can it be merged & released?

llamahunter:
Still waiting for feedback on this.  Does it need rework?  Can it be merged?

llamahunter:
Hey... it's been about a month now.... please comment.

CharlieC3:
Any activity on this review process? Would love to see this merged.

llamahunter:
I can only ask so many times....  @CharlieC3 do you have any feedback on it?

CharlieC3:
I'm no official reviewer of this repo, but...

I'm assuming you were able to build/test this on your end @llamahunter, but I was not due to dependency issues. Given, I only tried for maybe 15 minutes :) I'll give it another go this week to see if I can test it on my end, otherwise the source code looks good at a quick glance, and the PR comments you left are helpful.

llamahunter:
Hmm... I would hope the branch would be able to build.  Go is pretty good about packaging all the dependent foo together.  The Travis CI builds succeeded, so I think build problems are probably a local config issue on your side.

It seems to work pretty well for me.  I have a number of different IAM roles I can adopt to switch to different permissions levels within the cluster.  Here's a section of my .kube/config file

```
users:
- name: arn:aws:eks:us-east-1:########:cluster/ekstest-admin
  user:
    exec:
      args:
      - token
      - -i
      - ekstest
      command: aws-iam-authenticator
      env:
      - name: AWS_PROFILE
        value: default
- name: arn:aws:eks:us-east-1:########:cluster/ekstest-dev
  user:
    exec:
      args:
      - token
      - -i
      - ekstest
      - --cache
      command: aws-iam-authenticator
      env:
      - name: AWS_PROFILE
        value: dev
- name: arn:aws:eks:us-east-1:########:cluster/ekstest-ops
  user:
    exec:
      args:
      - token
      - -i
      - ekstest
      - --cache
      command: aws-iam-authenticator
      env:
      - name: AWS_PROFILE
        value: ops
```
and my corresponding .aws/config

```
[default]
region = us-east-1

[dev]
region = us-east-1
credential_process = okta-credential_process arn:aws:iam::########:role/aws-dev

[ops]
region = us-east-1
credential_process = okta-credential_process arn:aws:iam::########:role/aws-ops
```

My default profile has a corresponding entry in .aws/credentials

```
[default]
aws_access_key_id = XXXXXXXX
aws_secret_access_key = XXXXXXXX
```
and so I don't have the --cached argument for that user in my .kube/config.  The other two users use okta-credential-process to get SSO credentials

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+			}
+		}
+		return token, err
+	}
+}
+
+// IsExpired() implements the Provider interface, deferring to the cached token first,
+// but fall back to the underlying Provider if it is expired.
+func (f *FileCacheProvider) IsExpired() bool {
+	return f.cachedToken.IsExpired() && f.credentials.IsExpired()
```
mogren: Any chance that `cachedToken` or `credentials` might be empty here?

cmd/aws-iam-authenticator/token.go
```diff
@@ -77,8 +78,10 @@ func init() {
 	tokenCmd.Flags().Bool("forward-session-name",
 		false,
 		"Enable mapping a federated sessions caller-specified-role-name attribute onto newly assumed sessions. NOTE: Only applicable when a new role is requested via --role")
+	tokenCmd.Flags().Bool("cache", false, "Cache the token on disk until it expires")
```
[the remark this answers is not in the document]
nckturner: Change to "Cache the credentials on disk until they expire".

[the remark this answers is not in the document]
nckturner: Also: "Uses the aws profile specified by AWS_PROFILE or the default profile."

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+var newFlock = func(filename string) filelock {
+	return flock.New(filename)
+}
+
+// cacheFile is a map of clusterID/roleARNs to cached tokens
+type cacheFile struct {
+	// a map of clusterIDs/profiles/roleARNs to cachedTokens
+	ClusterMap map[string]map[string]map[string]cacheToken `yaml:"clusters"`
```
nckturner: Let's see.. profile is included in the key because the same role could have multiple entries in your ~/.aws/profile with e.g. different MFA settings?  I think it makes sense just wondering if there is a cleaner representation.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+// with an expiration time.
+func (f *FileCacheProvider) Retrieve() (credentials.Value, error) {
+	if !f.cachedToken.IsExpired() {
+		// use the cached token
+		return f.cachedToken.Token, nil
+	} else {
+		_, _ = fmt.Fprintf(os.Stderr, "No cached token available.  Refreshing...\n")
+		// fetch the token from the underlying Provider
+		token, err := f.credentials.Get()
+		if err == nil {
```
nckturner: Can we reverse these `if` statements to remove this nesting?  Capture error and return at the top of the block.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+
+// ExpiresAt implements the Expirer interface, and gives access to the expiration time of the token
+func (f* FileCacheProvider) ExpiresAt() time.Time {
+	return f.cachedToken.Expiration
+}
+
+// CacheFilename returns the name of the token cache file, which can either be
+// set by environment variable, or use the default of ~/.kube/cache/token.yaml
+func CacheFilename() string {
+	if filename, ok := e.LookupEnv("AWS_IAM_AUTHENTICATOR_CACHE_FILE"); ok {
```
nckturner: Define this environment variable at the top of the package.  Also, `~/.kube/cache/token.yaml` should have something in the path to denote if belongs to `aws-iam-authenticator`.

pkg/token/filecache.go
```diff
@@ -0,0 +1,277 @@
...
+					f.cachedToken = cacheToken{
+						token,
+						expiration,
+						nil,
+					}
+					// don't really care about read error.  Either read the cache, or we create a new cache.
+					cache, _ := readCacheWhileLocked(filename)
+					if _, ok := cache.ClusterMap[f.clusterID]; !ok {
+						// first use of this cluster id
+						cache.ClusterMap[f.clusterID] = map[string]map[string]cacheToken{}
```
nckturner: Why don't we create utility functions with this cache implementation to hide some of this logic?

```
type cacheKey struct {
    clusterId string
    profile string
    roleARN string
}


func (c *cache) Add(key cacheKey, creds credentials) {
    ...
}    

func (c *cache) Get(key cacheKey) creds *credentials {
    ...
}
```

llamahunter:
Thx for the feedback. Will review above recommendations and push new code to the PR in about a week. Currently out of the country traveling w/o computer.

pkg/token/filecache.go
llamahunter: 'cachedToken' is an object.  'credentials' is a pointer, but is passed in to NewFileCacheProvider().  Conceivably, I suppose the user could pass in nil for that.  I will add a check for that in NewFileCacheProvider() and return an error if nil.

pkg/token/filecache.go
llamahunter: I definitely have different profiles for the same cluster, with different IAM roles associated with them.  The aws-iam-authenticator *also* allows profiles to optionally assume different roles via the -r option, if they're able.  I don't use that, in my config, tho.

llamahunter:
Addressed the review comments and rebased against master.

nckturner:
Appreciate the changes, taking a look.

nckturner:
Tested this PR after the changes and it works great.  Let's get it merged! Specifically tested:
* MFA caching
* Switching between two roles with MFA
* Expired credentials

/lgtm

robwittman:
🙌 Any timeline as to when we can see this in a release?

nckturner:
@robwittman working on it!  Hopefully this week.

joegoggins:
Looks cool. Thanks for tackling this. Look forward to trying it in a release. I'd like to consider swapping out our existing MFA caching solution to replace it with this. I tried to derive answers to some technical questions about this by reviewing the code changes but wasn't able to. If these get addressed in the release notes, please ignore and just ping me with a link to that when it ships.

_some assumed questions / answers_

## How do you enable caching? 

I assume `aws-iam-authenticator --cached`

## Where does it store it's cached credentials on disk?

I assume somewhere in `~/.kube/config`

## What does it cache?

I assume something like this is cached on disk and that it lasts for 15 minutes before re-prompting for MFA. Yea?
```
{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{},"status":{"token":"k8s-aws-v1.REDACTED...."}}`
```

llamahunter:
See `aws-iam-authenticator token --help`. Note that this caching is for the SSO credentials used to retrieve the token, not the token itself.  Someone else is working on token caching, I believe, but SSO credential fetching takes a *long* time (7-8 seconds), so caching the credentials is very useful.

Credentials are stored in `~/.kube/cache/aws-iam-authenticator/credentials.yaml` unless you override the location with the `AWS_IAM_AUTHENTICATOR_CACHE_FILE` environment variable

It caches the `access_key_id`, `secret_access_key`, and `session_token` for as long as you have configured your credentials to be cacheable.  I have mine set in IAM for 12 hours (the max).

truncs:
@llamahunter Thanks for all the hard work on this! Much appreciated! 
On the session duration, even after setting the max duration time in the IAM, the session durations that we get back are for 15 mins. Thoughts on why this might be the case? Happy to open a separate issue for this.

llamahunter:
What is your SSO system?  If using Okta, you must set the time in the ~/.okta/config.properties to set the OKTA_STS_DURATION.  YMMV if using a different SSO system.

CharlieC3:
@truncs I don't believe the duration is configurable due to a lack of configuration options in the AWS API.
https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html
https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/63
https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/75

llamahunter:
Longer durations up to 12 hrs for federated access using roles for about a year now. https://aws.amazon.com/about-aws/whats-new/2018/03/longer-role-sessions/

CharlieC3:
@llamahunter According to the links in my previous comment, and comments from the other PRs, it seems like `aws-iam-authenticator` may be using the `sig-v4-authenticating-requests` method. And if it does, that would still be subject to the duration set by the `sig-v4-authenticating-requests` method, which is hard-set to 15 minutes.

So it's possible both duration timeouts are getting used here, but the shorter one is not configurable and is consequently the one we're being affected by.

Keep in mind I'm not familiar with this project's source code, but the description of the 15 minute timeout duration for the `aws-iam-authenticator` fits suspiciously well into the 15 minute auth timeout described in this page.
https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html

llamahunter:
I think you're talking about token timeouts.  This is about caching credentials, not tokens.  I think there's another issue addressing the caching of tokens, for which there is a built in limit by AWS of 10-15 minutes.

llamahunter:
See the list of 'possibly addresses in alternate way' issues listed at the start of this pull request.

CharlieC3:
@llamahunter You might be right, I was under the impression they were affected by the same issue described in this comment.
https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/63#issuecomment-438050396

llamahunter:
Yes, per their comment, they appear to be using Okta SSO 2FA stuff, like we do.  If they set the IAM federated role connected to their Okta sign on to expire in 12 hours, and enable caching in aws-iam-authenticator, I suspect their lives will be a lot more pleasant.  I believe aws-iam-authenticator still fetches the token every time, tho, which, in my experience, is a short time delay (< 1 second). It would be nice if the token were cached for the full 15 minutes to skip kubectl talking to anything other than the EKS k8s control plane endpoint on every invocation.

truncs:
@llamahunter we use 2FA functionality provided by AWS.

PR #195, opened 2019-01-17
Remove vendored BUILD files

thumphries:
The presence of these vendored `BUILD` files makes it impossible to build this project with Bazel; after removal it builds out of the box with gazelle.

I believe the upstream libraries no longer include these files, so maybe there is a better way to do this.


thumphries:
CLA is signed, bot didn't refresh

nckturner:
/assign

Compatibility with bazel would be a welcome addition, but I'd like review by a gazelle contributor or experienced user and I assume WORKSPACE and BUILD files for the authenticator itself should be checked in as well?

PR #216, opened 2019-04-11
Update aws-go-sdk to v1.19.11

### LINKED ISSUE #207: iam authenticator with latest aws-sdk-go since proxies does not work
Hi,

This is Badal from AWS Premium Support. One of the customers is having issues with using aws-iam-authenticator.

"""
aws-iam-authenticator currently uses a version of aws-sdk-go that doesn't work with proxies when using custom CA bundles via the AWS_CA_BUNDLE environment variable. This prevents us using IAM authentication with kubectl because our corporate man-in-the-middle proxies require us to set a custom CA bundle.

We believe that using a more recent version of aws-sdk-go will fix this (assuming it includes https://github.com/aws/aws-sdk-go/pull/2343 ), but until then, the official release is broken.
"""

Can you confirm if this is the issue? Is there any way to work-around this issue?

Awaiting your response.

nckturner:
Update aws-go-sdk to newest release for authenticator v0.4.0 release.

Fixes #207


christopherhein:
@nckturner Given the size albeit not user facing should we add something to the `CHANGELOG`?

nckturner:
@christopherhein Agreed, let me submit a PR for that.

PR #217, opened 2019-04-12
Document that base64 encoding padding must be removed from bearer token

### LINKED ISSUE #202: EKS Authenticaton From Outside Cluster
I have a use case where I need to interact with an EKS cluster from a lambda function and encountered an issue that took extended debugging before I figured out what was going on. The issue is around creating the bearer token that the `aws-iam-authenticator`. 

Here's some simple Python code to create the token.

```python
REGION = 'eu-west-1'

def get_bearer_token(cluster, region):
    # ...

    signer = RequestSigner(...)

    params = {
        'method': 'GET',
        'url': 'https://sts.{}.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15'.format(REGION),
        'body': {},
        'headers': {
            'x-k8s-aws-id': cluster
        },
        'context': {}
    }

    signed_url = signer.generate_presigned_url(
        params,
        region_name=REGION,
        expires_in=60,
        operation_name=''
    )

    base64_url = base64.urlsafe_b64encode(signed_url.encode('utf-8')).decode('utf-8')

    return 'k8s-aws-v1.' + base64_url 
```

The issue I encountered was that if the base64 encoded string has any padding (ie., either `=` or `==`) then if you pass the token created with the above function it fails. However if I remove the padding it works:

```python
```

I'm not sure if this is a bug but it is very surprising and took me days to figure so at the very least it should be documented. I'm more than happy to do a pull request to add that documentation and in fact, I found there was very little official documentation on how this Bearer token was created at all (I had to piece it together from several blog posts) so I'd be happy to add that to the documentation as well.

svozza:
Closes #202. I've documented the issue with base64 encoding padding and I've included a code sample to show users how to construct the token as it's very difficult to find this information. The reason why I've chose to use Python and not Go is because anyone writing in Go would be using the IAM Authenticator Go client directly and it has this functionality built in already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,68 @@ This can be helpful for quickly attempting to associate "who performed action X 
 Please note, **this should not be considered definitive** and needs to be cross referenced via the `role id` (which remains consistent) with CloudTrail logs
 as a user could potentially change this on the client side.
 
+## API Authorization from Outside a Cluster
+
+It is possible to make requests to the Kubernetes API from a client that is outside the cluster, be that using the 
+bare Kubernetes REST API or from one of the language specific Kubernetes clients 
+(e.g., [Python](https://github.com/kubernetes-client/python)). In order to do so, you must create a bearer token that
+is included with the request to the API. This bearer token requires you append the string `k8s-aws-v1.` with a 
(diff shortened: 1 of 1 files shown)


christopherhein:
Thank you for this @svozza.

PR #226, opened 2019-05-02
Refactored EC2 API calls to be testable

micahhausler:
This change will make for easier testing functions that call the EC2 API


christopherhein:
Given the PR title should there be tests too? Otherwise looks good to me.

Issue #209, opened 2019-04-05
Add support for `external_id` when assuming a role


benderillo:
At the moment there is no way to specify `external_id` when specifying a role.
This makes it impossible to use `aws-ima-authenticator` to assume role that has `sts:ExternalId` condition set.

Here is our use case:
Terraform is used to bring up EKS cluster. Terraform assumes a role that has `sts:ExternalId` condition set. Terraform allows to specify `external_id` so there is no issue here.

However, the last step in setting up an EKS cluster and adding worker nodes to it implies that we need to `kubectl apply` a ConfigMap with Auth settings to allow worker nodes to join the cluster.

Before we can do `kubectl apply` we need to configure `kubectl` to access the cluster.
To do so, we need to use `aws-iam-authenticator token` with `-r role-arn` argument. 

The problem is, it will not work because there is no way to tell `aws-iam-authenticator token` to use `external_id` to get an access.

So we stuck not being able to complete the cluster setup.
I have submitted another request for aws-cli to add support for `external_id` parameter when running `aws eks update-kubeconfig`: https://github.com/aws/aws-cli/issues/4053

bensussman:
This is a major blocker for us, as we need the `external_id` to protect us from the confused deputy problem, but we are open to this attack for all of our external EKS clusters because there is no `external_id` support.

Please help @nckturner!

nckturner:
After a quick glance at the code it looks like it would need to be passed down and set on `sessionSetter` here: https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go#L243

The aws-cli issue that you linked states that a change would need to be made to the server side as well, but I'm not sure I understand that yet--my understanding is that the externalID would be used by the client when the role is assumed, and once the credentials come back, then the externalID has been verified in the trust policy of the role and we can forget about it.  Am I missing something here?

benderillo:
@nckturner I have the same understanding and also not quite sure why the `server` was mentioned.
Maybe @justnance can elaborate on that comment?

bensussman:
@nckturner The issue that is clear to me is that `aws-iam-authenticator` cannot accept `external_id` when the `-r <role_arn>` argument is provided. I am able to reproduce by running:

```
aws-iam-authenticator token -i <my-eks-cluster> -r <role-that-has-cluster-perms>
```

[...]

PR #228, opened 2019-05-19
add support for passing externalID to assume role

jeffmhastings:
It is common practice to require an "external ID" when making cross-account calls to STS assume role. This adds an optional `"--external-id"`/`"-e"`argument to the token command. When included it is passed in the AssumeRoleProvider.

Issue: https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/209

### DIFF
--- a/cmd/aws-iam-authenticator/token.go
+++ b/cmd/aws-iam-authenticator/token.go
@@ -52,13 +52,12 @@ var tokenCmd = &cobra.Command{
 			fmt.Fprintf(os.Stderr, "could not get token: %v\n", err)
 			os.Exit(1)
 		}
-		if roleARN != "" {
-			// if a role was provided, assume that role for the token
-			tok, err = gen.GetWithRole(clusterID, roleARN, externalID)
-		} else {
-			// otherwise sign the token with immediately available credentials
-			tok, err = gen.Get(clusterID)
-		}
+
+		tok, err = gen.GetWithOptions(&token.GetTokenOptions{
+			ClusterID:            clusterID,
+			AssumeRoleARN:        roleARN,
+			AssumeRoleExternalID: externalID,
+		})
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "could not get token: %v\n", err)
 			os.Exit(1)
--- a/cmd/aws-iam-authenticator/token.go
+++ b/cmd/aws-iam-authenticator/token.go
@@ -32,6 +32,7 @@ var tokenCmd = &cobra.Command{
 	Long:  ``,
 	Run: func(cmd *cobra.Command, args []string) {
 		roleARN := viper.GetString("role")
+		externalID := viper.GetString("externalID")
 		clusterID := viper.GetString("clusterID")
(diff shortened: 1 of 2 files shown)


christopherhein:
@jeffmhastings sorry for the delayed review on this. My one concern with how it was implemented is it will break clients like https://github.com/weaveworks/eksctl wonder if we can do this so it wouldn’t potentially adding a new func entirely and having the original funcs calling it? @errordeveloper and @nckturner definitely good to have your eyes on this

jeffmhastings:
Yeah that makes sense. Adding a function or two could work if you don't anticipate needing to add more in the future. Though at that point the API might start to get a bit confusing.

We could consider using an options struct or options funcs (like stscreds.NewCredentials).

`GetWithOptions(options GetTokenOptions)` seems like it might be the simplest and most flexible.

Anyway I'm open to whatever you all think is best.

jeffmhastings:
I took a stab at refactoring to use an options struct. This shouldn't break the existing API.

jeffmhastings:
It doesn't seem like it actually re-ran the tests with the /retest command

nckturner:
This looks great to me, @micahhausler can you take a quick look and we'll get this merged?

AmitNahMesh:
Hey, I Have the same issue as well. 
Could you let me know if you found a solution for this? @jeffmhastings
Here or at amit@mesh.security
Thanks!

jeffmhastings:
@AmitNahMesh I'm not sure I understand what issue you're having. Since this was merged you can now pass an external ID to the `aws-iam-authenticator token` command

```
aws-iam-authenticator token --help                                                                                             CORE-703 ✭ ✱
Authenticate using AWS IAM and get token for Kubernetes

Usage:
  aws-iam-authenticator token [flags]

Flags:
      --cache                  Cache the credential on disk until it expires. Uses the aws profile specified by AWS_PROFILE or the default profile.
  -e, --external-id string     External ID to pass when assuming the IAM Role
      --forward-session-name   Enable mapping a federated sessions caller-specified-role-name attribute onto newly assumed sessions. NOTE: Only applicable when a new role is requested via --role
  -h, --help                   help for token
      --region string          AWS region to use for assume role calls
  -r, --role string            Assume an IAM Role ARN before signing this token
  -s, --session-name string    Session name to pass when assuming the IAM Role
      --token-only             Return only the token for use with Bearer token based tools

Global Flags:
  -i, --cluster-id ID                 Specify the cluster ID, a unique-per-cluster identifier for your aws-iam-authenticator installation.
  -c, --config filename               Load configuration from filename
      --feature-gates mapStringBool   A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                      AllAlpha=true|false (ALPHA - default=false)
                                      AllBeta=true|false (BETA - default=false)
                                      ConfiguredInitDirectories=true|false (ALPHA - default=false)
                                      IAMIdentityMappingCRD=true|false (ALPHA - default=false)
  -l, --log-format string             Specify log format to use when logging to stderr [text or json] (default "text")
```

PR #255, opened 2019-08-15
Update go version

christopherhein:
The projects path has been updated and it's been vendored. Also allow you to use `-o json`, `-o yaml`, or `-s`.


nckturner:
Looks like you need to rebase @christopherhein

christopherhein:
@nckturner updated.

Issue #80, opened 2018-04-28
error: username or group is not a DNS-1123 hostname


brosner:
I am seeing this error by largely following the documentation. Groups in example.yaml and from the Kubernetes docs typically have a colon in them. For example `system:masters`. However, this is failing authentication because they don't conform. I am not sure whether I should be changing the group name or if this constraint is either in the wrong place or should be removed altogether?

christopherhein:
@nckturner @mattmoyer I was digging into the username fields in https://github.com/kubernetes/kubernetes and it doesn't look like they have any validation on the field. Do you know of anything different?

jutley:
I just ran into this issue. I do not see anything in Kubernetes that says that usernames have the restrictions that this project imposes. 

The closest thing I could find is the restrictions on [resource names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names), and [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

Furthermore, our current setup uses users with "Firstname Lastname" as their username, which doesn't match DNS-1123. In the [official documentation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authentication-strategies), the sample username is an email address (which we would like to move to), which also does not match DNS-1123.

I think this bit of validation needs to be removed. From what I can tell, I think we just need to edit [this code](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/server/server.go#L340-L346) as follows:
```
template = strings.Replace(template, "{{SessionName}}", identity.SessionName, -1)
```

richardmarshall:
I ran into this as well. Given that k8s doesn't place a dns-1123 restriction on usernames or groups I definitely think this check as well as the replacement of `@`s should be removed.

Additionally the current regex used to check if a user/group name is conforming contains an error. 

```"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"```

The expression isn't end/beginning of line anchored so it will match any string that contains a sub-string that is a valid dns-1123 name.

/reopen
/remove-lifecycle rotten

PR #260, opened 2019-08-30
Remove DNS-1123 validation of usernames and groups

richardmarshall:
k8s doesn't require user or group names to be DNS-1123 labels so there isn't a need to enforce or transform these values to conform to that format.

To that end this PR removes the regexp check which validates DNS-1123 and removes the string replacement of `@`s in SessionName.

Also of note that the current regex is missing start and end of line anchors so the validation is only ensuring that a group or user contains a sub-string that is a valid DNS-1123 label.

The second change does have impacts on existing usage of the `{{SessionName}}` template value which could break rolebindings for users who are making use of this value and have email addresses for session names.

I see a few ways this could be addressed and am happy to update the PR to reflect a desired path for this.

* Liberal documentation updates to warn users of the change.
* Toggle the change on a server flag (something along the lines of `--legacy-session-name=true`)
* Creating a new template variable (`{{SessionNameRaw}}`) that doesn't do the replacement and leave the existing behavior for `{{SessionName}}` in place.

Related to #80

### DIFF
--- a/deploy/example.yaml
+++ b/deploy/example.yaml
@@ -103,6 +103,15 @@ data:
         username: admin:{{SessionName}}
         groups:
         - system:masters
+      # map federated users in my "KubernetesUsers" role to users like
+      # "alice@example.com". SessionNameRaw is sourced from the same place as
+      # SessionName with the distinction that no transformation is performed
+      # on the value. For example an email addresses passed by an identity
+      # provider will not have the `@` replaced with a `-`.
+      - roleARN: arn:aws:iam::000000000000:role/KubernetesUsers
+        username: "{{SessionNameRaw}}"
+        groups:
+        - developers
       # each mapUsers entry maps an IAM role to a static username and set of groups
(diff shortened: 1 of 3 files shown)


micahhausler:
I don't have an issue with removing the regex, but the `{{SessionName}}` a breaking change is a no-go.

> * Creating a new template variable ({{SessionNameRaw}}) that doesn't do the replacement and leave the existing behavior for {{SessionName}} in place.

I like this idea a lot and would happily accept this

richardmarshall:
> breaking change is a no-go.

I expected that but wanted to seed discussion with the lowest effort code change 😄 

> I like this idea a lot and would happily accept this

Great, will update the code and add some documentation for the new template variable.

richardmarshall:
@micahhausler Updated, PTAL

richardmarshall:
@micahhausler @nckturner Anything else needed on my end to help move this forward?

PR #272, opened 2019-10-25
Adding session name parameter to TokenGenerator

SaranBalaji90:
Adding sessionName parameter to authenticator client. Clients can set different session name when using roles to authenticate to API Server.

Changes include, 
cmd/../token.go -> 
i) Adding new parameter to read the session name. 
ii) Added validation to say either forwardSessionName or sessionName can be used when using role but not both.

pkg/.../token.go -> 
i) Added SessionName to GetTokenOptions struct
ii) Setting session name when creating session object through role.

cr https://code.amazon.com/reviews/CR-14787523

### DIFF
--- a/cmd/aws-iam-authenticator/token.go
+++ b/cmd/aws-iam-authenticator/token.go
@@ -36,10 +36,17 @@ var tokenCmd = &cobra.Command{
 		clusterID := viper.GetString("clusterID")
 		tokenOnly := viper.GetBool("tokenOnly")
 		forwardSessionName := viper.GetBool("forwardSessionName")
+		sessionName := viper.GetString("sessionName")
 		cache := viper.GetBool("cache")
 
 		if clusterID == "" {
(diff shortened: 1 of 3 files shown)


cmd/aws-iam-authenticator/token.go
```diff
@@ -36,13 +36,19 @@ var tokenCmd = &cobra.Command{
...
 		forwardSessionName := viper.GetBool("forwardSessionName")
+		sessionName := viper.GetString("sessionName")
 		cache := viper.GetBool("cache")
 
 		if clusterID == "" {
 			fmt.Fprintf(os.Stderr, "Error: cluster ID not specified\n")
 			cmd.Usage()
 			os.Exit(1)
 		}
+		if forwardSessionName && sessionName != "" {
```
ofiliz: Please leave one blank line between the if blocks. Also please fix the indentation for cmd.Usage and os.Exit calls.

PR #276, opened 2019-11-07
Update aws-sdk-go version to 1.25.29

h3poteto:
I updated `aws-sdk-go` version to 1.25.29. I want to use this library in EKS with IAM Role for Service Account, so it is required that the version is greater than 1.23.13.


micahhausler:
Hi @h3poteto thanks for this PR. I'm going to close in favor of #283 which includes IRSA support and IMDS v2.

PR #278, opened 2019-11-14
Allow server port customization

diversario:
I had to make this change to stop authenticator from binding to port `0` (unsure why). It preserves current behavior while making the port user-customizable.


micahhausler:
Can you rebase? otherwise LGTM

PR #279, opened 2019-11-22
Require to pass in interface instead of the concrete type

Skarlso:
Requiring the concrete type there ties this library to the whims of the sts structure and prevents any user from testing this part of their code as we can't pass in an implementation of STS.

As the Go mantra goes: Require Interfaces return Concretes.

This "fix" should be seamless. Anything should just work. 🤞


Skarlso:
Doesn't appear to be related.

PR #283, opened 2019-12-20
switch to use regional sts endpoint & imdsV2

M00nF1sh:
1. use regional STS endpoint when "--region" is specified.
   * when "--region ${AWS_REGION}" flag is explicitly specified, regional STS endpoint will be used.
   * when other AWS_SDK aware region config is specified(e.g. `AWS_REGION` or `AWS_DEFAULT_REGION` environment variable, `region` config in ~/.aws/config), the default logic of using global STS endpoint still applies. This is to maintain backwards-compatible for KOPS users that might have disabled regional STS endpoint. However, this is likely to be changed to use regional STS endpoints as well after a few versions.
2. upgrade AWS SDK to newest version(v1.26.7) which uses IMDSV2 by default.

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -29,10 +29,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws/credentials"
-
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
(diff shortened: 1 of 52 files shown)


pkg/token/token.go
```diff
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
```
wongma7: nit, import is out of order w/ other aws-sdk-go imports

M00nF1sh: nice catch :D

Issue #263, opened 2019-09-10
When using assumed roles, you can't attribute audit event actions to a specific aws user.


tjsingleton:
TL;DR: I don't think it's possible to know for certain who the actor is in the kubernetes audit logs when the caller assumed a role.

The goal is that I can trace an k8s audit record to a single aws user who made the api call. We manage our users in a central account and they assume a shared role into the accounts where the actual aws resources live. For events in cloudtrail aws logs the accessKeyId of the role assumption session. Therefore I can correlate that  accessKeyId with the original "log in". (Described here: https://aws.amazon.com/blogs/security/how-to-audit-cross-account-roles-using-aws-cloudtrail-and-amazon-cloudwatch-events/) 

However, the k8s audit logs do not include the accessKeyId. 

Example:

```
...
    "user": {
        "username": "platform:john.doe-617090408934",
        "uid": "heptio-authenticator-aws:617090408934:PBNMABOXMHFFMNIGGNZGX",
        "groups": [
            "system:masters",
            "system:authenticated"
        ]
    },
...
```

The uid includes the id of the assumed role, but that isn't unique to the session and doesn't identify who called it. The username field includes the role and the session name, however the session name is generated client side and could be forged. We've included the user's name by convention, but it can't be trusted. (The docs for the iam-authenticator call out that this can't be trusted https://github.com/kubernetes-sigs/aws-iam-authenticator#note-for-federated-users)

If the accessKeyId was available as part of how iam-authenticator generates the username or uid for the audit event, you could correlate it back to the cloudtrail entry.

jritchey:
@tjsingleton in your example is `PBNMABOXMHFFMNIGGNZGX` the accessKeyId you are looking for? If you search CloudTrail and filter on `Resource Name` and `PBNMABOXMHFFMNIGGNZGX` does it lead back to the IAM user you expected?

tjsingleton:
Yeah, it's the assumed role id.

PR #286, opened 2020-01-23
Add AWS AccessKeyID as an extra field in UserInfo

pepov:
Add AWS AccessKeyID as an extra field in UserInfo to be able to correlate the actual user with Kubernetes audit log events.

This PR is based and following up on:
- https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/263
- https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/282

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -330,6 +330,9 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
 				Username: username,
 				UID:      uid,
 				Groups:   groups,
+				Extra: map[string]authenticationv1beta1.ExtraValue{
+					"awsaccesskeyid": {identity.AccessKeyID},
+				},
 			},
 		},
 	})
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -330,6 +330,9 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
 				Username: username,
 				UID:      uid,
 				Groups:   groups,
+				Extra: map[string]authenticationv1beta1.ExtraValue{
+					"awsaccesskeyid": {identity.AccessKeyID},
```
micahhausler: nit: can you change this to `accessKeyId`?

pepov: of course

pepov:
@micahhausler This is great, thank you! Is there a way to know which release is available in which EKS version? In other words will I know whether this change has been landed other than looking at the actual audit logs?

micahhausler:
@pepov We haven't in the past announced what commit/version/features are in the authenticator server. You will see it show up in the audit logs eventually, but you could also open an issue on github.com/aws/containers-roadmap. Title it something like "IAM Role Traceability" that will open it up for more +1's, and help our PM team track and drive the issue.

PR #287, opened 2020-01-27
Restrict ClusterRole to readonly IAMIdentityMapping access

rifelpet:
In the spirit of least privilege security, we shouldn't give write access to the custom resources since it is not needed.

### DIFF
--- a/deploy/example.yaml
+++ b/deploy/example.yaml
@@ -27,9 +27,18 @@ rules:
 - apiGroups:
   - iamauthenticator.k8s.aws
   resources:
-  - "*"
+  - iamidentitymappings
   verbs:
-  - "*"
+  - get
+  - list
+  - watch
+- apiGroups:
+  - iamauthenticator.k8s.aws
+  resources:
+  - iamidentitymappings/status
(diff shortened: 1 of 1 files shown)


micahhausler:
I don't think this is quite right, the server [updates the status](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/mapper/crd/controller/controller.go#L214) on CRD objects.

rdrgmnzs:
@micahhausler do you think setting patch & update on `IAMIdentityMapping/status` would be enough? Give of course that the user has pre-created the identity mappings as `IAMIdentityMapping`. I skimmed over the 2 PRs that looked related to the CRD change and got a sense that would be enough but would like validation from someone more familiar with the code.

I'm in a fairly security conscious environment and can see giving the application performing authentication the ability to create its own identity mapping becoming an issue very fast.

rifelpet:
I've updated the ClusterRole to include write access to the status subresource

rifelpet:
@micahhausler any update on this? just checking in. TY

PR #301, opened 2020-03-13
Run entrypoint aws-iam-authenticator as root

wongma7:
I'm not sure why we want to run as user `10000` but it doesn't work for me because go builds binaries with permissions like so:
```
$ ls -la dist/goreleaserdocker010310734/aws-iam-authenticator 
-rwx------. 7 fedora fedora 38055936 Mar 13 20:39 dist/goreleaserdocker010310734/aws-iam-authenticator*
```

and user `10000/aws-iam-authenticator` then gets permission denied when trying to run it:
```
$ make build
...
$ docker run 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-iam-authenticator:v0.5.0-debian-stretch
standard_init_linux.go:211: exec user process caused "permission denied"
```

After these changes,
```
$ make build
...
$ 
docker run 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-iam-authenticator:v0.5.0-debian-stretch
A tool to authenticate to Kubernetes using AWS IAM credentials
...
```


M00nF1sh:
Hi, runs containers as root have potential security issues. (the root in container is actually root on host), there are multiple bench tool will report contains as unsafe, so we shouldn't change this.
BTW, there are multiple changes unrelated to this as well, (like you removed `yum update -y` and certificates).
/hold

Dockerfile.debian-jessie
```diff
@@ -13,16 +13,5 @@
...
-RUN adduser \
-    --system \
-    --no-create-home \
-    --uid 10000 \
-    --disabled-password \
-    --disabled-login \
-    aws-iam-authenticator && \
-    apt-get update && \
-    apt-get install -y ca-certificates && \
-    rm -rf /var/lib/apt/lists/*
```
wongma7: whoops, should not remove certificates

wongma7:
Problem is that hardcoding magic user id like 10000 breaks kube policies like MustRunAsUser because then the pod doesn't have expected permissions. However that is beside the point because it seems our image is broken anyway, when I run `make build` the executable is owned by me, not user 10000, and so 10000 doesn't have permission to run it?!

I'm going to try adding a chown step to the Dockerfile.

wongma7:
closing in favor of https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/302

PR #302, opened 2020-03-13
chown aws-iam-authenticator to avoid permission denied

wongma7:
Fixing the issue in #301 in a less controversial way :)

go builds binaries with permissions like so:
```
$ ls -la dist/goreleaserdocker010310734/aws-iam-authenticator 
-rwx------. 7 fedora fedora 38055936 Mar 13 20:39 dist/goreleaserdocker010310734/aws-iam-authenticator*
```
and user 10000/aws-iam-authenticator then gets permission denied when trying to run it:
```
$ make build
...
$ docker run 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-iam-authenticator:v0.5.0-debian-stretch
standard_init_linux.go:211: exec user process caused "permission denied"
```
After these changes the containers work as expected,
```
$ make build
...
$ 
docker run 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-iam-authenticator:v0.5.0-debian-stretch
A tool to authenticate to Kubernetes using AWS IAM credentials
...
```

### DIFF
--- a/Dockerfile.alpine-3.6
+++ b/Dockerfile.alpine-3.6
@@ -16,5 +16,6 @@ FROM alpine:3.6
 RUN adduser -D -u 10000 aws-iam-authenticator
 RUN apk add --update ca-certificates
 COPY aws-iam-authenticator /
+RUN chown aws-iam-authenticator /aws-iam-authenticator
 USER aws-iam-authenticator
 ENTRYPOINT ["/aws-iam-authenticator"]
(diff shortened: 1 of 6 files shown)


christopherhein:
Does it still make sense to have these Dockerfiles? Given that core has moved to `distroless` and subprojects like kubebuilder, cluster-api, etc has as well, it seems like that should be the path this the authenticator is going too.

wongma7:
> Does it still make sense to have these Dockerfiles? Given that core has moved to distroless and subprojects like kubebuilder, cluster-api, etc has as well, it seems like that should be the path this the authenticator is going too.

I had exactly this thought when making the same change to 6 almost-identical files... IMO no, I can clean up the Dockerfiles this in a follow-up but mainly this PR is to unblock my testing for #303

PR #310, opened 2020-04-20
Show log message when configmap can't be watched

woz5999:
I was running into issues setting this up in my cluster for the first time. It was extremely difficult to diagnose the cause without being able to see the actual error that was being thrown. Submitting this to improve the lives of future people like me.


woz5999:
It's not clear why the travis build is failing. It appears to be due to 
```The command "hack/check-vendor.sh" exited with 1.``` though no vendoring was touched in this PR. The source branch is also branched from the current latest commit in master.

PR #313, opened 2020-05-01
WIP e2e test runner

nckturner:


hack/authentication.yaml
```diff
@@ -0,0 +1,2 @@
+  authentication:
+    aws: {}
```
rifelpet, line 2: [these](https://pkg.go.dev/k8s.io/kops/pkg/apis/kops?tab=doc#AwsAuthenticationSpec) are the fields you can add here.

```yaml
  authentication:
    aws:
      image: foo
```

PR #314, opened 2020-05-04
Update Example

arhea:
In this pull request, I have updated the example configuration to update the cluster role and added an example of the `aws-auth` config map.

**Problem:**
The example configuration was missing the RBAC permissions to access the `aws-auth` ConfigMap and an example of the `aws-auth` ConfigMap which has a slightly different configuration than the `config.yaml`.

### DIFF
--- a/deploy/aws-auth-example.yaml
+++ b/deploy/aws-auth-example.yaml
@@ -1,0 +1,35 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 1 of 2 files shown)


christopherhein:
Weird that this test just started failing out of no where… seems it's not related to what you've done same this with this PR - https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/310#issuecomment-616752729

PR #315, opened 2020-05-05
Add experimental e2e test script

wongma7:
This little script should help with sanity checking the authenticator works end-to-end using the example.yaml.

It's experimental because it makes a lot of implicit assumptions (like you are currently using some IAM role) and it writes/reads/installs files all over the place.

Ideally it should be run in a container and we could set up a prow job, but I have not tested it in a container yet.

And in the future tests should be written in golang, not crappy bash. At least there should be tests for the other backend modes.

First time usage:
```
./test.sh
```

To avoid recompiling kubernetes:
```
env GO_OUT=$PWD/kubernetes/_output/bin ./test.sh
```

To skip tearing down the cluster (so you can debug)
```
env DOWN=false ./test.sh
```

### DIFF
--- a/hack/test.sh
+++ b/hack/test.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+
+DOWN=${DOWN:true}
+
+super_echo() {
+  echo ""
+  echo "**"
+  echo "**$1"
+  echo "**"
+}
+
+super_echo "WARNING: this script is experimental, run at your own risk!"
+
+super_echo "Install kubetest"
+if [ ! -d "test-infra" ] ; then
+  git clone https://github.com/kubernetes/test-infra.git
+fi
+pushd test-infra
+go install ./kubetest
+popd
+
+super_echo "Install hack/local-cluster-up.sh"
+if [ ! -d "kubernetes" ] ; then
+  git clone https://github.com/kubernetes/kubernetes.git
+fi
+pushd kubernetes
+git checkout release-1.17
+hack/install-etcd.sh
+
+super_echo "Init authenticator for the .kubeconfig"
+aws-iam-authenticator init -i my-dev-cluster.example.com
+cp cert.pem /var/aws-iam-authenticator/cert.pem
+cp key.pem /var/aws-iam-authenticator/key.pem
+cp aws-iam-authenticator.kubeconfig /etc/kubernetes/aws-iam-authenticator/kubeconfig.yaml
+
+down_and_exit() {
+  if [ "$DOWN" == "true" ]; then
+    super_echo "Down cluster"
+    kubetest --deployment local --down
+    sudo pkill -f kubelet
+  fi
+  exit $!
+}
+
+trap down_and_exit SIGINT
+
+super_echo "Up cluster"
+sudo pkill -f kubelet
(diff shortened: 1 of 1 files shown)


wongma7:
BTW, I forgot https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/313/files already existed, but anyway this can coexist with that. This uses local-up-cluster.sh and that one uses kops.

nckturner:
This might need:  `./hack/install-etcd.sh`, when I ran it, it failed because I didn't have etcd in my path.

wongma7:
Added  ./hack/install-etcd.sh .. tested by removing etcd from PATH and the script worked

nckturner:
Hmm, I got: 
```
**
**Test should be authorized when authenticator is running and role is in mapRoles
**
clusterrole.rbac.authorization.k8s.io/aws-iam-authenticator created
serviceaccount/aws-iam-authenticator created
clusterrolebinding.rbac.authorization.k8s.io/aws-iam-authenticator created
configmap/aws-iam-authenticator created
daemonset.apps/aws-iam-authenticator created
waiting for daemonset ready...
waiting for daemonset ready...
daemonset ready
error: You must be logged in to the server (Unauthorized)

**
**FAIL!
**

```

hack/test.sh
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
+
+DOWN=${DOWN:true}
```
nckturner, line 3: Should this be `${DOWN:-true}`?  I am unfamiliar with just a `:`.

hack/test.sh
```diff
@@ -0,0 +1,96 @@
...
+
+super_echo() {
+  echo ""
+  echo "**"
+  echo "**$1"
+  echo "**"
+}
+
+super_echo "WARNING: this script is experimental, run at your own risk!"
```
nckturner, line 12: I like how this prints but doesn't give you the opportunity to stop it.  Also, what is the risk?  Seems like a rather innocuous script.

hack/test.sh
```diff
@@ -0,0 +1,96 @@
...
+pushd test-infra
+go install ./kubetest
+popd
+
+super_echo "Install hack/local-cluster-up.sh"
+if [ ! -d "kubernetes" ] ; then
+  git clone https://github.com/kubernetes/kubernetes.git
+fi
+pushd kubernetes
+git checkout release-1.17
```
nckturner, line 27: Can we parameterize the version?

nckturner:
This could be causing the issue I'm getting: 

```
cp: cannot create regular file '/var/aws-iam-authenticator/cert.pem': Permission denied
cp: cannot create regular file '/var/aws-iam-authenticator/key.pem': Permission denied
cp: cannot create regular file '/etc/kubernetes/aws-iam-authenticator/kubeconfig.yaml': Permission denied

```

hack/test.sh
wongma7: yes

hack/test.sh
wongma7: IDK maybe I'll remove this warning. Or at least reword it to be clear that it's writing stuff into /var/kubernetes (since that's what aws-iam-authenticator init and hack/local-up-cluster.sh does)

hack/test.sh
wongma7: workingon it

Issue #300, opened 2020-03-07
Add latest alpine image


inductor:
Alpine linux is now 3.11 and your supporting versions are 3.6 and 3.7

PR #320, opened 2020-06-06
Add alpine 3.11 and debian buster

inductor:
Fixes https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/299 and https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/300


inductor:
@micahhausler @wongma7 @nckturner 

Can you have a look?
I don't think this is a complicated Pull Request...

PR #327, opened 2020-07-07
Added server side AWS account ID log redaction

micahhausler:
This feature allows administrators to prevent specific account IDs from being emitted in server logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,13 +288,13 @@ as a user could potentially change this on the client side.
 
 ## API Authorization from Outside a Cluster
 
-It is possible to make requests to the Kubernetes API from a client that is outside the cluster, be that using the 
-bare Kubernetes REST API or from one of the language specific Kubernetes clients 
+It is possible to make requests to the Kubernetes API from a client that is outside the cluster, be that using the
+bare Kubernetes REST API or from one of the language specific Kubernetes clients
 (e.g., [Python](https://github.com/kubernetes-client/python)). In order to do so, you must create a bearer token that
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -394,6 +394,11 @@ server:
   # role to assume before querying EC2 API in order to discover metadata like EC2 private DNS Name
   ec2DescribeInstancesRoleARN: arn:aws:iam::000000000000:role/DescribeInstancesRole
 
+  # AWS Account IDs to scrub from server logs. (Defaults to empty list)
+  scrubbedAccounts:
```
nckturner, line 398: Nit: To me "redacted" is more explicit than "scrubbed"

README.md
```diff
@@ -394,6 +394,11 @@ server:
   # role to assume before querying EC2 API in order to discover metadata like EC2 private DNS Name
   ec2DescribeInstancesRoleARN: arn:aws:iam::000000000000:role/DescribeInstancesRole
 
+  # AWS Account IDs to scrub from server logs. (Defaults to empty list)
```
nckturner, line 397: Do we want to give the ability to redact more than just account IDs?

README.md
micahhausler: I went back and forth on this, but in my mind "redacted" sounded too much like the whole log entry was omitted, and in this case I'm just redacting part of the log line.

README.md
micahhausler: I think we could certainly add that at some point

README.md
wongma7: I like and understand "scrub" FWIW. 👍

PR #334, opened 2020-08-28
Fix RBAC on example file: service account requires get to ConfigMap

pjbgf:
The existing example yaml does not define the RBAC for getting the ConfigMap that contains the configuration.

Please note that a least privileged approach was chosen, the service account will only be able to `get` the ConfigMap `aws-iam-authenticator`.


deploy/example.yaml
```diff
@@ -47,6 +47,14 @@ rules:
   - create
   - update
   - patch
+- apiGroups:
+  - ""
+  resources:
+  - configmap
+  resourceNames:
+  - aws-iam-authenticator
```
micahhausler, line 55: It looks like the ConfigMap mapper is hard coded to watch the [`aws-auth`](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/mapper/configmap/configmap.go#L59), can you correct that?

deploy/example.yaml
```diff
@@ -47,6 +47,14 @@ rules:
...
+  verbs:
+  - get
```
micahhausler, line 57:
```suggestion
  - get
  - watch
```

micahhausler (changes requested):
Have you tried this policy change and tested it? I'm not sure if `list` is also needed on ConfigMaps

deploy/example.yaml
pjbgf: Ok, will get the example to have that changed too.

micahhausler (approved):
/lgtm
/approve

Thanks!

PR #335, opened 2020-08-28
Security Improvements on the example yaml

pjbgf:
Sprinkling some security goodness on the example yaml file to help users do the right thing.
This change will:

- Drop ALL capabilities
- Enable seccomp
- Disable privilege escalation


wongma7:
/lgtm
/approve
i assume allowPrivilegeEscalation: false is the default, but doesnt hurt to set it expliclty

pjbgf:
> /lgtm
> /approve

thank you!

> i assume allowPrivilegeEscalation: false is the default, but doesnt hurt to set it expliclty

Actually, the default is actually `true`. 😅 

[Official docs](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation): _This defaults to allowed so as to not break setuid binaries..._

PR #348, opened 2020-10-16
(arn): validate partition against all partitions returned by the aws sdk

cmdallas:
Adds missing arn partitions to `checkPartition`.

### DIFF
--- a/pkg/arn/arn.go
+++ b/pkg/arn/arn.go
@@ -58,12 +58,10 @@ func Canonicalize(arn string) (string, error) {
 }
 
 func checkPartition(partition string) error {
-	switch partition {
-	case endpoints.AwsPartitionID:
-	case endpoints.AwsCnPartitionID:
-	case endpoints.AwsUsGovPartitionID:
-	default:
-		return fmt.Errorf("partion %s is not recognized", partition)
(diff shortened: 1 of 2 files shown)


pkg/arn/arn.go
```diff
@@ -62,6 +62,8 @@ func checkPartition(partition string) error {
 	case endpoints.AwsPartitionID:
 	case endpoints.AwsCnPartitionID:
 	case endpoints.AwsUsGovPartitionID:
+	case endpoints.AwsIsoPartitionID:
```
micahhausler, line 65: Can you modify this function so instead of enumerating partitions, we just check all the ones the SDK knows about? Check out [`endpoints.DefaultPartitions()`](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#DefaultPartitions)

micahhausler (changes requested):
Thanks for adding this, can you refactor that function?

pkg/arn/arn.go
cmdallas: +1, Feedback addressed.

```
=== RUN   TestUserARN
--- PASS: TestUserARN (0.00s)
PASS
```

micahhausler (approved):
/lgtm
/approve

PR #352, opened 2020-11-17
Support IPv6 listen address

### LINKED ISSUE #339: IPv4ism: Concatenates host:port without escaping
https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/cf4ec70edb2da559b6632bc6f683adec008b574a/pkg/server/server.go#L114

Naively concatenating an address + colon + port, only works when the address is either an IPv4 literal, or hostname.  It fails when the address contains embedded colons, as with IPv6 literals.  Using a function like [`net.JoinHostPort`](https://golang.org/pkg/net/#JoinHostPort) will escape correctly when required.

anguslees:
Clarify comments to separate `Config` into:
- `Address` - the IP address the server binds to, probably 0.0.0.0
- `Hostname` - the hostname/IP that clients should use

Also fix both so they handle IPv6 literals correctly.  Fixes #339


micahhausler (approved):
/lgtm
/approve

Thank you!

PR #360, opened 2021-01-30
Update aws sdk go v1.37.1

### LINKED ISSUE #296: Credentials not found when used with AWS SSO
I want to use `aws-iam-authenticator` to connect to a `k8s` cluster based on AWS IAM integrated with SSO. I connect to AWS based on these instructions:

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html

Having run the `aws configure sso` a browser tab opens and a number of redirects happen. Then I click a button to confirm AWS CLI access, I enter a few more preferences including a profile name and the command finishes. After this, I can successfully run:

    $ aws sts get-caller-identity --profile my-profile-name

to show I am authenticated. However, if I set this profile as my `$AWS_PROFILE` and try to run `aws-iam-authenticator`:

    $ aws-iam-authenticator token -i my-cluster-id
    could not get token: NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors

Digging into this in more detail, it looks like the SSO auth process has not produced the usual AWS credentials file in `$HOME/.aws/credentials`. Instead, there is a json file in `$HOME/.aws/cli/cache` that has the usual access ID, secret ID and session token. If I render these as environment variables or into a `credentials` file then my `aws-iam-authenticator` call works fine and I can wire it into my `.kube/config` file to run commands against my cluster. 

I also looked at the underlying [Go SDK methods](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/) and I can't, indeed, see any credentials providers for this type of cache file. Does this mean the problem is upstream in the Go SDK? Shall I raise an issue there?

Superficially, this issue looks similar to some others, but I think these are different:

- https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/141 - but I think this is about third-party credential providers, not those that are generated by native AWS SSO integration
- https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/193 - but I think this is about caching the AWS credential obtained from SSO in exactly the same way the AWS2.0 CLI is currently doing, rather than reading those back for adhoc `kubectl` usage.

Happy to be corrected on any of this - where am I going wrong?

gaffneyd4:
This updates `aws-sdk-go` to version 1.37.x, adding support for AWS Single Sign-On (SSO) Credential Provider (added in [v1.37.0](https://github.com/aws/aws-sdk-go/releases/tag/v1.37.0)) plus some additional fixes in [v1.37.1](https://github.com/aws/aws-sdk-go/releases/tag/v1.37.1).

Resolves #296


vikas027:
Hello maintainers, 
Any ETA of this getting merged? :)

nckturner:
Thanks!

nckturner (approved):

encron:
Sorry for commenting on a merged PR, but when can we expect this to be released? Kind of waiting for this one for a while to be able to use the AWS SSO provider. Thx!

PR #371, opened 2021-05-07
goreleaser: bump release to 0.164.0 and fix config deprecations

antoinedeschenes:
Required groundwork for Graviton and darwin arm64 support

Signed-off-by: Antoine Deschênes <antoine@antoinedeschenes.com>

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -2,7 +2,8 @@
 project_name: authenticator
 
 builds:
-  - binary: aws-iam-authenticator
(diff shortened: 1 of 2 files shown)


antoinedeschenes:
Still need to add golang 1.16, arm64 platform and multiarch docker images after this, would you rather have it in a single PR?

nckturner:
Sorry for the delay, either is fine but we can do a separate PR to keep it simple.

PR #387, opened 2021-07-22
Detect and panic when watch failure happen

PatrickXYS:
For iam-authenticator component, we need telemetry and metrics

Changes we've made in the PR:
1. Add detailed error message when issue happen
2. Add new prom metric "aws_iam_authenticator" for signaling 
3. Partial Improvements including formating

### DIFF
--- a/pkg/mapper/configmap/configmap.go
+++ b/pkg/mapper/configmap/configmap.go
@@ -18,7 +18,13 @@ import (
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
+)
 
+const (
+	metricSuccess       = "success"
+	metricFailure       = "fail"
+	metricSuccessUnit   = 1.0
+	metricFailureUnit   = 0.0
 )
 
 type MapStore struct {
@@ -47,7 +53,7 @@ func New(masterURL, kubeConfig string) (*MapStore, error) {
 
 // Starts a go routine which will watch the configmap and update the in memory data
 // when the values change.
-func (ms *MapStore) startLoadConfigMap(stopCh <-chan struct{}) {
+func (ms *MapStore) startLoadConfigMap(stopCh <-chan struct{}, metricsObj metrics) {
 	go func() {
 		for {
 			select {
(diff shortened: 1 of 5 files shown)


PatrickXYS:
/cc @jyotimahapatra

Can you help take a look? The PR is still WIP but want to get some feedback from your side

PatrickXYS:
/hold 

Hold until review finishes

pkg/mapper/configmap/configmap.go
```diff
@@ -21,6 +21,13 @@ import (
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
 )
 
+const (
+	metricSuccess     = "success"
```
nckturner, line 25: Seems weird to define these in the "configmap" package, since they look pretty generic.  They probably belong in a metrics package which we can create now.

pkg/mapper/configmap/mapper.go
```diff
@@ -11,6 +12,27 @@ type ConfigMapMapper struct {
...
+// metrics are handles to the collectors for prometheus for the various metrics we are tracking.
+type metrics struct {
+	watch *prometheus.GaugeVec
+}
+
+func createMetrics() metrics {
+	m := metrics{
+		watch: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: metricNS,
+			Name:      "authenticator_watch_success",
```
jyotimahapatra, line 28: would it make more sense to call it `authenticator_watch_active` (since there is no related failure gauge) ?

pkg/mapper/configmap/mapper.go
```diff
@@ -11,6 +12,27 @@ type ConfigMapMapper struct {
...
+			Help:      "Authenticator Watch Success",
+		}, []string{"result"}),
```
jyotimahapatra, line 30: what is the significance of string `result`?

pkg/mapper/configmap/mapper.go
jyotimahapatra: can the name be `apiServerWatch` ..`watch` is very generic

pkg/mapper/configmap/configmap.go
```diff
@@ -59,10 +66,12 @@ func (ms *MapStore) startLoadConfigMap(stopCh <-chan struct{}) {
 					FieldSelector: fields.OneTermEqualSelector("metadata.name", "aws-auth").String(),
 				})
 				if err != nil {
-					logrus.Warn("Unable to re-establish watch.  Sleeping for 5 seconds")
+					logrus.Errorf("Unable to re-establish watch: %v, sleeping for 5 seconds.", err)
+					metricsObj.watch.WithLabelValues(metricFailure).Set(metricFailureUnit)
```
jyotimahapatra, line 70: can you tell more why we need a label of success/failure here? isnt the gauge enough to determine the state?

pkg/mapper/configmap/configmap.go
jyotimahapatra: can we take care of this

jyotimahapatra (commented):
few comments

nckturner:
Replaced by #400

PR #395, opened 2021-09-16
Integration test framework

nckturner:
Framework to run integration tests:
- Uses kubernetes integration test framework
- Shallow clones k/k into a local directory for importing into tests to avoid complexity
  with generated code.
- Adds a test which adds a config map entry and attempts to authenticate.
- The test currently expects to be run on an EC2 instance because the authenticator server cannot start otherwise.  I'd like to get rid of this dependency.
- The diff currently does not include go.mod or vendor changes.  Should we continue vendoring dependencies?  I'm not sure what what the original reasoning for vendoring was but I'd like to avoid it if we don't have any hard dependencies on it.  
- I made some changes to config/certs and config/kubeconfig to make it more testable, and added pkg/testutils with helper functions to set up the test environment.


wongma7:
> The test currently expects to be run on an EC2 instance because the authenticator server cannot start otherwise. I'd like to get rid of this dependency.

why is this, do we need instance metadata for some reason?

.gitignore
```diff
@@ -1,6 +1,11 @@
 /dist
 /_output
 
+# For testing
+staging/k8s.io/kubernetes
```
wongma7, line 5: idk if "staging" makes sense here. I understand "staging" in k8s.io/kubernetes to mean that the code there is "staged" there before getting auto-mirrored to elsewhere. for example k8s.io/kubernetes is the source of truth for k8s.io/api so they do some go module magic to depend on k8s.io/api while also keeping the source of truth under staging dir . https://github.com/kubernetes/kubernetes/blob/master/go.mod#L482 .

some other name to indicate it is tmp for testing  would be better imo.. like testartifacts or sth.

cmd/aws-iam-authenticator/server.go
```diff
@@ -83,31 +83,31 @@ func init() {
 
 	serverCmd.Flags().String("generate-kubeconfig",
 		"/etc/kubernetes/aws-iam-authenticator/kubeconfig.yaml",
-		"Output `path` where a generated webhook kubeconfig (for `--authentication-token-webhook-config-file`) will be stored (should be a hostPath mount).")
+		"Output `path` where a generated webhook kubeconfig (for `--authentication-token-webhook-config-file`) will be stored.  When running as a container, this should probably be a hostPath mount.")
```
wongma7, line 86: why? this is assuming that apiserver is running on the same host and so can access this kubeconfig right? I would state that assumption, the point is to make the webhook readable by apiserver,  to further improve the comment

tests/integration/server/main_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+var (
+	authenticatorBinaryPath string
+	roleARN                 string
+)
+
+func TestMain(m *testing.M) {
+	flag.StringVar(&authenticatorBinaryPath, "authenticator-binary-path", "/usr/local/bin/aws-iam-authenticator", "Location of the aws-iam-authenticator binary to test with.")
+	flag.StringVar(&roleARN, "role-arn", "", "ARN of role to be authenticated.")
```
wongma7, line 33: the description could be better, my understanding is it should match the role of whoever is executing the test.  it would be cool if it defaulted to aws sts get-caller-identity.

tests/integration/server/server_test.go
```diff
@@ -0,0 +1,37 @@
...
+		AuthenticatorClientBinaryPath:   authenticatorBinaryPath,
+	})
+
+	t.Log("Creating aws-auth")
+	_, err := adminClient.CoreV1().ConfigMaps("kube-system").Create(context.TODO(), &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: "aws-auth"},
+		Data:       map[string]string{"mapRoles": fmt.Sprintf("    - rolearn: %s\n      username: system:node:{{EC2PrivateDNSName}}\n      groups:\n        - system:bootstrappers\n        - system:nodes", roleARN)},
+	}, metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("test failed: %v\n", err)
```
wongma7, line 29: golang will tell me the test failed, no need to prepend error with "test failed" . context would be more helpful , like "error creating configmap" or something.

pkg/testutils/testserver.go
```diff
@@ -0,0 +1,180 @@
...
+	AuthenticatorClientBinaryPath   string
+}
+
+func StartAuthenticatorTestFramework(t *testing.T, stopCh <-chan struct{}, setup AuthenticatorTestFrameworkSetup) (client.Interface, client.Interface) {
+	cfg, err := testConfig(t, setup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := cfg.GetOrCreateCertificate(); err != nil {
```
wongma7, line 46: where do these files get written? to pwd ?

pkg/testutils/testserver.go
```diff
@@ -0,0 +1,180 @@
...
+	partitionMap := map[string]endpoints.Partition{}
+	for _, p := range endpoints.DefaultPartitions() {
+		partitionMap[p.ID()] = p
+		partitionKeys = append(partitionKeys, p.ID())
+	}
+	if _, ok := partitionMap[cfg.PartitionID]; !ok {
+		return cfg, errors.New("Invalid partition")
+	}
+
+	if errs := mapper.ValidateBackendMode(cfg.BackendMode); len(errs) > 0 {
```
wongma7, line 157: is this and checking cfg.ClusterID necessary, shouldn't the server do these checks for us when you start it? when you call mapper.Start it should crap out if you provide an invalid one right?

pkg/testutils/testserver.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	adminClient, kubeAPIServerClientConfig := framework.StartTestServer(t, stopCh, framework.TestServerSetup{
+		ModifyServerRunOptions: func(opts *options.ServerRunOptions) {
+			opts.Authentication.WebHook.ConfigFile = cfg.GenerateKubeconfigPath
+		},
+		ModifyServerConfig: func(config *controlplane.Config) {},
+	})
+
+	cert, err := certs.LoadCertificate(kubeAPIServerClientConfig.TLSClientConfig.CAFile)
```
wongma7, line 61: what's the difference between LoadCertificate and LoadExistingCertificate, the naming is no good. how can it pass a CA and use that as a server cert?

wongma7 (commented):
I don't understand the cert loading changes otherwise lgtm

nckturner:
> why is this, do we need instance metadata for some reason?

I should say we need some type of AWS credentials for the initialization of the [ec2Provider](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/server/server.go#L176), if I remember correctly.  Haven't looked too closely at the dependency yet though.

tests/integration/server/main_test.go
nckturner: It is the test runner script.

pkg/testutils/testserver.go
nckturner: I renamed these, one of them is a certificate without the key, and the other is a keypair.  Terrible naming.

pkg/testutils/testserver.go
nckturner: They get written to the "stateDir".

pkg/testutils/testserver.go
nckturner: It calls `ValidateBackendMode()` in `cmd/aws-iam-authenticator/root.go`, so while I'm not sure if its strictly necessary for a test server, I'm just copying the `getConfig()` function there.

nckturner: I would like to move `BuildMapperChain()` from cmd to pkg, as I'm not using cmd in the integration tests at this point and I'd rather not.  Maybe I can do the same for `ValidateBackendMode()`.

PR #396, opened 2021-09-16
Fix deps

### LINKED ISSUE #383: Update deps
Hello!  This package depends on a very old version of cobra (both directly and indirectly) which causes users to use a very old version of Prometheus.  I was able to hack around this as a POC with [this commit](https://github.com/ZipRecruiter/aws-iam-authenticator/commit/e09b9a5cf0509b9410ce0685be9d3c28d18ce067).  Would this project be willing to accept a subset of these changes?  I assume removing all of the vendoring is not an option, I just did that to keep things simple.  But upgrading deps and removing the use of go.hein.dev/go-version (and instead just doing a subset of what it does) would be very helpful.

Let me know and if this will work for the project I'll submit a PR.

frioux:
This program ends up depending on go.hein.dev/go-version, which depends on a really old version of cobra.  I removed that package and reimplemented it's functionality to allow upgrading deps.  At the same time I removed all the vendored stuff, but I'd be willing to not do that part, it was just the easiest way for me to get this to work.

As a side note: I also submitted [a PR to go.hein.dev/go-version that would fix this](https://github.com/christopherhein/go-version/pull/34) once merged and this package updates to that version.

Fixes #383


frioux:
FYI I don't intend to rebase this again unless a human asks me to.  Seems like it's just being ignored.

nckturner:
If you want to update this PR to not remove all the vendored deps, I am willing to merge it. Fewer dependencies is a good thing and I don't think we need a separate package for a version command.

frioux:
ok, I'll do that soon 👍

frioux:
@nckturner done!

nckturner (approved):

PR #397, opened 2021-09-26
Upgrade CRD manifest to v1

### LINKED ISSUE #384: Upgrade the iamidentitymappings CRD to support apiextensions.k8s.io/v1
What happens?
-------------
The [iamidentitymappings](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/deploy/iamidentitymapping.yaml) CRD was created with version `apiextensions.k8s.io/v1beta1` of the api, which means that after version 1.22 of Kubernetes [it can no longer be used](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122). 

Expecting to happen
----------------------------------
Both kubernetes resources (including CRD) and aws-iam-authenticator source code supports version apiextensions.k8s.io/v1

The main problem is that after reviewing the code I see that the golang v1alpha1 kubernetes api package is widely used, so I would like to know what is your feedback on this change. If it doesn't affect anything other than the CRD I'd be happy to collaborate with a solution. 

Logs
----------------------------
```
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
```

Versions of software
----------------------------------------
**Toolchain Version:**
Kubernetes: v1.22.0-beta.0

rifelpet:
Fixes #384

### DIFF
--- a/deploy/iamidentitymapping.yaml
+++ b/deploy/iamidentitymapping.yaml
@@ -1,11 +1,10 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: iamidentitymappings.iamauthenticator.k8s.aws
 spec:
   group: iamauthenticator.k8s.aws
-  version: v1alpha1
   scope: Cluster
   names:
     plural: iamidentitymappings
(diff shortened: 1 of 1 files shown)


nckturner:
Awesome, thanks @rifelpet!

PR #399, opened 2021-09-28
Move inactives to emeritus_approvers and add active users

nckturner:
- Move inactives to emeritus
- Add wongma7 to reviewers because he was only in approvers
- Add Jay to approvers and reviewers


nckturner:
Let me remove Jay, and after he becomes a member we can add him.

PR #410, opened 2021-11-02
Fix panic when cache file can't be Stat-ed

### LINKED ISSUE #349: Crash when cache file has no permissions
When running the program to get the token, it crashes with sigsegv in case when
the cache file exists, but there are no permissions to read its parent
directory:

```
$ AWS_PROFILE=xxx aws-iam-authenticator token --region yyy --cluster-id xxx-eks --cache
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x15f1fbe]

goroutine 1 [running]:
sigs.k8s.io/aws-iam-authenticator/pkg/token.NewFileCacheProvider(0x7fffffffe84a, 0xe, 0xc0000442ec, 0xa, 0x0, 0x0, 0xc000414c00, 0x0, 0x0, 0x0, ...)
        /tmp/aws-iam-authenticator/pkg/token/filecache.go:189 +0x3fe
sigs.k8s.io/aws-iam-authenticator/pkg/token.generator.GetWithOptions(0x7ffff7d70100, 0xc00007c900, 0x0, 0x0, 0x0, 0x0, 0x0, 0x40d1af, 0xc00019fc50)
        /tmp/aws-iam-authenticator/pkg/token/token.go:253 +0x911
main.glob..func3(0x2797620, 0xc00007e820, 0x0, 0x5)
        /tmp/aws-iam-authenticator/cmd/aws-iam-authenticator/token.go:64 +0x366
github.com/spf13/cobra.(*Command).execute(0x2797620, 0xc00007e7d0, 0x5, 0x5, 0x2797620, 0xc00007e7d0)
        /home/wolf/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:830 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0x2797120, 0x1a5870b, 0x4, 0x1a866fd)
        /home/wolf/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x30b
github.com/spf13/cobra.(*Command).Execute(...)
        /home/wolf/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.Execute()
        /tmp/aws-iam-authenticator/cmd/aws-iam-authenticator/root.go:50 +0x31
main.main()
        /tmp/aws-iam-authenticator/cmd/aws-iam-authenticator/root.go:45 +0x25
```

That is bit subpar user experience and without looking into source code it's
imposible to figure out what is going on.

Fix seems to be fairly trivial, diff is 

```
$ g d --stat
 pkg/token/filecache.go | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)
```

Since I do not want to accept CLA, I cannot provide the diff itself.

sarahhodne:
The code currently assumes that Stat always return a `FileInfo` object,
but if the file can't be `Stat`-ed (e.g. due to a permissions error), `Stat`
can return a nil `FileInfo`, plus an error. The error is only handled if
it's a "not exists" error, but there are other errors that can be
returned too.

This wasn't caught by the tests, since the mocked out `Stat` always
returns a `FileInfo`, even when an error is returned.

This commit also changes from `os.IsNotExist(err)` to `errors.Is(err, fs.ErrNotExist)` as recommended by [the documentation for `os.IsNotExist`](https://pkg.go.dev/os#IsNotExist).

Fixes #349.

### DIFF
--- a/pkg/token/filecache.go
+++ b/pkg/token/filecache.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/gofrs/flock"
 	"gopkg.in/yaml.v2"
+	"io/fs"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -185,7 +186,7 @@ func NewFileCacheProvider(clusterID, profile, roleARN string, creds *credentials
 	cachedCredential := cachedCredential{}
 	// ensure path to cache file exists
 	_ = f.MkdirAll(filepath.Dir(filename), 0700)
-	if info, err := f.Stat(filename); !os.IsNotExist(err) {
+	if info, err := f.Stat(filename); err == nil {
 		if info.Mode()&0077 != 0 {
 			// cache file has secret credentials and should only be accessible to the user, refuse to use it.
 			return FileCacheProvider{}, fmt.Errorf("cache file %s is not private", filename)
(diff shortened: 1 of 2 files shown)


nckturner:
Thanks @sarahhodne, nice fix.

sarahhodne:
The test failure doesn't seem related to my change, and I can't seem to reproduce it locally, so I think it might be a flake.

/retest

nckturner:
Agreed.  If it fails again I'll take a closer look.

sarahhodne:
Looks like it was a flake, so opened a ticket to track that, since I think the bot said that in the last message (though that appears to have been deleted now): https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/413

Issue #333, opened 2020-08-13
Mapping AWS SSO managed roles to K8s permissions?


christophetd:
Hello,

I'm using AWS SSO to manage roles and permissions in several AWS accounts. Consequently, my roles are autoprovisioned in each AWS account by AWS SSO and their names look like **AWSReservedSSO_SomeRole_e11aa594f712e1dc**.

What would be the best way to manage this with aws-iam-authenticator? The current solution seems to force to specify the full role name, so the configuration looks like:

```
data:
  mapRoles: |
   groups:
        - system:masters
        rolearn: arn:aws:iam::xxx:role/AWSReservedSSO_SomeRole_e11aa594f712e1dc
        username: kubernetes-admin:{{SessionName}}
```

The issue is that the name of the role changes in each AWS account, so it makes things quite hard for automation. Do you have any guideline to suggest in this context?

AWS SSO is becoming more and more prevalent so I expect it should be a quite widely encountered issue

Thank you!

christophetd:
@nckturner (who seem to be the main repo contributor), do you have any insights on how this can be achieved in the current state? Otherwise, do you have some plans to support it? Thank you!

callum-p:
I have separate dev/prod/etc accounts. To achieve this I do something like this:

```sh
export ADMIN_ROLE=$(awscli iam list-roles --query 'Roles[?starts_with(RoleName, `AWSReservedSSO_KubernetesAdmin`) == `true`].RoleName' --output text)
export READ_ONLY_ROLE=$(awscli iam list-roles --query 'Roles[?starts_with(RoleName, `AWSReservedSSO_KubernetesReadOnly`) == `true`].RoleName' --output text)
export ACCOUNT_ID=$(awscli sts get-caller-identity --query Account --output text)

envsubst < specs/configmap.tpl.yml > specs/configmap.yml
```

Then the configmap:

```
---
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: kube-system
  name: aws-iam-authenticator
  labels:
    k8s-app: aws-iam-authenticator
data:
  config.yaml: |
    clusterID: $K8S_CLUSTER_NAME
    server:
      #  2) "{{SessionName}}" is the role session name.
      mapRoles:
      - roleARN: arn:aws:iam::${ACCOUNT_ID}:role/${ADMIN_ROLE}
        username: "{{SessionName}}"
        groups:
        - system:masters
[...]

PR #416, opened 2021-12-23
Add SSO Role suffix support

logandavies181:
This PR adds the ability to match AWS SSO roles without needing to know the random suffix.

This solves the issue of teams that use AWS SSO roles and aws-iam-authenticator (including EKS users) from needing to keep their configurations up-to-date with the random suffixes that AWS SSO applies to roles it creates. This also has the side-effect of making SSO roles more intuitive to work with.

This PR solves this issue:
 https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/333

In earlier commits, this PR also solved the below roadmap issue, as it allowed for full wildcard support, constrained via ArnLike condition checks.
https://github.com/aws/containers-roadmap/issues/474

A new nested type has been added to the mapRoles elements to specify a configuration for matching SSO created roles. 
This looks like:
```
  mapRoles: |
  - sso:
      permissionSetName: MyPermissionSet
      accountID: "000000000000"
    username: foo
    groups:
    - bar
    - baz
```

Changes:
- add.go has a new option for creating entries in the configmap for the new mapping subtype
- a new feature gate has been added to control whether this new feature is active or not
- a new package has been added to provide a robust way to perform ArnLike matching of ARNs
- config.UserMapping and config.RoleMapping have had methods added and their usage in filemapper and configmapper has been refactored to reduce duplication of logic
- greater unit test coverage 😎
- new type configmap.SSOARNMatcher for containing the new configuration and specifying new yaml keys

### DIFF
--- a/cmd/aws-iam-authenticator/add.go
+++ b/cmd/aws-iam-authenticator/add.go
@@ -42,7 +42,7 @@ var addUserCmd = &cobra.Command{
 	Long:  "NOTE: this does not currently support the CRD and file backends",
 	Run: func(cmd *cobra.Command, args []string) {
 		if userARN == "" || userName == "" || len(groups) == 0 {
-			fmt.Printf("invalid empty value in userARN %q, username %q, groups %q", userARN, userName, groups)
+			fmt.Printf("invalid empty value in userARN %q, username %q, groups %q\n", userARN, userName, groups)
 			os.Exit(1)
 		}
 
@@ -73,16 +73,52 @@ var addRoleCmd = &cobra.Command{
 	Short: "add a role entity to an existing aws-auth configmap, not for CRD/file backends",
 	Long:  "NOTE: this does not currently support the CRD and file backends",
 	Run: func(cmd *cobra.Command, args []string) {
-		if roleARN == "" || userName == "" || len(groups) == 0 {
-			fmt.Printf("invalid empty value in rolearn %q, username %q, groups %q", roleARN, userName, groups)
+		if (roleARN == "" && ssoRole == nil) || userName == "" || len(groups) == 0 {
+			fmt.Printf("invalid empty value in rolearn %q, username %q, groups %q\n", roleARN, userName, groups)
 			os.Exit(1)
 		}
 
-		checkPrompt(fmt.Sprintf("add rolearn %s, username %s, groups %s", roleARN, userName, groups))
+		var arnOrSSORole string
+		switch {
+		case roleARN != "" && ssoRole != nil:
+			fmt.Printf("only one of --rolearn or --sso can be supplied\n")
+			os.Exit(1)
+		case roleARN != "":
+			arnOrSSORole = "rolearn"
+		case ssoRole != nil:
+			arnOrSSORole = "sso"
+
+			for _, key := range []string{"permissionSetName", "accountID"} {
+				if _, ok := ssoRole[key]; !ok {
+					fmt.Printf("required key '%s' missing from --sso flag\n", key)
+					os.Exit(1)
+				}
+			}
+
+			var ssoPartition string
+			if partition, ok := ssoRole["partition"]; !ok {
+				ssoPartition = "aws"
+			} else {
+				ssoPartition = partition
+			}
+			ssoRoleConfig.PermissionSetName = ssoRole["permissionSetName"]
+			ssoRoleConfig.AccountID = ssoRole["accountID"]
+			ssoRoleConfig.Partition = ssoPartition
+
+			rm := config.RoleMapping{SSO: ssoRoleConfig}
+			err := rm.Validate()
+			if err != nil {
+				fmt.Printf("error validating --sso: %s\n", err)
+				os.Exit(1)
+			}
+		}
+
+		checkPrompt(fmt.Sprintf("add %s %s, username %s, groups %s", arnOrSSORole, roleARN, userName, groups))
 		cli := createClient()
 
 		cm, err := cli.AddRole(&config.RoleMapping{
 			RoleARN:  roleARN,
+			SSO:      ssoRoleConfig,
 			Username: userName,
 			Groups:   groups,
 		})
@@ -174,6 +210,10 @@ var (
 	userName string
 	groups   []string
 	roleARN  string
+	// ssoRole contains the settings for a config.SSOARNMatcher
+	// it expects the keys "permissionSetName", "accountID", and "partition" (optional)
+	ssoRole       map[string]string
+	ssoRoleConfig *config.SSOARNMatcher
 )
 
 func init() {
@@ -191,6 +231,7 @@ func init() {
 	addUserCmd.PersistentFlags().StringSliceVar(&groups, "groups", nil, "A new user groups")
 
 	addRoleCmd.PersistentFlags().StringVar(&roleARN, "rolearn", "", "A new role ARN")
+	addRoleCmd.PersistentFlags().StringToStringVar(&ssoRole, "sso", nil, `Settings for a new SSO role. Expects "permissionSetName", "accountID", and "partition" (optional)`)
 	addRoleCmd.PersistentFlags().StringVar(&userName, "username", "", "A new user name")
 	addRoleCmd.PersistentFlags().StringSliceVar(&groups, "groups", nil, "A new role groups")
 }
--- a/docs/sso_role_matcher.md
+++ b/docs/sso_role_matcher.md
@@ -1,0 +1,53 @@
+# SSO Role Matcher
+
+Maps configuration for an AWS SSO managed IAM Role to a Kubernetes username and groups.
+
+## Feature state
+
+Alpha
+
+## Use case
+
+Easy and robust configuration for AWS SSO managed roles, which currently have two main issues:
+
+- Confusing configuration. To use an SSO role, a user needs to map the Role ARN of the SSO ROle, minus the path.
+
+For example: given a permission set `MyPermissionSet`, region `us-east-1` and account number `000000000000`; AWS SSO
+creates a role: `arn:aws:iam::000000000000:role/aws-reserved/sso.amazonaws.com/us-east-1/AWSReservedSSO_MyPermissionSet_1234567890abcde`.
+
+To match this role, a user would need to create a mapRoles entry like:
+```
+  mapRoles: |
+  - rolearn: arn:aws:iam::000000000000:role/AWSReservedSSO_MyPermissionSet_1234567890abcde
+    username: ...
+    groups: ...
+```
+
+- Brittle configuration. If AWS SSO recreates IAM Roles, they receive a different random suffix and all the users of that
+role can no longer authenticate.
+
+## New UX
+
+Users can create a mapRoles entry that will automatically match roles created by AWS SSO without needing to be updated
+every time the roles are changed.
+
+Users will now create mapRoles entries like:
+```
+  mapRoles: |
+  - sso:
+      permissionSetName: MyPermissionSet
+      accountID: "000000000000"
+    username: ...
+    groups: ...
+```
+
+If the user is using the aws-us-govt or aws-cn partitions, they must specify the partition attribute in the `sso` structure.
+
+## Implementation
+
+config.RoleMapping will be extended with a nested structure containing the necessary information to construct a canonicalized
+Role Arn. The random suffix will not need to be specified and will instead be matched for the user by constructing the
+expect ARN and applying a wildcard to the end.
+
+Users are protected from non-AWS SSO created roles as the AWS API prevents roles being manually created with AWSReservedSSO
+at the beginning of their names.
--- a/docs/sso_role_matcher.md
+++ b/docs/sso_role_matcher.md
@@ -10,7 +10,7 @@ Alpha
 
 Easy and robust configuration for AWS SSO managed roles, which currently have two main issues:
 
-- Confusing configuration. To use an SSO role, a user needs to map the Role ARN of the SSO ROle, minus the path.
+Firstly - confusing configuration. To use an SSO role, a user needs to map the Role ARN of the SSO ROle, minus the path.
 
 For example: given a permission set `MyPermissionSet`, region `us-east-1` and account number `000000000000`; AWS SSO
 creates a role: `arn:aws:iam::000000000000:role/aws-reserved/sso.amazonaws.com/us-east-1/AWSReservedSSO_MyPermissionSet_1234567890abcde`.
@@ -23,8 +23,8 @@ To match this role, a user would need to create a mapRoles entry like:
     groups: ...
 ```
 
-- Brittle configuration. If AWS SSO recreates IAM Roles, they receive a different random suffix and all the users of that
-role can no longer authenticate.
+Secondly - brittle configuration. If AWS SSO recreates IAM Roles, they receive a different random suffix and all the users of that
+role can no longer authenticate to Kubernetes.
 
 ## New UX
 
@@ -42,12 +42,21 @@ Users will now create mapRoles entries like:
 ```
 
 If the user is using the aws-us-govt or aws-cn partitions, they must specify the partition attribute in the `sso` structure.
+```
+  mapRoles: |
+  - sso:
+      permissionSetName: MyPermissionSet
+      accountID: "000000000000"
+      partition: "aws-us-govt"
+    username: ...
+    groups: ...
+```
 
 ## Implementation
 
 config.RoleMapping will be extended with a nested structure containing the necessary information to construct a canonicalized
 Role Arn. The random suffix will not need to be specified and will instead be matched for the user by constructing the
-expect ARN and applying a wildcard to the end.
+expected ARN and applying a wildcard to the end.
 
 Users are protected from non-AWS SSO created roles as the AWS API prevents roles being manually created with AWSReservedSSO
 at the beginning of their names.
--- a/pkg/arn/arnlike.go
+++ b/pkg/arn/arnlike.go
@@ -1,0 +1,104 @@
+package arn
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+const (
+	arnDelimiter        = ":"
+	arnSectionsExpected = 6
+	arnPrefix           = "arn:"
+
+	// zero-indexed
+	sectionPartition = 1
+	sectionService   = 2
+	sectionRegion    = 3
+	sectionAccountID = 4
+	sectionResource  = 5
+
+	// errors
+	invalidPrefix   = "invalid prefix"
+	invalidSections = "not enough sections"
+)
+
+// ArnLike takes an ARN and returns true if it is matched by the pattern.
+// Each component of the ARN is matched individually as per
+// https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_ARN
+func ArnLike(arn, pattern string) (bool, error) {
+	// "parse" the input arn into sections
+	arnSections, err := parse(arn)
+	if err != nil {
+		return false, fmt.Errorf("Could not parse input arn: %v", err)
+	}
+	patternSections, err := parse(pattern)
+	if err != nil {
+		return false, fmt.Errorf("Could not parse ArnLike string: %v", err)
+	}
+
+	// Tidy regexp special characters. Escape the ones not used in ArnLike.
+	// Replace multiple * with .* - we're assuming `\` is not allowed in ARNs
+	preparePatternSections(patternSections)
+
+	for index := range arnSections {
+		patternGlob, err := regexp.Compile(patternSections[index])
+		if err != nil {
+			return false, fmt.Errorf("Could not parse %s: %v", patternSections[index], err)
+		}
+
+		if !patternGlob.MatchString(arnSections[index]) {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+// parse is a copy of arn.Parse from the AWS SDK but represents the ARN as []string
+func parse(input string) ([]string, error) {
+	if !strings.HasPrefix(input, arnPrefix) {
+		return nil, fmt.Errorf(invalidPrefix)
+	}
+	arnSections := strings.SplitN(input, arnDelimiter, arnSectionsExpected)
+	if len(arnSections) != arnSectionsExpected {
+		return nil, fmt.Errorf(invalidSections)
+	}
+
+	return arnSections, nil
+}
+
+// preparePatternSections goes through each section of the arnLike slice and escapes any meta characters, except for
+// `*` and `?` which are replaced by `.*` and `.?` respectively. ^ and $ are added as we require an exact match
+func preparePatternSections(arnLikeSlice []string) {
+	for index, section := range arnLikeSlice {
+		quotedString := quoteMeta(section)
+		arnLikeSlice[index] = `^` + quotedString + `$`
+	}
+}
+
+// the below is based on regexp.QuoteMeta to escape metacharacters except for `?` and `*`, changing them to `*` and `.*`
+
+// quoteMeta returns a string that escapes all regular expression metacharacters
+// inside the argument text; the returned string is a regular expression matching
+// the literal text.
+func quoteMeta(s string) string {
+	const specialChars = `\.+()|[]{}^$`
+
+	var i int
+	b := make([]byte, 2*len(s)-i)
+	copy(b, s[:i])
+	j := i
+	for ; i < len(s); i++ {
+		if strings.Contains(specialChars, s[i:i+1]) {
+			b[j] = '\\'
+			j++
+		} else if s[i] == '*' || s[i] == '?' {
+			b[j] = '.'
+			j++
+		}
+		b[j] = s[i]
+		j++
+	}
+	return string(b[:j])
+}
--- a/pkg/arn/arnlike_test.go
+++ b/pkg/arn/arnlike_test.go
@@ -89,6 +89,20 @@ func TestArnLikeNetagiveMatches(t *testing.T) {
 	}
 }
 
+func TestIncompleteArnLikePattern(t *testing.T) {
+	incompleteArnLikePattern := "arn:*"
+	validArn := `arn:aws:iam::000000000000:role/some-role`
+
+	ok, err := ArnLike(validArn, incompleteArnLikePattern)
+	if ok {
+		t.Errorf("Expected false result on error for input arn: %s, pattern: %s", incompleteArnLikePattern, validArn)
+	}
+	expectedErrorText := "Could not parse ArnLike string: not enough sections"
+	if !strings.EqualFold(expectedErrorText, err.Error()) {
+		t.Errorf("Did not receive expected error text. Expected: '%s', got: '%s'", expectedErrorText, err.Error())
+	}
+}
+
 func TestArnLikeInvalidArns(t *testing.T) {
 	invalidPrefixArn := `nar:aws:iam::000000000000:role/some-role`
 	invalidSectionsArn := `arn:aws:iam:000000000000:role/some-role`
--- a/pkg/arn/arnlike_test.go
+++ b/pkg/arn/arnlike_test.go
@@ -1,0 +1,154 @@
+package arn
+
+import (
+	"strings"
+	"testing"
+)
+
+type arnLikeInput struct {
+	arn, pattern string
+}
+
+type quoteMetaInput struct {
+	input, expected string
+}
+
+func TestArnLikePostiveMatches(t *testing.T) {
+	inputs := []arnLikeInput{
+		{
+			arn:     `arn:aws:iam::000000000000:role/some-role`,
+			pattern: `arn:aws:iam::000000000000:role/some-role`,
+		},
+		{
+			arn:     `arn:aws:iam::000000000000:role/some-role`,
+			pattern: `arn:aws:iam::000000000000:*`,
+		},
+		{
+			arn:     `arn:aws:iam::000000000000:role/some-role`,
+			pattern: `arn:*:*:*:*:*`,
+		},
+		{
+			arn:     `arn:aws:iam::000000000000:role/some-role`,
+			pattern: `arn:aws:iam::000000000000:**`,
+		},
+		{
+			arn:     `arn:aws:iam::000000000000:role/some-role`,
+			pattern: `arn:aws:iam::000000000000:*role*`,
+		},
+		{
+			arn:     `arn:aws:iam::000000000000:role/some-role`,
+			pattern: `arn:aws:iam::000000000000:ro*`,
+		},
+		{
+			arn:     `arn:aws:iam::000000000000:role/some-role`,
+			pattern: `arn:aws:iam::000000000000:??????????????`,
+		},
+		{
+			arn:     `arn:aws:testservice::000000000000:some/wacky-new-[resource]{with}\metacharacters`,
+			pattern: `arn:aws:testservice::000000000000:some/wacky-new-[resource]{with}\metacharacters`,
+		},
+		{
+			arn:     `arn:aws:testservice::000000000000:some/wacky-new-[resource]{with}\metacharacters`,
+			pattern: `arn:aws:testservice::000000000000:some/wacky-new-[reso*`,
+		},
+	}
+
+	for _, v := range inputs {
+		ok, err := ArnLike(v.arn, v.pattern)
+		if err != nil {
+			t.Errorf("Expected no error for input arn: %s pattern: %s", v.arn, v.pattern)
+		}
+
+		if !ok {
+			t.Errorf("Expected true for input arn: %s pattern: %s", v.arn, v.pattern)
+		}
+	}
+}
+
+func TestArnLikeNetagiveMatches(t *testing.T) {
+	inputs := []arnLikeInput{
+		{
+			arn:     `arn:aws:iam::111111111111:role/some-role`,
+			pattern: `arn:aws:iam::000000000000:role/some-role`,
+		},
+		{
+			arn:     `arn:aws:testservice::000000000000:some/wacky:resource:with:colon:delims`,
+			pattern: `arn:aws:testservice::**:delims`,
+		},
+	}
+
+	for _, v := range inputs {
+		ok, err := ArnLike(v.arn, v.pattern)
+		if err != nil {
+			t.Errorf("Expected no error for input arn: %s pattern: %s", v.arn, v.pattern)
+		}
+
+		if ok {
+			t.Errorf("Expected false for input arn: %s pattern: %s", v.arn, v.pattern)
+		}
+	}
+}
+
+func TestArnLikeInvalidArns(t *testing.T) {
+	invalidPrefixArn := `nar:aws:iam::000000000000:role/some-role`
+	invalidSectionsArn := `arn:aws:iam:000000000000:role/some-role`
+	validArn := `arn:aws:iam::000000000000:role/some-role`
+
+	// invalid prefix
+	ok, err := ArnLike(invalidPrefixArn, validArn)
+	if ok {
+		t.Errorf("Expected false result on error for input arn: %s, pattern: %s", invalidPrefixArn, validArn)
+	}
+
+	expectedErrorText := "Could not parse input arn: invalid prefix"
+	if !strings.EqualFold(expectedErrorText, err.Error()) {
+		t.Errorf("Did not receive expected error text. Expected: '%s', got: '%s'", expectedErrorText, err.Error())
+	}
+
+	// invalid sections
+	ok, err = ArnLike(invalidSectionsArn, validArn)
+	if ok {
+		t.Errorf("Expected false result on error for input arn: %s, pattern: %s", invalidSectionsArn, validArn)
+	}
+
+	expectedErrorText = "Could not parse input arn: not enough sections"
+	if !strings.EqualFold(expectedErrorText, err.Error()) {
+		t.Errorf("Did not receive expected error text. Expected: '%s', got: '%s'", expectedErrorText, err.Error())
+	}
+}
+
+func TestQuoteMeta(t *testing.T) {
+	inputs := []quoteMetaInput{
+		{
+			input:    `**`,
+			expected: `.*.*`,
+		},
+		{
+			input:    `??`,
+			expected: `.?.?`,
+		},
+		{
+			input:    `abdcEFG`,
+			expected: `abdcEFG`,
+		},
+		{
+			input:    `abd.EFG`,
+			expected: `abd\.EFG`,
+		},
+		{
+			input:    `\.+()|[]{}^$`,
+			expected: `\\\.\+\(\)\|\[\]\{\}\^\$`,
+		},
+		{
+			input:    `\.+()|[]{}^$*?`,
+			expected: `\\\.\+\(\)\|\[\]\{\}\^\$.*.?`,
+		},
+	}
+
+	for _, v := range inputs {
+		output := quoteMeta(v.input)
+		if !strings.EqualFold(v.expected, output) {
+			t.Errorf("Did not get expected output from quoteMeta. Expected: '%s', got: '%s'", v.expected, output)
+		}
+	}
+}
--- a/pkg/config/mapper.go
+++ b/pkg/config/mapper.go
@@ -50,7 +50,7 @@ func (m *RoleMapping) Validate() error {
 		}
 
 		switch m.SSO.Partition {
-		case "aws", "aws-cn", "aws-us-gov":
+		case "aws", "aws-cn", "aws-us-gov", "aws-iso", "aws-iso-b":
 			// valid
 		case "":
 			// treated as "aws"
--- a/pkg/config/mapper.go
+++ b/pkg/config/mapper.go
@@ -1,0 +1,125 @@
+package config
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
+
+	"github.com/sirupsen/logrus"
+)
+
+// SSOArnLike returns a string that can be passed to arnlike.ArnLike to
+// match canonicalized IAM Role ARNs against. Assumes Validate() has been called.
+func (m *RoleMapping) SSOArnLike() string {
+	if m.SSO == nil {
+		return ""
+	}
+
+	var partition string
+	if m.SSO.Partition == "" {
+		partition = "aws"
+	}
+
+	return strings.ToLower(fmt.Sprintf("arn:%s:iam::%s:role/AWSReservedSSO_%s_*", partition, m.SSO.AccountID, m.SSO.PermissionSetName))
+}
+
+// Validate returns an error if the RoleMapping is not valid after being unmarshaled
+func (m *RoleMapping) Validate() error {
+	if m == nil {
+		return fmt.Errorf("RoleMapping is nil")
+	}
+
+	if m.RoleARN == "" && m.SSO == nil {
+		return fmt.Errorf("One of rolearn or SSO must be supplied")
+	} else if m.RoleARN != "" && m.SSO != nil {
+		return fmt.Errorf("Only one of rolearn or SSO can be supplied")
+	}
+
+	if m.SSO != nil {
+		accountIDRegexp := regexp.MustCompile("^[0-9]{12}$")
+		if !accountIDRegexp.MatchString(m.SSO.AccountID) {
+			return fmt.Errorf("AccountID '%s' is not a valid AWS Account ID", m.SSO.AccountID)
+		}
+
+		// https://docs.aws.amazon.com/singlesignon/latest/APIReference/API_PermissionSet.html
+		permissionSetNameRegexp := regexp.MustCompile(`^[\w+=,.@-]{1,32}$`)
+		if !permissionSetNameRegexp.MatchString(m.SSO.PermissionSetName) {
+			return fmt.Errorf("PermissionSetName '%s' is not a valid AWS SSO PermissionSet Name", m.SSO.PermissionSetName)
+		}
+
+		switch m.SSO.Partition {
+		case "aws", "aws-cn", "aws-us-gov":
+			// valid
+		case "":
+			// treated as "aws"
+		default:
+			return fmt.Errorf("Partition '%s' is not a valid AWS partition", m.SSO.Partition)
+		}
+
+		ssoArnLikeString := m.SSOArnLike()
+		ok, err := arn.ArnLike(ssoArnLikeString, "arn:*:iam:*:*:role/*")
+		if err != nil {
+			return fmt.Errorf("SSOArnLike '%s' is not valid: %v", ssoArnLikeString, err)
+		} else if !ok {
+			return fmt.Errorf("SSOArnLike '%s' did not match an ARN for a canonicalized IAM Role", ssoArnLikeString)
+		}
+	}
+
+	return nil
+}
+
+// Matches returns true if the supplied ARN or SSO settings matches
+// this RoleMapping
+func (m *RoleMapping) Matches(subject string) bool {
+	if m.RoleARN != "" {
+		return strings.ToLower(m.RoleARN) == strings.ToLower(subject)
+	}
+
+	// Assume the caller has called Validate(), which parses m.RoleARNLike
+	// If subject is not parsable, then it cannot be a valid ARN anyway so
+	// we can ignore the error here
+	var ok bool
+	if SSORoleMatchEnabled {
+		var err error
+		ok, err = arn.ArnLike(subject, m.SSOArnLike())
+		if err != nil {
+			logrus.Error("Could not parse subject ARN: ", err)
+		}
+	}
+	return ok
+}
+
+// Key returns RoleARN or SSOArnLike(), whichever is not empty.
+// Used to get a Key name for map[string]RoleMapping
+func (m *RoleMapping) Key() string {
+	if m.RoleARN != "" {
+		return strings.ToLower(m.RoleARN)
+	}
+	return m.SSOArnLike()
+}
+
+// Validate returns an error if the UserMapping is not valid after being unmarshaled
+func (m *UserMapping) Validate() error {
+	if m == nil {
+		return fmt.Errorf("UserMapping is nil")
+	}
+
+	if m.UserARN == "" {
+		return fmt.Errorf("Value for userarn must be supplied")
+	}
+
+	return nil
+}
+
+// Matches returns true if the supplied ARN string matche this UserMapping
+func (m *UserMapping) Matches(subject string) bool {
+	return strings.ToLower(m.UserARN) == strings.ToLower(subject)
+}
+
+// Key returns UserARN.
+// Used to get a Key name for map[string]UserMapping
+func (m *UserMapping) Key() string {
+	return m.UserARN
+}
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -44,28 +44,55 @@ type IdentityMapping struct {
 // You can use plain values without parameters to have a more static mapping.
 type RoleMapping struct {
 	// RoleARN is the AWS Resource Name of the role. (e.g., "arn:aws:iam::000000000000:role/Foo").
-	RoleARN string `json:"rolearn"`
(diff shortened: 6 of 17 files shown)


logandavies181:
I signed it

moretea:
@logandavies181 is there any way that I can assist you with e.g. testing? I'd prefer that above writing a cronjob that would sync the list of roles to the CRD's every few minutes.

logandavies181:
@moretea Thanks for the offer. I'm all good RE: assistance and I've got plenty of free time to complete this, but I'm waiting for a reviewer to indicate that this contribution would actually be considered before putting the effort into implementing the last pieces.

Just giving a bit more grace period before I start making noise as it's just been the Christmas/New Year period

logandavies181:
@jaypipes @nckturner would be really nice to get some feedback on this, please. I believe it would solve https://github.com/aws/containers-roadmap/issues/474 which has a lot of interest

pkg/mapper/configmap/configmap.go
```diff
@@ -25,9 +26,10 @@ import (
 )
 
 type MapStore struct {
-	mutex sync.RWMutex
-	users map[string]config.UserMapping
-	roles map[string]config.RoleMapping
+	mutex           sync.RWMutex
+	users           map[string]config.UserMapping
+	roles           map[string]config.RoleMapping
+	arnLikeMappings []config.ARNLikeMapping
```
jaypipes, line 32: To be consistent with the field naming for the other maps (`users` and `roles`), consider naming this field `arnLikes`

pkg/config/types.go
```diff
@@ -68,6 +68,21 @@ type UserMapping struct {
...
+	ARNLike string `json:"arnLike"`
+
+	// Username is the username pattern that principals matching this
+	// string will have in Kubernetes.
+	Username string `json:"username"`
+
+	// Groups is a list of Kubernetes groups principals matching this string will
+	// authenticate as (e.g., `system:masters`)
+	Groups []string `json:"groups"`
+}
```
jaypipes, line 84: After reading through this PR a number of times, it struck me that it might be better to add a `UserMapping.UserARNLike` and a `RoleMapping.RoleARNLike` field instead of adding in a new `ARNLikeMapping` struct to the configuration. This would still be backwards-compatible with older configurations and more neatly nest the ARNLike functionality into the existing configuration blocks (for `users` and `roles` respectively)

It would also mean that you would not need to modify function signatures and return values for things like `add()` and `ParseMap()`

pkg/mapper/configmap/configmap_test.go
```diff
@@ -266,7 +266,7 @@ func TestParseMap(t *testing.T) {
 	}
 	accounts := []string{}
 
-	u, r, a, err := ParseMap(m1)
+	u, r, _, a, err := ParseMap(m1)
```
jaypipes, line 269: Please add a unit test to validate configuration file parsing for the newly-added config elements.

pkg/mapper/crd/apis/iamauthenticator/v1alpha1/types.go
```diff
@@ -35,9 +35,10 @@ type IAMIdentityMapping struct {
 
 // IAMIdentityMappingSpec is the spec for a IAMIdentityMapping resource
 type IAMIdentityMappingSpec struct {
-	ARN      string   `json:"arn"`
-	Username string   `json:"username"`
-	Groups   []string `json:"groups"`
+	ARN         string   `json:"arn"`
+	Username    string   `json:"username"`
+	ARNPatterns string   `json:"arnPatterns"`
```
jaypipes, line 40: AFAICT, this doesn't belong here. The `IAMIdentityMappingSpec` refers to the relationships between a specific Kubernetes user and a specific IAM principal.

pkg/mapper/file/mapper.go
```diff
@@ -83,6 +93,22 @@ func (m *FileMapper) Map(canonicalARN string) (*config.IdentityMapping, error) {
 		}, nil
 	}
 
+	for _, arnLikeMapping := range m.arnLikeList {
+		//g := glob.MustCompile(arnLikeMapping.ARNLike)
```
jaypipes, line 97: please remove this commented-out code.

vendor/modules.txt
```diff
@@ -71,6 +71,15 @@ github.com/go-openapi/jsonpointer
 github.com/go-openapi/jsonreference
 # github.com/go-openapi/swag v0.19.14
 github.com/go-openapi/swag
+# github.com/gobwas/glob v0.2.3
+github.com/gobwas/glob
```
jaypipes, line 75: the last commit to this library was >3 years ago. Are you confident it is well-maintained?

jaypipes (changes requested):
@logandavies181 thank you very much for this excellent contribution! I like the idea (pun intended) but have a number of suggestions inline for you to ponder.

pkg/mapper/crd/apis/iamauthenticator/v1alpha1/types.go
logandavies181: I think you might be right. I might just roll back my changes to that feature - perhaps I'll leave that for my next PR if this makes it in :)

pkg/config/types.go
logandavies181: Ok I'll give that a go. I wonder if it would be better as a plural (i.e. `UserARNLikes []string`) but on the other hand it's probably a little tidier and I think it still meets one of my use-cases - where I have many similarly named IAM roles I'd like to map to a static series of groups

go.sum
```diff
@@ -145,6 +145,8 @@ github.com/go-openapi/swag v0.19.5/go.mod h1:POnQmlKehdgb5mhVOsnJFsivZCEZ/vjK9gh
 github.com/go-openapi/swag v0.19.14 h1:gm3vOOXfiuw5i9p5N9xJvfjvuofpyvLA9Wr6QfK5Fng=
 github.com/go-openapi/swag v0.19.14/go.mod h1:QYRuS/SOXUCsnplDa677K7+DxSOj6IPNl/eQntq43wQ=
 github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/melR3HDY=
+github.com/gobwas/glob v0.2.3 h1:A4xDbljILXROh+kObIiy5kIaPYD8e96x1tgBhUI5J+Y=
+github.com/gobwas/glob v0.2.3/go.mod h1:d3Ez4x06l9bZtSvzIay5+Yzi0fmZzPgnTbPcKjJAkT8=
```
jaypipes, line 149: Maybe run `go mod tidy` to remove this leftover entry?

pkg/mapper/configmap/configmap.go
```diff
@@ -25,9 +26,11 @@ import (
...
 
 type MapStore struct {
-	mutex sync.RWMutex
-	users map[string]config.UserMapping
-	roles map[string]config.RoleMapping
+	mutex        sync.RWMutex
+	users        map[string]config.UserMapping
+	userArnLikes map[string]config.UserMapping
+	roles        map[string]config.RoleMapping
+	roleArnLikes map[string]config.RoleMapping
```
jaypipes, line 33: I'm not following why you'd need to have separate `userArnLikes` and `roleArnLikes` fields here. Now that you've placed the `UserArnLike` and `RoleArnLike` fields within the `UserMapping` and `RoleMapping` structs respectively I'm unsure why you would need a separate collection of those structs here in `MapStore`?

pkg/mapper/configmap/mapper.go
```diff
@@ -52,6 +52,24 @@ func (m *ConfigMapMapper) Map(canonicalARN string) (*config.IdentityMapping, err
...
+
+	ualm, err := m.UserArnLikeMapping(canonicalARN)
+	if err == nil {
+		return &config.IdentityMapping{
+			IdentityARN: canonicalARN,
+			Username:    ualm.Username,
+			Groups:      ualm.Groups,
+		}, nil
+	}
+
```
jaypipes, line 72: This should not be necessary. the `m.RoleMapping()` and `m.UserMapping()` methods should simply check whether the `canonicalARN` variable matches any of the `RoleMapping` or `UserMapping` structs in the `mapStore`, regardless of whether the `RoleMapping` contains an ARN-specific string match or an ARNLike match.

pkg/mapper/configmap/configmap.go
```diff
@@ -111,31 +116,77 @@ func (err ErrParsingMap) Error() string {
...
+					if err != nil {
+						errs = append(errs, err)
+					} else if !ok {
+						errs = append(errs, fmt.Errorf("RoleARNLike '%s' did not match an ARN for an IAM Role", roleMapping.RoleARNLike))
+					}
+					roleArnLikeMappings = append(roleArnLikeMappings, roleMapping)
+				default:
+					errs = append(errs, fmt.Errorf("Unexpected error parsing roleMapping: %v", roleMapping))
+				}
+			}
```
jaypipes, line 189: I'm confused why you need separate maps for `UserMapping` and `RoleMapping` for ArnLike or non-ArnLike matchers. You could probably simplify things quite a bit by creating to helper methods for `UserMapping` and `RoleMapping` called `Validate` and `Matches`:

```go
// Validate returns whether the RoleMapping is valid after being
// unmarshaled from JSON configuration
func (m *RoleMapping) Validate() error {
	if m ==  nil {
		return fmt.Errorf("RoleMapping was nil")
	}
	if m.RoleARN == "" && m.RoleARNLike == "" {
		return fmt.Errorf("One of rolearn or rolearnLike must be supplied"))
	} else if m.RoleARN != "" && m.RoleARNLike != "":
		return fmt.Errorf("Only one of rolearn or rolearnLike can be supplied"))
	}
	if m.RoleARNLike != "" {
		ok, err := arnlike.ArnLike(m.RoleARNLike, "arn:*:iam:*:*:role/*")
		if err != nil {
			return err
		} else if !ok {
			return fmt.Errorf("RoleARNLike '%s' did not match an ARN for an IAM Role", m.RoleARNLike)
		}
	}
	return nil
}

// Matches returns true if the supplied ARN or ARN-like thing matches
// this RoleMapping
func (m *RoleMapping) Matches(subject string) bool {
	if m.RoleARN != "" {
		return m.RoleARN == subject
	}
	ok, _ := arnlike.ArnLike(subject, m.RoleARNLike)
	return ok
}
```

jaypipes (changes requested):
Hi @logandavies181 thanks for your latest push! Getting closer but I still think we can improve the simplicity and readability of the solution. Please check my inline comments.

pkg/config/mapper.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return nil
+}
+
+// Matches returns true if the supplied ARN or ARN-like string matches
+// this RoleMapping
+func (m *RoleMapping) Matches(subject string) bool {
+	if m.RoleARN != "" {
+		return strings.ToLower(m.RoleARN) == strings.ToLower(subject)
```
logandavies181, line 38: @jaypipes I'm not 100% sure why previously RoleARNs have been lowercased before being put in maps, but anyway I've moved all the lowercasing to this comparison so the unit tests still pass

cmd/aws-iam-authenticator/add.go
```diff
@@ -73,18 +85,30 @@ var addRoleCmd = &cobra.Command{
...
-		checkPrompt(fmt.Sprintf("add rolearn %s, username %s, groups %s", roleARN, userName, groups))
+		var arnOrArnLike string
+		switch {
+		case roleARN != "" && roleARNLike != "":
+			fmt.Printf("only one of --rolearn or --rolearnlike can be supplied")
+			os.Exit(1)
+		case roleARN != "":
+			arnOrArnLike = "userarn"
+		case roleARNLike != "":
+			arnOrArnLike = "userarnlike"
```
jaypipes, line 101: these should be `rolearn` and `rolearnlike` respectively, right?

go.sum
```diff
@@ -278,6 +278,8 @@ github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/kr/text v0.2.0 h1:5Nx0Ya0ZqY2ygV366QzturHI13Jq95ApcVaJBhpS+AY=
 github.com/kr/text v0.2.0/go.mod h1:eLer722TekiGuMkidMxC/pM04lWEeraHUUmBw8l2grE=
+github.com/logandavies181/arnlike v1.1.0 h1:QkXQjpE4KMxJsdMubv48HFdUAx/dW3AHUbtnlWuLaoI=
+github.com/logandavies181/arnlike v1.1.0/go.mod h1:JQDNVkP+wbOAOIAcG02N6X/QpRU6hg/yAJUKc/xnOhg=
```
jaypipes, line 282: :+1:

pkg/config/mapper.go
jaypipes: Yup, no worries. ++

pkg/mapper/configmap/client/client.go
```diff
@@ -78,18 +78,27 @@ func (cli *client) add(role *config.RoleMapping, user *config.UserMapping) (cm *
...
 		if user != nil {
+			err = user.Validate()
+			if err != nil {
+				return fmt.Errorf("user is invalid: %v", err)
+			}
 			for _, r := range userMappings {
-				if r.UserARN == user.UserARN {
-					return fmt.Errorf("cannot add duplicate user ARN %q", user.UserARN)
+				if r.Key() == user.Key() {
+					return fmt.Errorf("cannot add duplicate user ARN %q", user.Key())
```
jaypipes, line 101: Nice. Much simpler and cleaner. ++

jaypipes (commented):
@logandavies181 you did a great job on this, thank you for hanging in there through the revisions.

There's one remaining typo/copypasta to address and then I'd like to discuss this patch with provider-aws at the Friday SIG meeting.

Note that we (AWS) need to do an internal review for security purposes of this new functionality. While I like the feature very much (pun intended), I do think that it would be wise to gate this feature behind an opt-in boolean feature flag in the webhook configuration so that folks who want to use the feature can enable it by setting something like an `enableARNLikeMatch` option in their config file. I think doing that, you will find less resistance to the feature and allow a security review to proceed apace.

/cc @nckturner @jqmichael @wongma7 

Best,
-jay

cmd/aws-iam-authenticator/add.go
logandavies181: Nice catch!

logandavies181:
I've added some logic to use the featureGate library that was there already. I'm not sure the other featureGate is actually being used, and it felt a bit awkward to use with `viper` so hopefully you're happy with my implementation.

Anyway I'm, looking forward to being able to use this feature in EKS :) so I'd love to know how the SIG meeting and security review go

christophetd:
>  it would be wise to gate this feature behind an opt-in boolean feature flag in the webhook configuration so that folks who want to use the feature can enable it by setting something like an enableARNLikeMatch option in their config file

How would that work when using EKS? Honestly, the lack of support for AWS SSO is pretty annoying in there (c.f. #333)

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/aws/aws-sdk-go v1.38.49
 	github.com/gofrs/flock v0.7.0
+	github.com/logandavies181/arnlike v1.1.0
```
christophetd, line 8: This library is used for quite a critical check, I would suggest re-implementing in `aws-iam-authenticator` rather than depending on a library that's new and not used anywhere else

jaypipes:
> Hi @jaypipes. Are you happy with the featureGate implementation. Are we just waiting on the security review?

@logandavies181 yes, we are waiting on a security review. Please bear with us. Thanks for your patience!

jaypipes:
> > Hi @jaypipes. Are you happy with the featureGate implementation. Are we just waiting on the security review?
> 
> @logandavies181 yes, we are waiting on a security review. Please bear with us. Thanks for your patience!

@logandavies181 we have a security review scheduled for Friday at 3:30pm EST. Will discuss this then and hopefully Monday have a decision. Thanks again :)

go.mod
ktham: I agree, this library is more security-sensitive, and the amount of code is not that much either https://github.com/logandavies181/arnlike/blob/main/arnlike.go, I think it makes more sense to move that code here, and have the code audited and reviewed within this repo

Logankwb: If one of the maintainers requests it, I'll be happy to oblige. However, all the dependencies for this project are vendored, and the whole source of that module is available in this PR, so I'm sure the security folks will be reviewing it

blakepettersson:
What was the outcome of the security review?

jaypipes (changes requested):
@logandavies181 apologies for the long delay in response.

We had an internal review of this PR last Friday and there were a number of concerns raised. I will try to outline them here, but if you'd like to speak in person about the requests, I'm happy to hop on a zoom/chime call with you.

OK, so the group reached a conclusion that this PR as it currently stands would be too much of a security risk due to the ease with which users could enable the feature and configure an ARNLike match that would easily match any IAM user or role using star/glob wildcards for either the account ID part of an ARN or the name part of the ARN.

For instance, an admin might configure the `roleArnLike` field as `arn:*:iam:*:*:role/*` which would match any IAM role in any account and region...

Because of this concern, we debated how to ask you to change the PR to limit the potential fo easy abuse of the wildcard matching. One person suggested maybe requiring an account ID component to the ARNLike strings and/or disallowing an empty `*` without any prefix. But other folks remarked that such constraints would inevitably be hard to come up with and end up being fairly arbitrarily decided -- for example, if we constrained the name part of the ARN to not be an empty `*` and forced some prefix, how long of a prefix would we enforce?

So, instead of trying to somewhat arbitrarily decide on such a prefix constraint, what we eventually decided was that it would be better to rework this feature to *specifically solve the SSO auto-generated Role ARN name use case and pain point*. Yes, this is somewhat antithetical to good feature design, in that we're recommending a non-generic, inflexible solution instead of the extensible solution you've come up with here. But, for security reasons, we think it would be best to have a myopic view for this feature.

We think the feature should be enabled as configuration block within RoleMapping that is specific to SSO.

Something like the following:

```go
// SSOARNMatcher contains fields used to match Role ARNs that
// are generated for AWS SSO sessions. These SSO Role ARNs
// follow this pattern:
//
// arn:aws:sts::<ACCOUNT_ID>:assumed-role/AWSReservedSSO_<SSO_PermissionSetName>_<RANDOM_STRING>/<USER_NAME>
//
// This struct enables iam-authenticator to match SSO generated Role ARNs with
// those random string suffixes.
type SSOARNMatcher struct {
    // PermissionSetName is the name of the SSO Permission Set that will be found
    // after the "AWSReservedSSO_" string in the Role ARN.
    // See: https://docs.aws.amazon.com/singlesignon/latest/userguide/permissionsets.html
    PermissionSetName string `json:"permissionSetName"`
    // AccountID is the AWS Account ID to match in the Role ARN
    AccountID string `json:"accountID"`
    // Username is the AWS SSO username to match in the Role ARN
    Username string `json:"username"`
}
...
type RoleMapping struct {
...
    // SSO contains fields used to match Role ARNs that
    // are generated for AWS SSO sessions.
   SSO *SSOARNMatcher `json:"sso,omitempty"`
}
```

And then the matching code that you have put in this PR would need to match specific fields in the SSO session Role ARN as configured in the `SSOARNMatcher` structs...

We realize this is a very big change we're requesting and really appreciate you sticking this out with us and being patient. Again, if you would like to chat about the request in person, I'm happy to hop on a call with you. You can find me as jaypipes on Kubernetes Slack community.

Best,
-jay

logandavies181:
Hi @jaypipes thanks for the response. I'm glad you're still keen to solve the SSO pain point, but I can't say I fully agree with the stance.

I understand the code as it is right now gives the ability for an admin to open permissions far more than would be sensible, but to do so they would have to make a very deliberate action to map `arn:*:iam::*:role/*` to privileged k8s users/groups.

I agree limiting this to not allowing a wildcard account or partition would be a good thing, but consider this example for a use-case for using `role/*`:
- Product teams use some federated identity provider (not using AWS SSO) to log into the account hosting the EKS cluster (I believe the IdP can control the {{SessionName}} in this case(?))
- K8s admin team maps `arn:aws:iam::012345678910:role/*` to `system:basic-user` group and username: `product-team:{{SessionName}}`
- K8s admin team creates namespaces for product teams and uses RoleBindings in their namespaces to grant permissions to individuals based on their k8s username

Anyway, the current way to use SSO roles is to map the canonicalized rolearn, which looks like: `arn:aws:iam::012345678910:role/AWSReservedSSO_<SSO_PermissionSetName>_<RANDOM_STRING>` (without the `path` part that you see in the actual full ARN in the IAM console). This doesn't give any information into the SSO <USER_NAME>, although I believe users could still use that value via {{SessionName}}; so I'd be keen to omit that feature.

In the end the comparison for SSO roles would look almost identical to how I've implemented it for the general case, so it'd be a shame to add extra code to restrict the functionality to SSO only.

I do quite like the approach you have above for this, so would you consider this middle ground?
```
// names hastily chosen

type RoleARNMatcher struct {
    RolePattern string `json:"rolePattern"`
    // AccountID is the AWS Account ID to match in the Role ARN
    AccountID string `json:"accountID"`
    // Partition is the AWS Partition
    Partition string `json:"partition" // defaulting to "aws" if left blank
}
...
type RoleMapping struct {
...
   RoleMatcher *RoleARNMatcher `json:"roleMatcher,omitempty"`
}
```
This way we enforce the use of accountId in an obvious way, and but still gives admins a powerful and flexible way to manage roles, should they choose to use it.

I'll reach out to discuss further as SSO makes up the majority of the pain in my use-case, but thought there might still be interest from the community in solving the general case

Many thanks!
Logan

johnkeates:
Just as a 'we use this too': besides local AWS SSO ARNs we also have non-AWS-SSO Idp roles that would then still not work. We were working on an internal tool that would triple-jump roles as to have the 'correct' ARN in the end, but it is really messy to have to work around limitations like that. (You either IdP or AWS SSO authenticate, then assume a role that the Idp or AWS SSO allows you to, and then in the same account that you are already in again assume a local IAM role that matches the ARN in the ConfigMap)

While the risk of exposing more ARNs than you should exists, this also exists when there are extra IAM-hoops created to jump through to mitigate the wildcard problems in the first place. If we have to make a local role that itself has a wildcard assume policy then we essentially just moved the problem around instead of solving it. Same for trying to make the IdP spit out the correct ARN, because now the IdP RBAC has to have wildcards to allow uses to select the correct role...

logandavies181:
Hey @jaypipes, I made the change to be SSO-only over the weekend. I still think there's value in the generic case and I don't think this solves https://github.com/aws/containers-roadmap/issues/474 anymore; but I suppose with the interface how it is currently, it wouldn't be too hard to extend it later

pkg/config/mapper.go
```diff
@@ -0,0 +1,98 @@
...
+	if m.SSO == nil {
+		return ""
+	}
+
+	var partition string
+	if m.SSO.Partition == "" {
+		partition = "aws"
+	}
+
+	return strings.ToLower(fmt.Sprintf("arn:%s:iam::%s:role/AWSReservedSSO_%s_*", partition, m.SSO.AccountID, m.SSO.PermissionSetName))
```
jaypipes, line 22: what would happen if a k8s admin set `sso.accountID` or `sso.permissionSetName` to `*`?

pkg/config/mapper.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if m.RoleARN == "" && m.SSO == nil {
+		return fmt.Errorf("One of rolearn or SSO must be supplied")
+	} else if m.RoleARN != "" && m.SSO != nil {
+		return fmt.Errorf("Only one of rolearn or SSO can be supplied")
+	}
+
+	if m.SSO != nil {
+		ssoArnLikeString := m.SSOArnLike()
+		ok, err := arnlike.ArnLike(ssoArnLikeString, "arn:*:iam:*:*:role/*")
```
jaypipes, line 39: To continue my comment from above, I *think* that if a k8s admin set `sso.accountID` or `sso.permissionSetName` to `*`, this `arnlike.ArnLike()` check would return `true, nil`, right? Wouldn't that be a way around the whole "make sure an account ID and a permission set name is specified" thing?

jaypipes (changes requested):
Apologies for another long delay in reviewing, @logandavies181.

This updated PR I think still has a security hole in it -- please correct me if I'm incorrect, though! See inline for an explanation why.

Best,
-jay

pkg/config/mapper.go
logandavies181: This is more a sanity check that the settings look like a role when parsed. I've added some checks for those settings in Validate() now

logandavies181:
@jaypipes good catch. Ready to review again!

go.mod
jaypipes: @logandavies181 actually one of the comments from the security reviewers was to move this small library into the main Go package instead of importing as a library. Could you do that please? Sorry I missed this on earlier reviews..

jaypipes (changes requested):
@logandavies181 thanks for the updates! Go ahead and copy your arnlike.go file into the main package instead of importing it as a library (this was a suggestion from one of the other security reviewers I had forgotten to ask you). And there's a couple minor nits inline, otherwise I'm good with this approach and will ask for a quick followup review with the same security reviewer group.

Thank you for your patience!
-jay

logandavies181:
@jaypipes good to go again!

pkg/arn/arnlike.go
```diff
@@ -0,0 +1,131 @@
...
+// special reports whether byte b needs to be escaped by QuoteMeta.
+func special(b byte) bool {
+	return b < utf8.RuneSelf && specialBytes[b%16]&(1<<(b/16)) != 0
+}
+
+func init() {
+	for _, b := range []byte(`\.+()|[]{}^$`) {
+		specialBytes[b%16] |= 1 << (b / 16)
+	}
+}
```
jaypipes, line 96: Would an easier-to-read (and easier to vet for security purposes) approach be to do something like this instead of the bitmap approach?

```go
var specialChars []byte(`\.+()|[]{}^$`)

func containsSpecial(subject string) bool {
	for _, sc := range specialChars {
		if strings.ContainsRune(subject, sc) {
			return true
		}
	}
	return false
}
```

I mean, this isn't in the performance-sensitive code path, right? This is just used when preparing the matching regexes...

logandavies181: That implementation was copied from regexp.QuoteMeta. While it's not performance critical, I do feel like it's going to be called a lot of times.

Anyway I've simplified the implementation considerably

logandavies181:
@jaypipes ready again!

jaypipes (approved):
I'm happy with this now, @logandavies181, thank you for persevering!

@micahhausler @nckturner @jyotimahapatra can I ask all three of you to please review this latest revision in the coming week and sign off on this new, more restricted approach?

Thanks much!
-jay

logandavies181:
Hi @micahhausler @nckturner @jyotimahapatra, could you folks please review? I'm really keen to get this merged

logandavies181:
@jaypipes do you reckon you could get your colleagues @micahhausler @nckturner @jyotimahapatra to review please? We're so close to getting this in

jyotimahapatra:
Apologies I've been oncall and couldn't get to it yet. The soonest i can get to this is Friday. :(

logandavies181:
@jaypipes @micahhausler @nckturner @jyotimahapatra Hey again, sorry to pester but I'm really keen to get this in. My team at work manages a large number of clusters and we make heavy use of SSO roles so this would be really helpful for us.

And there are many other users in the community that have expressed interest in this exact feature, so I'd really appreciate your help in getting this over the line

johnkeates:
@logandavies181 we are also eagerly awaiting this feature ourselves; even our "read only" users really want this just for the AWS EKS console to work for their SSO users without double-assuming roles. And that's just the tip of the iceberg (user-wise).

pkg/mapper/configmap/configmap_test.go
```diff
@@ -170,7 +196,7 @@ func TestLoadConfigMap(t *testing.T) {
 
 	watcher.Add(&core_v1.ConfigMap{ObjectMeta: meta, Data: data})
 
-	time.Sleep(2 * time.Millisecond)
+	time.Sleep(2 * time.Second)
```
jyotimahapatra, line 199: sleeping in a test could cause flakiness. I imagine the change was required because the tests failed otherwise. Probably in a different PR and time, we could remove dependency on sleep in test.

pkg/mapper/file/mapper.go
```diff
@@ -2,41 +2,60 @@ package file
...
+		}
+		var key string
+		if m.UserARN != "" {
+			canonicalizedARN, err := arn.Canonicalize(m.UserARN)
+			if err != nil {
+				return nil, fmt.Errorf("error canonicalizing ARN: %v", err)
+			}
+			key = canonicalizedARN
+		} else {
+			key = m.Key()
```
jyotimahapatra, line 56: i think i have a gap in my understanding about this part. The UserArn part should not be changing in my impression(since all changes are about adding sso in rolemap). Can you tell more about why this segment of change is neded when UserArn is nil.

jyotimahapatra (commented):
The change looks good. I've a few questions mostly for my understanding. Also, can we add to the readme about a usage example. Also, since we're doing a bit of if else around if sso/rolearn is present in a role mapping, i was  thinking whether keeping a separate mapping separately would be worth it. Let me know what you think about it. Today we have `mapRoles`, `mapUsers` and `mapAccounts`. We can add a separate segment for `mapSso` and the change will be compartmentalized.

pkg/mapper/configmap/configmap_test.go
nckturner: This test has been flaky.  Let's just merge this change in a separate PR first as well, and we can rebase this one on it.

pkg/config/mapper.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if m.SSO != nil {
+		accountIDRegexp := regexp.MustCompile("^[0-9]{12}$")
+		if !accountIDRegexp.MatchString(m.SSO.AccountID) {
+			return fmt.Errorf("AccountID '%s' is not a valid AWS Account ID", m.SSO.AccountID)
+		}
+
+		// https://docs.aws.amazon.com/singlesignon/latest/APIReference/API_PermissionSet.html
+		permissionSetNameRegexp := regexp.MustCompile(`^[\w+=,.@-]{1,32}$`)
```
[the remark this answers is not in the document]
logandavies181: Ah yes, very tricky indeed. We aren't missing that, the `{1,32}` means "between 1 and 32 occurrences of the previous character set"

pkg/config/mapper.go
```diff
@@ -0,0 +1,119 @@
...
+	if m.RoleARN != "" {
+		return strings.ToLower(m.RoleARN) == strings.ToLower(subject)
+	}
+
+	// Assume the caller has called Validate(), which parses m.RoleARNLike
+	// If subject is not parsable, then it cannot be a valid ARN anyway so
+	// we can ignore the error here
+	var ok bool
+	if SSORoleMatchEnabled {
+		ok, _ = arn.ArnLike(subject, m.SSOArnLike())
```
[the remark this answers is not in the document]
logandavies181: So in the comments above, there's an assumption that the caller has first called `Validate`, which will check if the SSO arn is parsable. If there's an error returned here, it means that the canonicalized ARN of the authenticating user is not valid even after the sts:GetCallerIdentity call.

I can add an error log if that occurs

pkg/mapper/configmap/configmap_test.go
logandavies181: Yeah no idea what's up with this but the test was much less flaky after increasing the time. I don't have any plans to fix it but agree it needs fixing

pkg/mapper/file/mapper.go
logandavies181: In a previous commit, I had a top level ARNLikeMapping struct which had UserARNLike and RoleARNLike attributes so I had changes to both. @jaypipes suggested adding the changes in this file to both of those which made both much cleaner.
There's a fair bit of copy/paste code between the two so I though it would be best to keep the implementations similar so they can both benefit from the refactor as part of this PR

logandavies181:
Hi @jyotimahapatra, thanks for your feedback. I will make the following changes later today:
- update the README with the new feature
- check that error and log it at error level if not nil

Could I please ask you to check with @jaypipes as I initially had this feature up a level, with mapUsers and mapRoles but he asked for it to  be under mapRoles. I won't make any changes to the implementation until I have agreement between you

nckturner:
Can you:

1. Create a new markdown file in `docs/` that goes over the desired UX of this change in detail, and update the PR description with similar intended behavior, usage, CLI and configmap mapping changes.
2. Describe your test plan?  I see new unit tests in arnlike_test.go, config/mapper_test.go, file/mapper_test.go, and configmap_test.go, though I haven't looked at each case.  Are there any areas that you know of, that this PR doesn't cover, that we should discuss?  What about integration tests or e2e tests?  I am aware we don't have an e2e test framework yet, but this is something that is needed soon so if there is a good use case here we can prioritize adding it in the near future.

logandavies181:
> @logandavies181 can you:
> 
> 1. Create a new markdown file in `docs/` that goes over the desired UX of this change in detail, and update the PR description with similar intended behavior, usage, CLI and configmap mapping changes.
> 2. Describe your test plan?  I see new unit tests in arnlike_test.go, config/mapper_test.go, file/mapper_test.go, and configmap_test.go, though I haven't looked at each case.  Are there any areas that you know of, that this PR doesn't cover, that we should discuss?  What about integration tests or e2e tests?  I am aware we don't have an e2e test framework yet, but this is something that is needed soon so if there is a good use case here we can prioritize adding it in the near future.

Hi @nckturner, thanks for your feedback:
1. Will do, once I have confirmation from @jaypipes and @jyotimahapatra about the implementation
2. I've added unit tests for all the new code and some existing code that I modified. The PR integration runs the unit tests and integration tests and those both pass. I've done my best to cover positive and negative cases for all the new code. I've run it end-to-end locally with an ubuntu based k8s install and real AWS SSO that I configured on my personal account.
It would be nice to include AWS SSO in your E2E testing when you get around to adding it

nckturner:
> Will do, once I have confirmation from @jaypipes and @jyotimahapatra about the implementation

Understood that its more effort, but its important to detail the desired UX to assist reviewers when looking at the code.  Please at least update the PR description before waiting for sign off.

pkg/mapper/file/mapper.go
jyotimahapatra: Thanks for explaining. I think i have a gap in understanding. Can you tell more about how sso relates to roles and users with examples.

nckturner:
/hold for rebase.

After we have agreement on implementation, we can rebase into fewer, commits scoped into understandable chunks.

logandavies181:
> > Will do, once I have confirmation from @jaypipes and @jyotimahapatra about the implementation
> 
> Understood that its more effort, but its important to detail the desired UX to assist reviewers when looking at the code. Please at least update the PR description before waiting for sign off.

I've made a quick update to the PR description now, I'll add more detail later when I work on the readme and docs etc. Currently 9am Friday my time so my day job is getting in the way :)

go.mod
logandavies181: @nckturner this has been done on Jay's request

pkg/mapper/file/mapper.go
logandavies181: There's a bit more context in Jay's comment https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/416#pullrequestreview-907912125.

AWS SSO allows federated users to assume roles that it manages. It creates role arns that are long and ugly like: arn:aws:iam::000000000000:role/aws-reserved/sso.amazonaws.com/us-east-1/AWSReservedSSO_MyPermissionSet_1234567890abcde. However the current UX is you have:
```
  mapRoles: |
    - rolearn: arn:aws:iam::000000000000:role/AWSReservedSSO_MyPermissionSet_1234567890abcde
      username: {{SessionName}}
      groups:
        - some-k8s-group
```
Until that suffix changes, which I guess can occur when the SSO permission set changes. If your K8s admins are using SSO roles, this is very bad because another team changing stuff can lock everyone out of k8s.

Also it's confusing because the IAM Role path disappears when it's canonicalized, and there's a few github issues raised due to this.

With the PR as it is currently, the new UX is you create a mapping like this:
```
  mapRoles: |
  - sso:
      permissionSetName: MyPermissionSet
      accountId: "000000000000"
```
and you never need to think about that pesky random suffix or the IAM role path ever again

mapUsers is technically not relevant to this change but I changed it in earlier commits and I think the refactor is valuable. Also I'm sure those having the same method signatures will be useful for generically doing CRD mappings, although I haven't looked into that too much

jyotimahapatra: Thank you. I have 2 buckets of question to brainstorm.
-  `mapUsers is technically not relevant to this change but I changed it in earlier commits and I think the refactor is valuable` . I agree, Im aligned on most of the refactorngs. The one im having question about is [this](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/2f55274303e9d197f6bec4431b6a185943930c35/pkg/mapper/file/mapper.go#L49-L57)
```
		if m.UserARN != "" {
			canonicalizedARN, err := arn.Canonicalize(m.UserARN)
			if err != nil {
				return nil, fmt.Errorf("error canonicalizing ARN: %v", err)
			}
			key = canonicalizedARN
		} else {
			key = m.Key()
		}
```
Under what circumstance is the `m.UserARN` nil?

- Does the sso have to belong in the `mapRoles`? From a user experience perspective could sso matching be a separate stanza altogether? What is your opinion?

logandavies181: m.UserARN is a string so it's never nil 😎. But yeah it looks like that block is a relic of the previous implementation so that check is unnecessary and I'll clean it up.

SSO does not have to belong under mapRoles and it didn't initially. @jaypipes requested it be nested under mapRoles and I've requested that you check with him to agree on where you'd like it to sit.
My previous implementation had it as its own stanza so I'd like you folks to agree on it so we can avoid rubber banding the implementation between the two

logandavies181: @jyotimahapatra - the comment in question from Jay: https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/416#pullrequestreview-907912125

logandavies181: Also, just to share my opinion on this:

I like the idea of having the SSO mapping nested under mapRoles because:
- It's less code to add features to it than add more top-level mappings - particularly in the configmap package and the client CLI commands and tests
- We could make the if/else blocks nicer by having a method for saying what kind of mapping is at play:
```
switch mappingType := rolemapping.Type(); mappingType {
case config.RoleARN:
        // do rolearn specific stuff
case config.SSORole:
        // do sso specific stuff
}
```
- Lastly, it's less rework for me :)

jyotimahapatra: @jaypipes which of teh following do you prefer
```
apiVersion: v1
data:
  mapRoles: |
    - rolearn: arn:aws:iam::111122223333:role/my-console-viewer-role
      username: my-console-viewer-role
      groups:
        - eks-console-dashboard-full-access-group
   mapSSO: |
       - permissionSetName: ...
       - accountId: ...
       - userName: ...
       - groups:
         - eks-console-dashboard-restricted-access-group
```
vs
```
apiVersion: v1
data:
  mapRoles: |
    - rolearn: arn:aws:iam::111122223333:role/my-console-viewer-role
      username: my-console-viewer-role
      groups:
        - eks-console-dashboard-full-access-group
    - sso:
      - permissionSetName: ...
      - accountId: ...
      - userName: ...
      - groups:
        - eks-console-dashboard-restricted-access-group
```

Logankwb: option 2 - my reasoning above 👍

jyotimahapatra: Agree from code perspective we can handle the if elses..what do you think about potential errors when adding a new entry. Say a rolemap exists for unique sso users. In the current approach, does adding an arnlike at the top of the list swallow existing rules? See my example in the previous comment.

logandavies181: So technically there's no documentation on the expected behaviour for two matching rolearns. It's undefined behaviour, however the implementation in master will use the last item that matches. E.g.
```
  mapRoles: |
  - rolearn: arn:aws:iam::111122223333:role/my-console-viewer-role
    username: foo
    groups: 
    - foo-group
  - rolearn: arn:aws:iam::111122223333:role/my-console-viewer-role
    username: bar
    groups:
    - bar-group
```
Always maps `arn:aws:iam::111122223333:role/my-console-viewer-role` as username `bar` with groups `bar-group`

The behaviour at the time of writing for this PR is that `sso` blocks and `rolearn` blocks can collide an override each other in the same way. So latest defined item in the array with take precedence. E.g.
```
  - sso:
      permissionSetName: "ViewOnlyAccess"
      accountID: "111122223333"
    username: foo
    groups:
    - foo-group
  - rolearn: arn:aws:iam::111122223333:role/AWSReservedSSO_ViewOnlyAccess_123456789abcde
    username: bar
    groups:
    - bar-group
```
Always maps to username `bar` with groups `bar-group`

My main point here is that this is undefined behaviour, but at least this implementation preserves that

pkg/mapper/configmap/configmap_test.go
logandavies181: I'd prefer to leave it in this PR so we end up with less git operations.
I'll give it its own commit when I rebase as a compromise

pkg/mapper/file/mapper.go
logandavies181: @jyotimahapatra having another look at your comment, trying to match the SSO role ARN using the existing rolearn field and using the new SSO matcher is a misconfiguration as the user would be specifying potentially two different k8s users and groups  that they would want a single role mapped against.
Also, there is no "more narrow match" as each rolemapping is inspected and either passes or fails. The first mapping that  succeeds is the one that is applied, and they are tested by doing a range loop over the map that contains the rolemappings, which has an undefined order

logandavies181:
@nckturner @jaypipes @jyotimahapatra thanks again for your feedback. I'd like to keep the momentum, so please do get back to me as soon as you can about whether the SSO match should sit aside mapRoles or beneath it.

In the meantime I've added a markdown file describing this new feature and further updated the readme with the latest details.

I have not touched the readme as I recall Jay mentioned having this behind a feature-flag (I assumed to start with alpha) https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/416#pullrequestreview-883330621

pkg/mapper/file/mapper.go
jaypipes: @jyotimahapatra I prefer nesting the `sso` block within the `mapRoles` blocks, because SSO configuration only relates to roles, never users.

jaypipes (approved):
I'm still :+1: on this. I think @logandavies181 sufficiently answered the questions that reviewers had. @jyotimahapatra may I ask for one last review from you please? :)

jyotimahapatra:
> because SSO configuration only relates to roles, never users

Ok. I was thinking about a separate block apart from both roles and users. There is indeed the perspective that sso is still a role and should be nested, and i dont have a lot to argue against that. Apart from that, im good with teh changes.

jyotimahapatra (approved):

logandavies181:
Thanks @jyotimahapatra and @jaypipes. I will rebase into fewer commits later today as per @nckturner's request

logandavies181:
@nckturner I have now rebased off of master and updated my commits with incremental changes from master to how my branch was before the rebase. I'd be stoked to get your review on this.
Also @wongma7 I believe @jaypipes also requested your review so I'd appreciate your help in this also

The pre-rebase branch with original history can be found here:
https://github.com/logandavies181/aws-iam-authenticator/tree/pre-rebase

logandavies181:
@wongma7 @nckturner  bump

nckturner:
@logandavies181 sorry for the delay.  @wongma7 is on vacation and I'm finishing up an oncall shift, so time has been pretty scarce.  Planning to get to this early next week.

logandavies181:
> @wongma7 @nckturner bump

bump

sumanthkumarc:
+1 for the feature

logandavies181:
@jaypipes any chance we can merge with the two approvals there are so far? The docs say that one reviewer and one approver is the normal amount, and I feel like there has been plenty of time to review this PR - it's coming up to six months old soon

docs/sso_role_matcher.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Feature state
+
+Alpha
+
+## Use case
+
+Easy and robust configuration for AWS SSO managed roles, which currently have two main issues:
+
+- Confusing configuration. To use an SSO role, a user needs to map the Role ARN of the SSO ROle, minus the path.
```
nckturner, line 13: Super small nit, I'd prefer a numbered list or just `First...`, `Second...` given the list is kind of split up.

pkg/config/mapper.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		// https://docs.aws.amazon.com/singlesignon/latest/APIReference/API_PermissionSet.html
+		permissionSetNameRegexp := regexp.MustCompile(`^[\w+=,.@-]{1,32}$`)
+		if !permissionSetNameRegexp.MatchString(m.SSO.PermissionSetName) {
+			return fmt.Errorf("PermissionSetName '%s' is not a valid AWS SSO PermissionSet Name", m.SSO.PermissionSetName)
+		}
+
+		switch m.SSO.Partition {
+		case "aws", "aws-cn", "aws-us-gov":
```
nckturner, line 53:
```suggestion
		case "aws", "aws-cn", "aws-us-gov", "aws-iso", "aws-iso-b":
```

pkg/arn/arnlike_test.go
```diff
@@ -0,0 +1,154 @@
...
+			t.Errorf("Expected no error for input arn: %s pattern: %s", v.arn, v.pattern)
+		}
+
+		if ok {
+			t.Errorf("Expected false for input arn: %s pattern: %s", v.arn, v.pattern)
+		}
+	}
+}
+
+func TestArnLikeInvalidArns(t *testing.T) {
```
nckturner, line 92: Also test incomplete arnLikes: `arn:*` to ensure we are not allowing a glob to match the entire arn.

logandavies181: Good call

pkg/config/mapper.go
logandavies181: TIL

docs/sso_role_matcher.md
```diff
@@ -0,0 +1,53 @@
+# SSO Role Matcher
+
+Maps configuration for an AWS SSO managed IAM Role to a Kubernetes username and groups.
+
+## Feature state
+
+Alpha
```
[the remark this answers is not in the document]
logandavies181: Up to you folks as the maintainers to decide that. As a contributor, I'd prefer if there were some .md file detailing the criteria for promotion, so maybe a separate PR containing that would be best for deciding the criteria

logandavies181:
Thanks for the feedback @nckturner. Back to you

nckturner:
Let's merge this given its an alpha feature and get feedback on it. Sorry for the long review cycles and thanks for staying on top of it.

christophetd:
Awesome to have this merged! Is there any timeline for when this versino of aws-iam-authenticator will be used in EKS?

nckturner:
I think it makes sense for us to get this feature to beta and then launch it in EKS.  Might be able to do that in 1.24?

universam1:
Any workaround to have this available **now**, not yet in 1.24 @nckturner ? We need a way to access EKS using SSO from a root account.

tommy31:
Hey, any update on when we could use this on EKS ?

zelch:
Am I correct in reading that this was reverted by #509 ?

@nnmin-aws

Was there discussion about this somewhere?  And/or a plan going forward?

nnmin-aws:
thanks @zelch for your query. apology if any inconvenience.

#416 has never been picked up in any release. Then it was reverted by #509 in branch release-0.6.

It is still available in master branch. You can build image with master branch if you want to use this feature.

Thank you

zelch:
What are the release plans for this change at this point?

justinas-b:
Hey @nnmin-aws , could you elaborate more on why this has been reverted? Our organisation is also experiencing this issue and are looking forward for release of this PR.

OverStruck:
@justinas-b Looks to have been removed due #504 (reasons why are explain there)

johnkeates:
Well, that's a bummer. But I can see how even undocumented usage would still preferably not be broken by a new feature. It does seem to be in need of some pushing/tickling to get this back on track considering not much happened after the bug fix (unless it's waiting for that SSO e2e test mentioned in #506 ?).

logandavies181:
> Well, that's a bummer. But I can see how even undocumented usage would still preferably not be broken by a new feature. It does seem to be in need of some pushing/tickling to get this back on track considering not much happened after the bug fix (unless it's waiting for that SSO e2e test mentioned in #506 ?).

It looked like a bug, so I squashed it ¯\\\_(ツ)_/¯. There was nothing to indicate that it wasn't a bug or that it was expected.

Seeing as there has been this interest, I'll have a go at the E2E test - maybe they'll even review it

jku8:
Hi @logandavies181 I'm on the EKS team and we are looking into issues in regards to SSO. Would you be open to hopping on a call to discuss the topic?

johnkeates:
Looked like a bug to me as well 🤷 Oh well...

nnmin-aws:
Apology for late reply

The background:

(1) https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/416 was merged into master but it was not picked  in any official release. 

(2) in v0.6.0 https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/416 was picked but then issue https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/504 was found. Hence it was removed

(3) author contributed a fix #506. But there is a new SSO feature design is under review. The new design will replace #416. Hence #416 is not re-picked. 

We appreciate your interest in SSO feature and can organize some discussion with the new SSO design in kubernetes slack channel if you are interested. [jku8](https://github.com/jku8) can help arrange the discussion.

ktham:
>We appreciate your interest in SSO feature and can organize some discussion with the new SSO design in kubernetes slack channel if you are interested. @jku8 can help arrange the discussion.

nckturner:
Let me add some context -- the "new SSO feature" is an integration between Identity Center and EKS.  EKS uses the Authenticator, so there is some overlap.  However, #416 can exist in the Authenticator code base without being enabled in EKS, so we *could* choose to support it in the meantime.  I think the question is how many users want this feature who are not using EKS, because we will not be enabling it in EKS.

zelch:
We are currently running kubernetes in AWS, not using EKS, so this continues to be of extreme interest.

logandavies181:
> because we will not be enabling it in EKS.

#416 also refactors some duplicated code and improves uinit test coverage. Can you please just review #506? It does not look possible to sensibly test SSO E2E in CI as getting SSO credentials requires user interaction

dm3ch:
@nckturner Is there any link where I can track the progress of new IAM Identity Center SSO auth in EKS feature?

logandavies181:
> We are currently running kubernetes in AWS, not using EKS, so this continues to be of extreme interest.

@nckturner seeing as there is interest in this but it will not be enabling it in EKS, would there be any possibility of the original wildcard feature getting merged in?

Originally this PR allowed for full wildcard support, and there's a great deal of interest in the feature: https://github.com/aws/containers-roadmap/issues/474

It would be great to contribute to the openness of Kubernetes recognize the non-EKS users that are interested in this feature by allowing it to be merged (with a feature flag of course) without going down the fork route

cc. @zelch - would that be useful to non-EKS users such as yourself? I'd certainly like it, but my team is committed to EKS

zelch:
@logandavies181 Absolutely, it would be very useful.

@dm3ch My understanding is that they are working on an EKS specific direct integration between Identity Center and EKS,

dm3ch:
But is there any links to announces or issues that would allows to track AWS implementation?

PR #391, opened 2021-08-09
Add arm64 builds

antoinedeschenes:
Following #371, this adds arm64 builds to releases.

Another PR will add arm64 Docker builds.

PR #417, opened 2021-12-28
Build multi-arch images

glebiller:
Build new platform binaries & container image for the following platforms:
* darwin arm64
* linux arm64

Extended the work done in PR #391


nckturner:
Thanks, this is great.  
/lgtm

PR #422, opened 2022-01-18
Add kind-based development environment

nckturner:
- Scripts to create and tear down the environment.
- Uses a kind cluster and a stand-alone authenticator container.
- Sets up a kubeconfig to test client-server interactions end to end.

Example use: 
```
$ make start-dev ADMIN_ARN=arn:aws:iam::352684330888:role/dev-k8s-01 AUTHENTICATOR_IMAGE=aws-iam-authenticator:v0.5.3_02a86a549cee91b37baff12
d2528f185594fb98c_20220117T003254Z                                                                                                                                                           
make /home/ubuntu/go/src/sigs.k8s.io/aws-iam-authenticator/_output/bin/aws-iam-authenticator        
make[1]: Entering directory '/home/ubuntu/go/src/sigs.k8s.io/aws-iam-authenticator'                 
make[1]: '/home/ubuntu/go/src/sigs.k8s.io/aws-iam-authenticator/_output/bin/aws-iam-authenticator' is up to date.
make[1]: Leaving directory '/home/ubuntu/go/src/sigs.k8s.io/aws-iam-authenticator'                  
./hack/start-dev-env.sh                                                                             
Creating network authenticator-dev                                                                  
644eb11eaf9b459d7702b1463bf01b66988077fb7119aa1ee5fb0b6e3b109e5e                                                                                                                             
9abc964c717778714e3aa8208abb01c3d67eeb3c6a7ffe6e5689c62b1ff7d15c
Authenticator running at 172.30.0.10                                                                                                                                                         
Creating cluster "authenticator-dev-cluster" ...                                                                                                                                             
WARNING: Overriding docker network due to KIND_EXPERIMENTAL_DOCKER_NETWORK                                                                                                                   
WARNING: Here be dragons! This is not supported currently.    
 ✓ Ensuring node image (kindest/node:v1.21.1) 🖼                                                                                                                                              
 ✓ Preparing nodes 📦                                                                                                                                                                        
 ✓ Writing configuration 📜                                
 ✓ Starting control-plane 🕹️                              
 ✓ Installing CNI 🔌                                                                                                                                                                         
 ✓ Installing StorageClass 💾                                                                                                                                                                
Set kubectl context to "kind-authenticator-dev-cluster"                                                                                                                                      
You can now use your cluster with:                                                                                                                                                           
                                                                                                                   
kubectl cluster-info --context kind-authenticator-dev-cluster --kubeconfig /home/ubuntu/go/src/sigs.k8s.io/aws-iam-authenticator/_output/dev/client/kind-kubeconfig.yaml
                                                                     
Thanks for using kind! 😊                                   
.                                                                                                       
Test authenticator with:                                                               
kubectl --kubeconfig="/home/ubuntu/go/src/sigs.k8s.io/aws-iam-authenticator/_output/dev/client/kubeconfig.yaml" --context="test-authenticator"
.   

```

And the authenticator client and server are exercised with the following parameters:
```
$ kubectl --kubeconfig="/home/ubuntu/go/src/sigs.k8s.io/aws-iam-authenticator/_output/dev/client/kubeconfig.yaml" --context="test-authenticator" get pods                                                                                                                                                                                 
No resources found in default namespace.        
```


hack/lib/dev-env.sh
```diff
@@ -0,0 +1,177 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# Check that required binaries are installed
+command -v make >/dev/null 2>&1 || { echo >&2 "make is required but it's not installed.  Aborting."; exit 1; }
+command -v docker >/dev/null 2>&1 || { echo >&2 "docker is required but it's not installed.  Aborting."; exit 1; }
+command -v kind >/dev/null 2>&1 || { echo >&2 "kind is required but it's not installed.  Aborting."; exit 1; }
```
jaypipes, line 24: this is fine, but consider having a check that validates a particular version of these dependencies. We've found in ACK land that just relying on docker or kind (and not a specific version of those) leads to ambiguous dev environments and lots of headache

Makefile
```diff
@@ -111,3 +112,31 @@ codegen:
...
 clean:
 	rm -rf $(shell pwd)/_output
+	
+.PHONY: clean-dev
+clean-dev:
+	rm -rf $(shell pwd)/_output/dev
+
+# Use make start-dev when you want to create a kind cluster for
+# testing the authenticator.  You must pass in the admin ARN and
+# the image under test.
```
jaypipes, line 122: See note below. Consider defaulting the image to one built from the locally checked-out source.

hack/lib/dev-env.sh
```diff
@@ -0,0 +1,177 @@
...
+host_path_webhook_kubeconfig="${authenticator_export_host_dir}/apiserver-auth-webhook-kubeconfig.yaml"
+# Kubectl kubeconfig template
+kubectl_kubeconfig_template="${REPO_ROOT}/hack/dev/kubeconfig.yaml"
+client_dir="${OUTPUT}/dev/client"
+# Generated kubeconfig for testing
+kubectl_kubeconfig="${client_dir}/kubeconfig.yaml"
+# Admin kubeconfig generated by kind
+kind_kubeconfig="${client_dir}/kind-kubeconfig.yaml"
+
+function create_network() {
```
jaypipes, line 81: is there a requirement to create a separate virtual network? Why not use the default docker virtual network?

hack/lib/dev-env.sh
```diff
@@ -0,0 +1,182 @@
...
+        --publish ${AUTHENTICATOR_PORT}:${AUTHENTICATOR_PORT} \
+        --rm \
+        "${AUTHENTICATOR_IMAGE}" \
+        server \
+        --config "${authenticator_config_dest_dir}/authenticator.yaml"
+}
+
+function kill_authenticator() {
+    docker kill aws-iam-authenticator || true
+}
```
jaypipes, line 146: the issue I have with this is that it's not actually testing the installation/operating practice for `aws-iam-authenticator` that is described in the README: deploy the authenticator as a Daemonset on each control plane node in the k8s cluster, using [the example manifest](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/deploy/example.yaml). Why not just kubectl apply a slightly-modified version of the example YAML files against the KinD cluster? That would do dual service of validating the recommended installation approach for the authenticator.

jaypipes (changes requested):
@nckturner this is a great start. I have some ideas inline that might simplify things, however.

hack/lib/dev-env.sh
nckturner: Hm, that's a good point.  At least printing a warning with expected/tested versions.

hack/lib/dev-env.sh
nckturner: Kind already creates/uses its own network, so in order to include the authenticator in that network I needed to do it this way (or run authenticator on the kind node along with the apiserver, etc, but in that case I would have had to do some setup slightly differently, including generating all config/certs up front instead of letting authenticator generate them.  This is because the ca.cert needs to embedded in a webhook kubeconfig and passed to the apiserver on startup).  Also interacting with the apiserver as a standalone docker container is quite easy, and we don't have to interact with kind for debugging... e.g. port-forwarding in order to curl the metrics endpoint, etc.

hack/lib/dev-env.sh
nckturner: Yeah.  The reason I went with the above approach is because I decided to let the authenticator generate its own certificate and the webhook kubeconfig on startup.  Since the path to this config needs to be passed to the apiserver as a flag on startup, this suggests starting the authenticator before the kind cluster, because the API server is started automatically on cluster creation and is configured in the kind cluster configuration.  That being said, we can just use the [init](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/cmd/aws-iam-authenticator/init.go) command to pre-generate the configuration, and then launch authenticator, and then patch the API server.

hack/lib/dev-env.sh
```diff
@@ -0,0 +1,182 @@
...
+        -e "s|{{AUTHENTICATOR_KUBECONFIG}}|${authenticator_kubeconfig}|g" \
+        -e "s|{{AUTHENTICATOR_PORT}}|${AUTHENTICATOR_PORT}|g" \
+        -e "s|{{AUTHENTICATOR_IP}}|${AUTHENTICATOR_IP}|g" \
+        -e "s|{{CLUSTER_NAME}}|${CLUSTER_NAME}|g" \
+        "${authenticator_config_template}" > "${authenticator_config}"
+}
+
+function write_kind_config() {
+    mkdir -p "${kind_config_host_dir}"
+    sed -e "s|{{AUTHENTICATOR_EXPORT_HOST_DIR}}|${authenticator_export_host_dir}|g" \
```
jyotimahapatra, line 116: do these commands work on mac? i know sed has different syntax for mac and linux. I suggest to keep them close, we can mandate installation of `gsed`?

hack/lib/dev-env.sh
jyotimahapatra: We should try the approach that allows quicker iteration that does not need a tear down and restart of the kind cluster. In both approaches i think we get it, probably bit easier with daemonset approach. 
The second criteria for choosing is closeness to setups. I'm not sure which one is more prevalent, but i'd lean towards the approach implemented due to my familiarity with managed setups.

jaypipes:
Chatted about this offline. Good with improvements in followup PRs :)

/lgtm

hack/lib/dev-env.sh
nckturner: Yeah, I have no idea if they work on Mac but I've definitely ran into problems with sed commands that were written by mac users :)

PR #426, opened 2022-02-02
Embed go-runner into the image

jyotimahapatra:
go-runner is essential in the minimal distroless image to redirect stderr logs. For example panic logs not captured by klog land up in stderr. Without go-runner, we'd lose these logs.


Dockerfile
```diff
@@ -11,14 +11,17 @@
...
-ARG image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2021-08-26-1630012071
+ARG image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:latest
 
 FROM golang:1.16 AS builder
 WORKDIR /go/src/github.com/kubernetes-sigs/aws-iam-authenticator
 COPY . .
 RUN make bin
 RUN chown 65532 _output/bin/aws-iam-authenticator
 
+FROM public.ecr.aws/eks-distro/kubernetes/go-runner:v0.9.0-eks-1-21-4 as go-runner
```
nckturner, line 22: Let's just use go-runner as the base and have 1 less step?  Its already a distroless image ([see the entire Dockerfile here](https://github.com/kubernetes/release/blob/master/images/build/go-runner/Dockerfile)).  This is what we use for other CNCF-owned projects as well ([cloud-provider-aws](https://github.com/kubernetes/cloud-provider-aws/blob/master/Dockerfile#L25)).

Dockerfile
```diff
@@ -11,14 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-ARG image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:2021-08-26-1630012071
+ARG image=public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-nonroot:latest
```
nckturner, line 14: Update the tag but don't use latest, which results in less deterministic builds.

nckturner (changes requested):

Dockerfile
jyotimahapatra: im trying to be consistent about which registry we use. Since we're using the `public.ecr.aws/eks-distro-build-tooling` already , i preferred the go-runner from the same source.

nckturner: Ok I didn't even see that we were pulling from the eks-distro version of go-runner (which I can't find the dockerfile for, so I'm not sure how its built).  My suggestion was use go runner as the base instead of the minimal base to remove a step, but this works too.

nckturner:
Looks good to me, just update the minimal base tag to something other than latest.

Dockerfile
jyotimahapatra: updated latest tag. The eks-distro go-runner is just a build of upstream go-runner[ref](https://github.com/aws/eks-distro-build-tooling/blob/b729ea295aec9b2d271179b004ba2624cb950969/generate-attribution/tests/cases/kubernetes/release/_output/attribution/go-deps.json#L3287-L3297).
In case of cve findings, it will be much easier to update the eks-distro minimal base image, and hence keeping the dependency on everything eks-distro than using upstream go-runner. Of course we can revisit that later.

nckturner (approved):

PR #428, opened 2022-02-09
emit metric for EC2 describeInstance calls

jyotimahapatra:
Emitting a metric for the count of EC2 DescribeInstance calls that the authenticator makes

### DIFF
--- a/pkg/ec2provider/ec2provider.go
+++ b/pkg/ec2provider/ec2provider.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/aws-iam-authenticator/pkg"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/httputil"
+	"sigs.k8s.io/aws-iam-authenticator/pkg/metrics"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/ec2provider/ec2provider.go
```diff
@@ -54,10 +55,11 @@ type ec2Requests struct {
...
 type ec2ProviderImpl struct {
-	ec2                ec2iface.EC2API
-	privateDNSCache    ec2PrivateDNSCache
-	ec2Requests        ec2Requests
-	instanceIdsChannel chan string
+	ec2                  ec2iface.EC2API
+	privateDNSCache      ec2PrivateDNSCache
+	ec2Requests          ec2Requests
+	instanceIdsChannel   chan string
+	authenticatorMetrics metrics.Metrics
```
nckturner, line 62: We shouldn't need this line anymore?

jyotimahapatra: right...its an artifact of bad merge from master..removing it

jyotimahapatra: done

pkg/metrics/metrics.go
```diff
@@ -49,5 +50,12 @@ func CreateMetrics(reg prometheus.Registerer) Metrics {
 			},
 			[]string{"result"},
 		),
+		EC2DescribeInstanceCallCount: factory.NewCounter(
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "ec2_describe_instances_call",
```
wongma7, line 56: should be "ec2_describe_instances_calls" (plural) or "ec2_describe_instances_calls_total" according to https://prometheus.io/docs/practices/naming/

wongma7: prefer ec2_describe_instances_calls_total cuz it is an 'accumulating count'

nckturner: Yeah I think it should be `_total` if we go with the prometheus guide.  E.g.

```
process_cpu_seconds_total (for an accumulating count with unit)
```

This will match user's expectations.

PR #430, opened 2022-02-10
add sts error metric

jyotimahapatra:

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -17,7 +17,7 @@ const (
 var authenticatorMetrics Metrics
 
 func InitMetrics(registerer prometheus.Registerer) {
-	authenticatorMetrics = CreateMetrics(registerer)
+	authenticatorMetrics = createMetrics(registerer)
 }
 
 func Get() Metrics {
@@ -29,9 +29,11 @@ type Metrics struct {
 	ConfigMapWatchFailures       prometheus.Counter
 	Latency                      *prometheus.HistogramVec
 	EC2DescribeInstanceCallCount prometheus.Counter
+	StsConnectionFailure         prometheus.Counter
+	StsResponses                 *prometheus.CounterVec
 }
 
-func CreateMetrics(reg prometheus.Registerer) Metrics {
+func createMetrics(reg prometheus.Registerer) Metrics {
 	factory := promauto.With(reg)
 
 	return Metrics{
@@ -41,6 +43,20 @@ func CreateMetrics(reg prometheus.Registerer) Metrics {
 				Name:      "configmap_watch_failures_total",
 				Help:      "EKS Configmap watch failures",
 			},
+		),
+		StsConnectionFailure: factory.NewCounter(
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "sts_connection_failures_total",
+				Help:      "Sts call could not succeed or timedout",
+			},
+		),
+		StsResponses: factory.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "sts_responses_total",
+				Help:      "Sts responses with error code label",
+			}, []string{"ResponseCode"},
 		),
 		Latency: factory.NewHistogramVec(
 			prometheus.HistogramOpts{
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -41,6 +41,7 @@ import (
 	clientauthv1beta1 "k8s.io/client-go/pkg/apis/clientauthentication/v1beta1"
 	"sigs.k8s.io/aws-iam-authenticator/pkg"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
+	"sigs.k8s.io/aws-iam-authenticator/pkg/metrics"
 )
 
 // Identity is returned on successful Verify() results. It contains a parsed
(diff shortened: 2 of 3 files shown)


pkg/token/token.go
```diff
@@ -514,6 +515,7 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 		if urlErr, ok := err.(*url.Error); ok {
 			return nil, NewSTSError(fmt.Sprintf("error during GET: %v", urlErr.Err))
```
wongma7, line 516: why not inc here?

jyotimahapatra: right..should do here too...sending an update

pkg/metrics/metrics.go
```diff
@@ -41,6 +43,20 @@ func CreateMetrics(reg prometheus.Registerer) Metrics {
...
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "sts_connection_failures",
+				Help:      "Sts call could not succeed or timedout",
+			},
+		),
+		StsResponseError: factory.NewCounter(
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "sts_failures",
```
wongma7, line 56: wat about name sts_request_failures instead? something more specific

pkg/metrics/metrics.go
```diff
@@ -41,6 +43,20 @@ func CreateMetrics(reg prometheus.Registerer) Metrics {
...
 			},
 		),
+		StsConnectionFailure: factory.NewCounter(
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "sts_connection_failures",
+				Help:      "Sts call could not succeed or timedout",
+			},
+		),
+		StsResponseErrorCodeNot200: factory.NewCounter(
```
nckturner, line 53: I think we should have a metric called `sts_responses_total` and use labels for the exact response code.  This is in line with the recommendations here https://prometheus.io/docs/practices/instrumentation/#use-labels. This will also be much more useful in the long run then only knowing if the code is not 200.

pkg/metrics/metrics.go
```diff
@@ -41,6 +43,20 @@ func CreateMetrics(reg prometheus.Registerer) Metrics {
 				Help:      "EKS Configmap watch failures",
 			},
 		),
+		StsConnectionFailure: factory.NewCounter(
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "sts_connection_failures",
```
nckturner, line 49:
```suggestion
				Name:      "sts_connection_failures_total",
```

pkg/token/token_test.go
```diff
@@ -152,7 +160,7 @@ func TestSTSEndpoints(t *testing.T) {
 }
 
 func TestVerifyTokenPreSTSValidations(t *testing.T) {
-	b := make([]byte, maxTokenLenBytes+1, maxTokenLenBytes+1)
```
nckturner, line 155: Just curious, what's the reason for removing capacity?

pkg/token/token.go
```diff
@@ -564,7 +567,7 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 func hasSignedClusterIDHeader(paramsLower *url.Values) bool {
 	signedHeaders := strings.Split(paramsLower.Get("x-amz-signedheaders"), ";")
 	for _, hdr := range signedHeaders {
-		if strings.ToLower(hdr) == strings.ToLower(clusterIDHeader) {
```
nckturner, line 567: Nit: This seems like an improvement but should be in a different PR

pkg/token/token_test.go
jyotimahapatra: the IDE told me to? it was redundant

jyotimahapatra: i can get rid of it for now

pkg/token/token.go
jyotimahapatra: i can get rid of it for now

pkg/metrics/metrics.go
jyotimahapatra: I was following the metric convention from https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats due to my familiarity. We can use labels here

pkg/token/token.go
```diff
@@ -523,6 +525,7 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 		return nil, NewSTSError(fmt.Sprintf("error reading HTTP result: %v", err))
 	}
 
+	metrics.Get().StsResponseErrorCode.WithLabelValues(fmt.Sprint(response.StatusCode)).Inc()
```
wongma7, line 528: but do we care about 200s?  is 200 an error code? why is the metric name sts_responses_total if the variable name is sts_response_error_code (it's hard to tell intention from the naming)

nckturner: Yeah we should probably either rename `StsResponseErrorCode` or only increment for `response.StatusCode != 200`.  The benefit of including successes is the ability of getting success % metric.

nckturner: So the first option would be to just rename `StsResponseErrorCode` to `StsResponseCode`.

jyotimahapatra: yes..i missed that

jyotimahapatra: done..changed the variable name

nckturner:
@jyoti please fix the merge conflict

jyotimahapatra:
@nckturner ..rebased and ready

PR #431, opened 2022-02-10
Simplify goreleaser Dockerfiles

jyotimahapatra:
aws-iam-authenticator intends to move away from fat images into distroless images.
`goreleaser` is the current choice of tool to create the release binary and images. This PR intends to use the eks-distro minimal distroless image to create simple amd and arm multiarch images and publish them.


.goreleaser.yaml
```diff
@@ -21,130 +21,24 @@ builds:
       - "-s -w -X pkg.Version={{.Version}} -X pkg.CommitID={{.Commit}} -buildid=''"
 
 dockers:
```
nckturner, line 23: Looks good! I think we are missing `build_flag_templates`:   

```
build_flag_templates:
  - "--platform=linux/arm64/v8"
```


(ref https://goreleaser.com/customization/docker_manifest/)

Dockerfile.goreleaser
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 by the contributors.
+# Copyright 2019 The Kubernetes Authors.
```
nckturner, line 1: nit: Might as well do 2022 if its changing.

.goreleaser.yaml
jyotimahapatra: i saw this one..but i got rid of it since im using `goos` instead

nckturner: The official goreleaser docs use `build_flag_templates:` for manifests instead of `goos` which is just set to linux so I'd prefer that, unless you have some reason to prefer `goos` (is it even equivalent for docker manifest builds?)

jyotimahapatra: my assumption was incorrect reading through https://github.com/goreleaser/goreleaser/blob/8e1b6a1bbb39712d92ea1b4bee2b424a59823522/www/docs/customization/docker.md?plain=1#L46

Fixed the platform

Issue #440, opened 2022-03-16
Breaking changes with patch release


edify42:
Hey Maintainers :wave: 

Bit of a funny one but I feel like such changes should be breaking and be reflected as such with a major release?

https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/0221afb8e2a5d14a20a93473edc4c2aa7676ce95

No point being shy with version numbers!

For more context, I was running version `0.5.0` and a team  mate was running `0.5.5` and we found though we had the same `kube_config` the latter would through the API error:
```
Error: UPGRADE FAILED: Kubernetes cluster unreachable: Get "https://our-cluster.aws-region.eks.amazonaws.com/version?timeout=32s": getting credentials: exec plugin is configured to use API version client.authentication.k8s.io/v1alpha1, plugin returned version client.authentication.k8s.io/v1beta1

```

nckturner:
Thanks for pointing this out, we will revert and bump the major version.

nckturner:
Sorry for the delay.  So I think with https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/439, its no longer a breaking change and we don't need to bump the major version.  I'm inclined to put out an v0.5.6 for that reason since there's only a handful of changes.  LMK if anyone thinks differently.

nckturner:
I'll add that I would like to do a 1.0 release, but perhaps we can discuss it separately.

edify42:
Thanks for your response @nckturner :) 

I'm currently stuck trying to figure out how to solve our macos issues as the [EKS docs](https://docs.aws.amazon.com/eks/latest/userguide/create-kubeconfig.html) say we should continue to use `v1alpha1` and the newly built (but still older than 0.5.2 versions of) `aws-iam-authenticator` binaries throw this error on our systems:

```
./aws-iam-authenticator
fatal error: runtime: bsdthread_register error
```

Hence my annoyed tone 😅

edify42:
Thanks @wongma7 as well for reverting the commit. Looking forward to 0.5.6 🎉 

It's less of a headache for us now 👍

PR #443, opened 2022-03-24
Revert "[pkg/token]: Update credential API version"

wongma7:
This reverts commit 0221afb8e2a5d14a20a93473edc4c2aa7676ce95.

See https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/440, this should not have been released in a patch release. Let's revert, make a 0.5.6, and then release 1.x.x (and maintain a 0.5.x branch if necessary).

(context: i updated authenticator today and encountered same issue as https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/440)


wongma7:
closing in favor of https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/439, actually https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/439 to support both alpha and beta will be less effort then this which i thought would be a simple revert : (

PR #454, opened 2022-05-03
revert use of upstream yaml parsing

jyotimahapatra:
We found that users have a dependency on an undocumented behavior of the permissive yaml library that was changed [here](https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/a51796fd152aba4afe0c28c41bf01076e406d071).

Using a block like below caused the yaml parsing to fail the whole `roleArn` section. This is a contract change and should be done behind a feature flag . Any yaml value starting with `{` would cause the failure.
If not done, customers upgrading to the new server version would see mapping error when cache is lost on authenticato restart.
```
    - groups:
      - systems:master
      rolearn: arn:aws:iam::xxx:role/testrole
      username: {{SessionName}}
```

### DIFF
--- a/pkg/mapper/configmap/configmap.go
+++ b/pkg/mapper/configmap/configmap.go
@@ -2,11 +2,12 @@ package configmap
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
-	"fmt"
-	"strings"
 	"sync"
+
+	"fmt"
+
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


willthames:
We were a customer affected by this during the 1.22 upgrade, and while it was frustrating, I'd argue that the change that this reverts would be fine if documented in the 1.22 upgrade notes.

dshackith:
We just had a production issue due to this. We are running 1.21 and a normal AWSWesleyClusterManagerLambda-Add-AddonManagerRole process ran against each cluster and generated the following error: 

"There was an error parsing the config maps.  Only saving data that was good, error parsing config map: [yaml: invalid map key: map[interface {}]interface {}{\"SessionName\":interface {}(nil)}]"

This broke our aws-auth file and nodes were no longer bound to the control plane, and our user mappings were broken, preventing us from accessing the cluster.

Our aws-auth format using {{AccountID}} and {{SessionID}}has been in use since 2018 without issue.

This is a serious oversight and should be corrected before more orgs are impacted by it.

dshackith:
I would like to also note that the use of {{AccountID}} and {{SessionID}} is mentioned in the [README](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/README.md#full-configuration-format), so that seems fairly well documented.

Issue #461, opened 2022-05-31
Get rid of `vendor/` directory


jaypipes:
It's no longer 2017. We should use Go modules and not vendor the world in our source code repository.

- [ ] [Run `go mod download` from within the Dockerfile](https://github.com/aws-controllers-k8s/code-generator/blob/6acf40fe3e3cfd97b799ef7cbf1e89e01c3db8f7/Dockerfile#L31) (in a separate Dockerfile stage)
- [ ] Call `go build` directly from Dockerfile (no need to use Makefile targets which are confusing and harder to understand for many developers IMHO)
- [ ] Delete `vendor/` directory entirely.

PR #463, opened 2022-06-09
remove vendor directory

richzli:
closes issue #461

did not address the make dependency in the Dockerfile; the Makefile pulls a lot of environment variables + command outputs that would be difficult to replicate. it seems to work fine as-is, and that bullet can be left for later


jaypipes (approved):
`+1 −1,308,752` <-- that's what I love to see! :)

PR #464, opened 2022-06-10
Lowercase the ARN keys

mlissner:
I recently deployed this tool and it was working well, but a couple days ago it broke without us having made any changes to the cluster. Out of nowhere, our users were unable to take actions on the cluster that they could previously do. For example, this started failing whereas it was working a few days prior:

```
↪ kubectl describe nodes
error: You must be logged in to the server (Unauthorized)
```

To fix the problem, we changed the keys in our `configmap` manifest from `userARN` to `userarn` and applied that manifest. This PR changes the readme to have the lowercased versions. To check our work, we changed the keys back to uppercase, and the authentication problem came right back, so we're certain lowercased versions work and uppercased versions do not.

What's more, we're certain that the upper case version _was_ working, and we're very confused about what changed to make it stop working (we use EKS, so perhaps there was an upgrade in the background?). 

So here's our fix to the documentation, so others don't have the issue. Does anybody know why this might have broken under us? 

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,7 +408,7 @@ server:
   #     transliteration (available in version >= 0.5).
   mapRoles:
   # statically map arn:aws:iam::000000000000:role/KubernetesAdmin to cluster admin
-  - roleARN: arn:aws:iam::000000000000:role/KubernetesAdmin
+  - rolearn: arn:aws:iam::000000000000:role/KubernetesAdmin
     username: kubernetes-admin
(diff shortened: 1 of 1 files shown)


nckturner:
Can you DM me the details of your cluster? (e.g. on Kubernetes slack, my name and picture is the same)  I can check if there was an update or if something happened.

PR #10, opened 2017-10-14
Support pre-generating assets by running the init sub-command

mumoshu:
This adds `init` sub-command to be used to pre-generate all the assets required by the kubernetes-aws-authenticator server.
Generated assets should be installed on master nodes of your kubernetes cluster.

The server has been modified not to generate a key, a certificate and a kubeconfig by default accordingly.

Resolves #5

## Usage

Before provisioning your cluster with a cluster provisioning tool like kops, kube-aws, etc., run:

```
kubernetes-aws-authenticator -c config.yaml init
```

so that it generates the following assets:

* ./key.pem
* ./cert.pem
* ./kubernetes-aws-authenticator.kubeconfig

In the default configuration, those files should respectively put to:

* /var/kubernetes-aws-authenticator/key.pem
* /var/kubernetes-aws-authenticator/cert.pem
* The path passed to the apiserver via `--authentication-token-webhook-config-file`

Finally, deploy kubernetes-aws-authenticator to your kubernetes cluster as usual.

## Testing

This change is verified to work on a kube-aws cluster with a kubernetes-aws-authenticator deployment.

```
$ kubernetes-aws-authenticator init -c webhook.yaml
INFO[2017-10-14T19:56:16+09:00] generated a new private key and certificate   certBytes=818 keyBytes=1190
INFO[2017-10-14T19:56:16+09:00] saving new key and certificate                certPath=cert.pem keyPath=key.pem
INFO[2017-10-14T19:56:16+09:00] please put ./cert.pem to /var/kubernetes-aws-authenticator/cert.pem on kubernetes master node(s)
INFO[2017-10-14T19:56:16+09:00] please put ./key.pem to /var/kubernetes-aws-authenticator/key.pem on kubernetes master node(s)
INFO[2017-10-14T19:56:16+09:00] please put kubernetes-aws-authenticator.kubeconfig to /etc/kubernetes/kubernetes-aws-authenticator.kubeconfig on kubernetes master node(s)

$ kubectl --token $(kubernetes-aws-authenticator token -c webhook.yaml) --user kubernetes-admin get no
NAME                                            STATUS    AGE       VERSION
ip-10-0-0-114.ap-northeast-1.compute.internal   Ready     40m       v1.7.6+coreos.0
ip-10-0-0-184.ap-northeast-1.compute.internal   Ready     1h        v1.7.6+coreos.0

$ cat kubeconfig
apiVersion: v1
kind: Config
clusters:
- cluster:
    certificate-authority: credentials/ca.pem
[...]

PR #465, opened 2022-06-10
Allow init to respect the state directory and kubeconfig paths

thefirstofthe300:
I'm working on automating the install of AWS IAM authenticator as much as possible. It would be nice if init would just place the certificates in the correct place on boot.

### DIFF
--- a/cmd/aws-iam-authenticator/init.go
+++ b/cmd/aws-iam-authenticator/init.go
@@ -36,19 +36,15 @@ var initCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		localCfg := cfg
-		localCfg.GenerateKubeconfigPath = "aws-iam-authenticator.kubeconfig"
-		localCfg.StateDir = "./"
-
-		err = localCfg.GenerateFiles()
+		err = cfg.GenerateFiles()
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "could not initialize: %v\n", err)
 			os.Exit(1)
 		}
 
-		logrus.Infof("copy %s to %s on kubernetes master node(s)", localCfg.CertPath(), cfg.CertPath())
-		logrus.Infof("copy %s to %s on kubernetes master node(s)", localCfg.KeyPath(), cfg.KeyPath())
-		logrus.Infof("copy %s to %s on kubernetes master node(s)", localCfg.GenerateKubeconfigPath, cfg.GenerateKubeconfigPath)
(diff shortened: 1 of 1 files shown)


nckturner:
This makes a ton of sense to me.  Going back all the way to: https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/10 it seems the idea is that there is some external process which needs to move stuff around.  But if we allow init to place stuff in the correct spot, then we can do init from an initContainer or something.  Can you please sign the CLA?

richzli:
I ran into issues with this while trying to manually set up a cluster, as well.

From the documentation in the repository's README, it does seem like that `init` should be putting the generated files into their respective locations automatically. It'd simplify the process a lot if we didn't have to move the files ourselves afterwards.

thefirstofthe300:
I need to get approval from my company on this. I don't expect any issues, but it may take a couple days.

nckturner:
@thefirstofthe300 I'd like to merge this, can you get the CLA signed today?

cmd/aws-iam-authenticator/init.go
```diff
@@ -36,19 +36,15 @@ var initCmd = &cobra.Command{
...
+		err = cfg.GenerateFiles()
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "could not initialize: %v\n", err)
 			os.Exit(1)
 		}
 
-		logrus.Infof("copy %s to %s on kubernetes master node(s)", localCfg.CertPath(), cfg.CertPath())
-		logrus.Infof("copy %s to %s on kubernetes master node(s)", localCfg.KeyPath(), cfg.KeyPath())
-		logrus.Infof("copy %s to %s on kubernetes master node(s)", localCfg.GenerateKubeconfigPath, cfg.GenerateKubeconfigPath)
+		logrus.Infof("copy %s to %s on kubernetes master node(s)", cfg.CertPath(), cfg.CertPath())
```
nckturner, line 45: This log line doesn't make sense anymore.

nckturner:
So can you address how this change would affect a user who is relying on the current behavior of generating certificates/kubeconfig in the current directory?

nckturner:
Per @wongma7's suggestion, in order to maintain backwards compatibility with the current behavior of the init command, we could generate them in the current directory *and* copy them to the desired location.  Or another option would be to add a feature flag that is default disabled (and warns of deprecated behavior), but when the flag is enabled then we get the desired behavior above.

PR #466, opened 2022-06-21
update k8s dependencies to v1.24

kinarashah:


nilesh-akhade:
@kinarashah This PR includes `go mod vendor` changes too. Is that intentional? The PR includes 300+ file changes which are just result of `go mod vendor`. The commit, https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/466/commits/d0e490d3e017a752db48143ffa34e2005b57a70f alone is sufficient to upgrade to v0.24

nilesh-akhade:
@micahhausler Would you please merge this PR? Go modules causing conflict in our CI, because of the missing package `v1alpha1` in client-go, which is used by [token_test.go](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/v0.5.9/pkg/token/token_test.go#L22)

pallav-trilio:
@micahhausler can you please merge this PR. Go modules causing conflict in our CI, because of the missing package `v1alpha1` in client-go

madhurtrilio:
@micahhausler please merge this soon as it has become a blocker for us to update our packages

nilesh-akhade:
Those who are awaiting for merge, you can go ahead and fork this branch and use the `replace` directive in your *go.mod*.

For lazy people, add the following line in your `go.mod`

    replace sigs.k8s.io/aws-iam-authenticator => github.com/catalogicsoftware/aws-iam-authenticator v0.5.10

melnikalex:
We can’t merge this CR as it's gonna break clients that still have kubectls and kubeconfigs configured for  `v1alpha1` api.

In a nutshell, we'll need to:
* import 2 versions of client-go (this will be much easier if we start using go mods instead of vendors)
* support both v1alpha1 and v1 options from `KUBERNETES_EXEC_INFO`
* default to `v1beta1` if no `KUBERNETES_EXEC_INFO` var is provided

If you could pick this up that'd be amazing. But if not, someone from AWS will pick it up in the next month or so.

For more context see:
* https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuration
* https://github.com/aws/aws-cli/pull/6940

wongma7:
> We can’t merge this CR as it's gonna break clients that still have kubectls and kubeconfigs configured for v1alpha1 api.

no it won't, authenticator does not depend on client-go for this at all, whatever the apiversion is in the kubeconfig the ExecCredential returned is set to the same and this works for v1, v1alpha1, v1beta1, because the ExecCredential returned is valid for all of those.

pkg/token/token_test.go
```diff
@@ -347,11 +346,6 @@ func TestFormatJson(t *testing.T) {
 			EnvKey:           "KUBERNETES_EXEC_INFO",
 			ExpectApiVersion: clientauthv1beta1.SchemeGroupVersion.String(),
 		},
-		{
-			Name:             "KUBERNETES_EXEC_INFO with v1alpha1",
-			EnvKey:           "KUBERNETES_EXEC_INFO",
-			ExpectApiVersion: clientauthv1alpha1.SchemeGroupVersion.String(),
```
wongma7, line 353: keep the test but replace this with client.authentication.k8s.io/v1alpha1.

If there is further doubt about whether simply updating client-go could affect authenticator's compatibility to client.authentication.k8s.io/v1alpha1 i suggest someone implement in a follow-up PR an e2e test to validate kubeconfigs with all versions, v1alpha1 v1beta1, v1. (Could be me, IDK, I'm not volunteering at this time just suggesting)

nckturner:
/close

Vendor was also removed in a previous PR so I'm going to close this in favor of a new PR.

PR #475, opened 2022-07-20
Have init generate files directly in destination folder

richzli:
Continuation of pull request #465

### DIFF
--- a/cmd/aws-iam-authenticator/init.go
+++ b/cmd/aws-iam-authenticator/init.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
@@ -19,6 +19,8 @@ package main
 import (
 	"fmt"
 	"os"
+
+	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


nckturner:
So this is better since its not a breaking change, but the downsides are that its confusing and in general, users will only want the files to be generated in one place.  Therefore I think we should add a something that makes our intentions clear.  I propose a feature gate which will enable the new behavior, and depending on the value of the feature gate the files will either be generated in the current directory or the "proper" destination folder.   The feature gate will start out disabled, and eventually when we think enough time has passed we will flip the feature gate to enabled by default, meaning anyone who breaks at that point but doesn't want to change anything can just disable the feature gate.

cmd/aws-iam-authenticator/init.go
```diff
@@ -36,20 +36,26 @@ var initCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		localCfg := cfg
-		localCfg.GenerateKubeconfigPath = "aws-iam-authenticator.kubeconfig"
-		localCfg.StateDir = "./"
+		oldCfg := cfg
```
nckturner, line 39: nit: `deprecatedCfg` instead of `oldCfg`.

cmd/aws-iam-authenticator/init.go
```diff
@@ -36,20 +36,26 @@ var initCmd = &cobra.Command{
...
+        	if err != nil {
+                	fmt.Fprintf(os.Stderr, "could not initialize: %v\n", err)
+            		os.Exit(1)
+        	}
+
+        	err = cfg.GenerateFiles()
+        	if err != nil {
+            		fmt.Fprintf(os.Stderr, "could not initialize: %v\n", err)
+            		os.Exit(1)
+        	}
```
nckturner, line 53: Something like: 
```go
		if feature.DefaultMutableFeatureGate.Enabled(features.ConfigurableInitDir) {
        		if err = cfg.GenerateFiles(); err != nil {
            			fmt.Fprintf(os.Stderr, "could not initialize: %v\n", err)
            			os.Exit(1)
        		}
		} else {
        		if err = deprecatedCfg.GenerateFiles(); err != nil {
               	 	fmt.Fprintf(os.Stderr, "could not initialize: %v\n", err)
            			os.Exit(1)
			}
        	}

```

nckturner (changes requested):

richzli:
The authenticator doesn't seem to implement the features boilerplate code--perhaps a more lightweight solution would be to just use command-line flags? If the flag is not present, then original behavior is preserved.

richzli:
We should probably add tests for this new functionality, too.

nckturner:
> The authenticator doesn't seem to implement the features boilerplate code

Can you get it from importing [component-base](https://github.com/kubernetes/component-base/blob/master/featuregate/feature_gate.go) in this file?

richzli:
I was referring to some of the setup that needed to be done past that (e.g. the apiserver, [here](https://github.com/kubernetes/apiserver/blob/master/pkg/util/feature/feature_gate.go) and [here](https://github.com/kubernetes/apiserver/blob/master/pkg/features/kube_features.go)). I can set up a few files with a similar structure, but my thinking was that it may be a bit redundant considering this is the only feature that the authenticator would have.

richzli:
I see now that SSO roles uses features already... I will use that, then.

cmd/aws-iam-authenticator/root.go
```diff
@@ -119,6 +119,11 @@ func getConfig() (config.Config, error) {
 				logrus.Info("SSORoleMatch feature enabled")
 				config.SSORoleMatchEnabled = true
 			}
+			if strings.Contains(fg, string(config.ConfiguredInitDirectories)) &&
```
nckturner, line 122: Isn't there a utility function to do this?

richzli: There does seem to be. I made a few changes there--the code looks nicer now.

cmd/aws-iam-authenticator/init.go
```diff
@@ -36,19 +38,32 @@ var initCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		localCfg := cfg
-		localCfg.GenerateKubeconfigPath = "aws-iam-authenticator.kubeconfig"
-		localCfg.StateDir = "./"
+		if featureGates.Enabled(config.ConfiguredInitDirectories) {
+			err = cfg.GenerateFiles()
+			if err != nil {
```
nckturner, line 43:
```suggestion
			if err := cfg.GenerateFiles(); err != nil {

```

nckturner (approved):
/lgtm

PR #476, opened 2022-08-03
End-to-end testing framework for IAM authenticator

richzli:
Created an end-to-end testing framework for the authenticator. README in `./hack/e2e`.


tests/e2e/apiserver_test.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+const (
+    restartDelaySec     = 30
+    testTimeoutSec      = 300
+)
+
+var _ = Describe("[apiserver] [Disruptive] the apiserver", func() {
+    f := framework.NewDefaultFramework("auth")
+    // TODO: figure out how to incorporate the framework!
```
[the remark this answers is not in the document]
richzli: I'm honestly not sure about the framework--despite seeing it being used in other projects, I couldn't get it to work here. The kubeconfig is passed through the `--kubeconfig` flag and can be accessed using some package-level variables, but the ClientSet that *should* have been created in f.ClientSet just... didn't work for me when I tried it.

Besides, I don't think you can extract the configuration from the ClientSet anyways, so... yeah, this can probably be deleted.

tests/e2e/tests.go
```diff
@@ -0,0 +1,173 @@
...
+
+            res, err := cs.AuthorizationV1().
+                SelfSubjectAccessReviews().
+                Create(context.TODO(), &selfCheck, metav1.CreateOptions{})
+
+            Expect(err).ToNot(HaveOccurred())
+            Expect(res.Status.Allowed).To(BeFalse())
+
+            // check no access to nodes
+
```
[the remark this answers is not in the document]
richzli: Mildly related: probably need to `gofmt` the code before merging up the PR.

tests/e2e/apiserver_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+        AfterEach(func() {
+            cs.BatchV1().Jobs(kubeSystemNs).Delete(context.TODO(), "apiserver-restarter", metav1.DeleteOptions{})
+        })
+
+        It("restarts successfully", func() {
+            startTime := time.Now()
+            success := false
+
+            for time.Now().Sub(startTime).Seconds() < testTimeoutSec {
```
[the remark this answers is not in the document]
richzli: That's a good idea.

hack/e2e/apiserver-update.yaml
```diff
@@ -0,0 +1,52 @@
...
+          path: /srv/kubernetes/kube-apiserver
+          type: Directory
+
+      containers:
+      - name: apiserver-updater
+        image: busybox
+        # this shell command adds the --authentication-token-webhook-config-file flag
+        # to the apiserver manifest, which routes authentication through the webhook.
+        # it adds the relevant line after the first appearance of the string "- args;",
+        # which is a little messy--can possibly be improved by installing yq and using it (?)
```
wongma7, line 36: Long-term, it would be easier to leverage kops existing mechanism for installing aws-iam-authenticator, and then patching it for the purpose of the test, instead of maintaining hack/e2e/deploy.yaml and these apiserver files.

From https://kops.sigs.k8s.io/authentication/#turning-on-iam-authenticator-on-an-existing-cluster:

kops equivalent to apiserver-update.yaml is:
```
Edit the clusters configuration kops edit cluster ${NAME} and add the Authentication and Authorization configs to the YAML config.
Update the clusters configuration kops update cluster ${CLUSTER_NAME} --yes
Temporarily disable aws-iam-authenticator DaemonSet kubectl patch daemonset -n kube-system aws-iam-authenticator -p '{"spec": {"template": {"spec": {"nodeSelector": {"disable-aws-iam-authenticator": "true"}}}}}'
```
kops requivalent to apiserver-restart.yaml is:
```
Perform a rolling update of the masters kops rolling-update cluster ${CLUSTER_NAME} --instance-group-roles=Master --force --yes
```

Then we just patch the `image` field of DaemonSet to be the image we are testing.

 Another option would be to reuse deploy/example.yaml instead of maintaining hack/e2e/deploy.yaml , this would have the secondary benefit of testing that the excample actually works.

I am making suggestions for what we should do in future because I do think it is a little messy and fragile to be rolling our own installation/update/restart mechanism instead of leveraging the existing ones (the purpose of the test is to test changes to the authenticator itself, not so much the installation/update/restart mechanism) , but I am not blocking this PR based on this.

hack/kops-patch.yaml
```diff
@@ -0,0 +1,145 @@
+spec:
```
wongma7, line 1: this can be left as empty file, i assume it was copy/pasted from ebs so it's not needed by us

hack/e2e/deploy.yaml
```diff
@@ -0,0 +1,310 @@
...
+        args:
+        - server
+        # uncomment if using EKS-Style ConfigMap
+        # - --backend-mode=EKSConfigMap
+        - --config=/etc/aws-iam-authenticator/config.yaml
+        - --state-dir=/var/aws-iam-authenticator
+        - --generate-kubeconfig=/etc/kubernetes/aws-iam-authenticator/kubeconfig.yaml
+      # uncomment if using the Kops Usage instructions https://sigs.k8s.io/aws-iam-authenticator#kops-usage
+      # the kubeconfig.yaml is pregenerated by the 'aws-iam-authenticator init' step
+        - --kubeconfig-pregenerated=true
```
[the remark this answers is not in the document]
richzli: These were leftover comments from the template deploy in the repo README; I can just remove them.

hack/e2e/apiserver-update.yaml
richzli: I see. I remember when I was first setting this up and looked into kops authenticator as an option, but couldn't figure out a way to put in a custom image. `kubectl patch` may do the trick, though. I'll see what I can do.

richzli: And about the restart--I apply this job in a side test. I didn't want to end up writing some messy `exec.Command` stuff inside of a test suite, so I came up with this instead.

hack/kops-patch.yaml
richzli: Looks like I can use it to patch the cluster file with the authenticator setup.

hack/e2e/deploy.yaml
```diff
@@ -0,0 +1,310 @@
...
+      # - config (mounted from the ConfigMap at /etc/aws-iam-authenticator/config.yaml)
+      # - state (persisted TLS certificate and keys, mounted from the host)
+      # - output (output kubeconfig to plug into your apiserver configuration, mounted from the host)
+      containers:
+      - name: aws-iam-authenticator
+        image: <IMAGE_NAME>
+        args:
+        - server
+        # uncomment if using EKS-Style ConfigMap
+        # - --backend-mode=EKSConfigMap
```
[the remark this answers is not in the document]
richzli: I think the default backend mode (the one without the flag, using a ConfigMap) is the one I'm using, so leaving this commented out is fine. I just left it in in case extra cases for different backend modes are added in the future.

Although now that I think about it, I should probably just get rid of it.

hack/e2e/apiserver-update.yaml
richzli: And just using deploy/example.yaml doesn't work because I added a few extra mappings to it to facilitate testing.

richzli:
If this looks good, I'll rebase it.

hack/e2e/kops.sh
```diff
@@ -61,7 +61,7 @@ function kops_create_cluster() {
     cat <(echo "[") "${CLUSTER_FILE}.tmp" <(echo "]") > "$CLUSTER_FILE"
     rm "${CLUSTER_FILE}.tmp"
 
-    kops_patch_cluster_file "$CLUSTER_FILE" "$KOPS_PATCH_FILE" "$KUBECTL_BIN"
+    kops_patch_cluster_file "$CLUSTER_FILE" "$KOPS_PATCH_FILE" "$KUBECTL_BIN" "$IMAGE_URL"
```
richzli, line 64: (reminder for tomorrow--get rid of the this last argument)

tests/e2e/apiserver_test.go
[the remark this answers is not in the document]
nckturner: Ok, well if this works in prow then its fine for a first pass, we can look into the framework later/when needed.

PR #488, opened 2022-09-15
Add new backend mode DYNAMICFILE

nnmin-aws:
Description:

This PR introduces a new backend mode DYNAMICFILE and an according DYNAMICFILEPATH configurable option to authenticator. It will enable authenticator to load the identity and its mapping from a file specified by DYNAMICFILEPATH  and dynamically reload whenever there is any change in the file.

- [x] make test
- [X] make start-dev-env
- [ ] hack/run/e2e

### DIFF
--- a/cmd/aws-iam-authenticator/root.go
+++ b/cmd/aws-iam-authenticator/root.go
@@ -102,6 +102,7 @@ func getConfig() (config.Config, error) {
 		EC2DescribeInstancesQps:           viper.GetInt("server.ec2DescribeInstancesQps"),
 		EC2DescribeInstancesBurst:         viper.GetInt("server.ec2DescribeInstancesBurst"),
 		ScrubbedAWSAccounts:               viper.GetStringSlice("server.scrubbedAccounts"),
+		DynamicFilePath:                   viper.GetString("server.dynamicfilepath"),
 	}
 	if err := viper.UnmarshalKey("server.mapRoles", &cfg.RoleMappings); err != nil {
 		return cfg, fmt.Errorf("invalid server role mappings: %v", err)
@@ -133,6 +134,20 @@ func getConfig() (config.Config, error) {
 	}
 	if _, ok := partitionMap[cfg.PartitionID]; !ok {
 		return cfg, errors.New("Invalid partition")
+	}
+
+	// DynamicFile BackendMode and DynamicFilePath are mutually inclusive.
+	var dynamicFileModeSet bool
+	for _, mode := range cfg.BackendMode {
+		if mode == mapper.ModeDynamicFile {
+			dynamicFileModeSet = true
+		}
+	}
+	if dynamicFileModeSet && cfg.DynamicFilePath == "" {
+		logrus.Fatal("dynamicfile is set in backend-mode but dynamicfilepath is not set")
+	}
+	if !dynamicFileModeSet && cfg.DynamicFilePath != "" {
+		logrus.Fatal("dynamicfile is not set in backend-mode but dynamicfilepath is set")
 	}
 
 	if errs := mapper.ValidateBackendMode(cfg.BackendMode); len(errs) > 0 {
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -32,8 +32,8 @@ type IdentityMapping struct {
 // list of Kubernetes groups. The username and groups are specified as templates
 // that may optionally contain two template parameters:
 //
-//  1) "{{AccountID}}" is the 12 digit AWS ID.
-//  2) "{{SessionName}}" is the role session name.
+//  1. "{{AccountID}}" is the 12 digit AWS ID.
+//  2. "{{SessionName}}" is the role session name.
 //
 // The meaning of SessionName depends on the type of entity assuming the role.
 // In the case of an EC2 instance role this will be the EC2 instance ID. In the
@@ -169,4 +169,6 @@ type Config struct {
 	// understand we don't need to change
 	EC2DescribeInstancesQps   int
 	EC2DescribeInstancesBurst int
+	//Dynamic File Path for DynamicFile BackendMode
+	DynamicFilePath string
 }
--- a/pkg/mapper/dynamicfile/dynamicfile.go
+++ b/pkg/mapper/dynamicfile/dynamicfile.go
@@ -1,0 +1,244 @@
+package dynamicfile
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
+)
+
+type DynamicFileMapStore struct {
+	mutex sync.RWMutex
+	users map[string]config.UserMapping
+	roles map[string]config.RoleMapping
+	// Used as set.
+	awsAccounts map[string]interface{}
+	filename    string
+}
+
+type DynamicFileData struct {
+	// RoleMappings is a list of mappings from AWS IAM Role to
+	// Kubernetes username + groups.
+	RoleMappings []config.RoleMapping `json:"mapRoles"`
+
+	// UserMappings is a list of mappings from AWS IAM User to
+	// Kubernetes username + groups.
+	UserMappings []config.UserMapping `json:"mapUsers"`
+	// AutoMappedAWSAccounts is a list of AWS accounts that are allowed without an explicit user/role mapping.
+	// IAM ARN from these accounts automatically maps to the Kubernetes username.
+	AutoMappedAWSAccounts []string `json:"mapAccounts"`
+}
+
+type ErrParsingMap struct {
+	errors []error
+}
+
+func (err ErrParsingMap) Error() string {
+	return fmt.Sprintf("error parsing dynamic file: %v", err.errors)
+}
+
+func waitUntilFileAvailable(filename string) error {
+	for {
+		_, err := os.Stat(filename)
+		if os.IsNotExist(err) {
+			time.Sleep(1 * time.Second)
+			continue
+		} else {
+			return err
+		}
+	}
+}
+
+func (m *DynamicFileMapStore) loadDynamicFile() error {
+	err := waitUntilFileAvailable(m.filename)
+	if err != nil {
+		logrus.Errorf("LoadDynamicFile: failed to wait till dynamic file available %v", err)
+		return err
+	}
+	logrus.Infof("LoadDynamicFile: %v is available. loading", m.filename)
+	// load the initial file content into memory
+	userMappings, roleMappings, awsAccounts, err := ParseMap(m.filename)
+	if err != nil {
+		logrus.Errorf("LoadDynamicFile: There was an error parsing the dynamic file: %+v. Map is not updated. Please correct dynamic file", err)
+		return err
+	} else {
+		m.saveMap(userMappings, roleMappings, awsAccounts)
+	}
+	return nil
+}
+
+func NewDynamicFileMapStore(filename string) (*DynamicFileMapStore, error) {
+	ms := DynamicFileMapStore{}
+	ms.filename = filename
+	return &ms, nil
+}
+
+func (m *DynamicFileMapStore) startLoadDynamicFile(stopCh <-chan struct{}) {
+	go wait.Until(func() {
+		m.loadDynamicFile()
+		// start to watch the file change
+		watcher, err := fsnotify.NewWatcher()
+		if err != nil {
+			logrus.Errorf("startLoadDynamicFile: failed when call fsnotify.NewWatcher, %+v", err)
+		}
+		err = watcher.Add(m.filename)
+		if err != nil {
+			logrus.Errorf("startLoadDynamicFile: could not add file to watcher %v", err)
+		}
+
+		defer watcher.Close()
+		for {
+			select {
+			case <-stopCh:
+				return
+			case event := <-watcher.Events:
+				switch {
+				case event.Op&fsnotify.Write == fsnotify.Write, event.Op&fsnotify.Create == fsnotify.Create:
+					// reload the access entry file
+					logrus.Info("startLoadDynamicFile: got WRITE/CREATE event reload it the memory")
+					m.loadDynamicFile()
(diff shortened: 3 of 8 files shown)


melnikalex (changes requested):
Awesome work Min! It's a big PR and I didn't find bugs or anything but have some style/refactoring considerations to make this more maintainable in the future 😛

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+	defer ms.mutex.RUnlock()
+	_, ok := ms.awsAccounts[id]
+	return ok
+}
+
+func (ms *DynamicFileMapStore) LogMapping() {
+	ms.mutex.RLock()
+	defer ms.mutex.RUnlock()
+	for _, user := range ms.users {
+		logrus.Info(user)
```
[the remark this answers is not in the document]
nnmin-aws: This LogMapping() function is not called. I keep this function here only to provide courtesy that user can call this function to log and needn't to modify default logging level.

cmd/aws-iam-authenticator/root.go
```diff
@@ -135,6 +136,20 @@ func getConfig() (config.Config, error) {
 		return cfg, errors.New("Invalid partition")
 	}
 
+	//add validation for DynamicFile BackendMode and DynamicFilePath, they should appear together
```
nckturner, line 139: Style nit: space after // in comments.

Wording nit: slightly more succint description.
```
// DynamicFile BackendMode and DynamicFilePath are mutually inclusive.
```

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+
+	dynamicContent, err := os.ReadFile(filename)
+	if err != nil {
+		logrus.Errorf("ParseMap: could not read from dynamic file")
+		return userMappings, roleMappings, awsAccounts, err
+	}
+
+	var dynamicFileData DynamicFileData
+	err = json.Unmarshal([]byte(dynamicContent), &dynamicFileData)
+	if err != nil {
```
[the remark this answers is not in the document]
nnmin-aws: There is no overlapping with static file mapper as the static file config is read by root.go through viper.

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
+)
+
+type DynamicFileMapStore struct {
+	mutex sync.RWMutex
+	users map[string]config.UserMapping
+	roles map[string]config.RoleMapping
+	// Used as set.
+	awsAccounts map[string]interface{}
```
nckturner, line 21: Its common in Kubernetes org projects to use the explicitly defined `sets.String` when a set is desired.  Its a part of k8s.io/apimachinery (https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go).  I would recommend it because its more readable.

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+	filename    string
+}
+
+type DynamicFileData struct {
```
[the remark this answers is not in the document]
nnmin-aws: it is different from `Config` struct, can't reuse the struct. I referred to [this](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/mapper/configmap/configmap.go#L26)

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+		}
+		break
+	}
+	return nil
+}
+
+func NewDynamicFileMapStore(filename string) (*DynamicFileMapStore, error) {
+	ms := DynamicFileMapStore{}
+	ms.filename = filename
```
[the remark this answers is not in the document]
nnmin-aws: The mapper will wait for this file if it is not found. so whether the file was found or not, it shouldn't bother a log?

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+type ErrParsingMap struct {
+	errors []error
+}
+
+func (err ErrParsingMap) Error() string {
+	return fmt.Sprintf("error parsing dynamic file: %v", err.errors)
+}
+
+func waitUntilFileAvailable(filename string) error {
+	waitFileTimer := time.NewTimer(1 * time.Second)
```
nckturner, line 47: I agree that the timer is awkward here, because you have to reset it every iteration.  However, I think this code should be refactored.  I think we might want to consider using a separate goroutine which simple waits for the file, then sets up fsnotify.  If the file doesn't exist or is deleted, the goroutine can exit and be restarted by a ticker or it can handle that internally. This goroutine will handle everything in terms of fsnotify, reading the file, and unmarshalling it into a struct. 

The goroutine should be managed by something like `wait.Until()`, such as in `go wait.Until(c.runWorker, time.Second, ctx.Done())`.  This allows us to manage the goroutine with a context and to control how frequently the outermost loop runs.  

We can then either save the struct into the cache in this same goroutine, or if any additional processing needs to happen we can follow the pattern of adding a cache lookup key to a work queue for a worker to pick up.

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+					if os.IsNotExist(err) {
+						//the "REMOVE" event is  not triggered by vi or cp cmd
+						//reset memory
+						userMappings := make([]config.UserMapping, 0)
+						roleMappings := make([]config.RoleMapping, 0)
+						awsAccounts := make([]string, 0)
+						m.saveMap(userMappings, roleMappings, awsAccounts)
+					}
+					//wait for the file
+					err = waitUntilFileAvailable(m.filename)
```
nckturner, line 124: If using my suggestion above, this should end the goroutine and it should be restarted by something like `go wait.Until(watchDynamicFile, time.Second, ctx.Done())`.  This will remove the need for duplicate calls to `waitUntilFileAvailable()`.

pkg/mapper/dynamicfile/mapper.go
```diff
@@ -0,0 +1,59 @@
...
+func (m *DynamicFileMapper) Name() string {
+	return mapper.ModeDynamicFile
+}
+
+func (m *DynamicFileMapper) Start(stopCh <-chan struct{}) error {
+	m.startLoadDynamicFile(stopCh)
+	return nil
+}
+
+func (m *DynamicFileMapper) Map(canonicalARN string) (*config.IdentityMapping, error) {
```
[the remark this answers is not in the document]
nnmin-aws: This Map function is defined in an [interface](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/mapper/mapper.go#L36) already.

pkg/mapper/dynamicfile/dynamicfile.go
melnikalex: I had a comment about a refactor as well but looks like it wasn't saved. Another idea I had for a refactor was:
```
for {
  select {
    case <-stopCh:
    case fsnotifyevent: // any event
       // add mappings if file updated
       // clear mappings if file removed
    default:
       setupFileIfAvailable
       // this bit can do the job of waitUntilFileAvailable
  }
}
```

This would remove the fencepost problem of settting up a file before entering the forloop, and also consolidate some of the duplicate code that's in the select statement right now. 

Just an option :)

pkg/mapper/dynamicfile/dynamicfile.go
nnmin-aws: _This goroutine will handle everything in terms of fsnotify, reading the file, and unmarshalling it into a struct_ -- is this what we have here?

_We can then either save the struct into the cache in this same goroutine, or if any additional processing needs to happen we can follow the pattern of adding a cache lookup key to a work queue for a worker to pick up._. -- is this adding more complexity?

nckturner: The difference, which I did not clearly state, is that the goroutine should terminate and be restarted at an interval using `wait.Until().` It should also respect a context, and terminate when `context.Done()`.  1 goroutine might be fine in this case, but if additional complexity arises outside of reading from the file then I'd advocate for separating into 2 goroutines with distinct jobs.  Also, what about locking between this reader and the file writer?  I took a quick glance at fsnotify and it wasn't clear to me if we'd need the ability to use something like [flock](https://www.man7.org/linux/man-pages/man2/flock.2.html) too.

cmd/aws-iam-authenticator/root.go
```diff
@@ -135,6 +136,20 @@ func getConfig() (config.Config, error) {
...
+	var dynamicFileModeSet bool
+	for _, mode := range cfg.BackendMode {
+		if mode == "DynamicFile" {
```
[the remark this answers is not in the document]
nnmin-aws: thank you for the catch. updated

cmd/aws-iam-authenticator/root.go
```diff
@@ -102,6 +102,7 @@ func getConfig() (config.Config, error) {
 		EC2DescribeInstancesQps:           viper.GetInt("server.ec2DescribeInstancesQps"),
 		EC2DescribeInstancesBurst:         viper.GetInt("server.ec2DescribeInstancesBurst"),
 		ScrubbedAWSAccounts:               viper.GetStringSlice("server.scrubbedAccounts"),
+		DynamicFilePath:                   viper.GetString("server.dynamicfilepath"),
 	}
 	if err := viper.UnmarshalKey("server.mapRoles", &cfg.RoleMappings); err != nil {
```
melnikalex, line 107: I think we should define an "NestedMappings" struct in `[pkg/config/types.go](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/config/types.go)` and embed it in both the viper config as well as use it for the dynamic file mode. There will be minimal changes required for the existing code and a lot of cutting of the code from this CR.

I made an example of how it works here: https://go.dev/play/p/RXYXdE3mvoI

Additional resources:
* https://gobyexample.com/struct-embedding
* https://go.dev/doc/effective_go#embedding

pkg/mapper/dynamicfile/dynamicfile.go
nnmin-aws: if there is any pending change when the reader read the file, it is ok. When the pending change is completed, fsnotify will trigger a new WRITE event, reader will read the file again. If we put a lock there, then reader will be blocked till the writer completes. Then fsnotify still will trigger a WRITE event. So wondering flock is not necessary here

pkg/mapper/mapper.go
```diff
@@ -20,15 +20,17 @@ const (
 	ModeEKSConfigMap string = "EKSConfigMap"
 
 	ModeCRD string = "CRD"
+
+	ModeDynamicFile string = "DynamicFile"
```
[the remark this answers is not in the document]
nckturner: A type in the config package would be nice to specify the allowed values in one place.  Change it here too: https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/config/types.go#L166

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,244 @@
...
+		logrus.Errorf("LoadDynamicFile: failed to wait till dynamic file available %v", err)
+		return err
+	}
+	logrus.Infof("LoadDynamicFile: %v is available. loading", m.filename)
+	// load the initial file content into memory
+	userMappings, roleMappings, awsAccounts, err := ParseMap(m.filename)
+	if err != nil {
+		logrus.Errorf("LoadDynamicFile: There was an error parsing the dynamic file: %+v. Map is not updated. Please correct dynamic file", err)
+		return err
+	} else {
```
minj131, line 71: style nit but the else here is not needed

pkg/config/types.go
```diff
@@ -169,4 +169,6 @@ type Config struct {
 	// understand we don't need to change
 	EC2DescribeInstancesQps   int
 	EC2DescribeInstancesBurst int
+	//Dynamic File Path for DynamicFile BackendMode
```
minj131, line 172: small nit, can you add a space after the `//`

cmd/aws-iam-authenticator/root.go
```diff
@@ -102,6 +102,7 @@ func getConfig() (config.Config, error) {
 		EC2DescribeInstancesQps:           viper.GetInt("server.ec2DescribeInstancesQps"),
 		EC2DescribeInstancesBurst:         viper.GetInt("server.ec2DescribeInstancesBurst"),
 		ScrubbedAWSAccounts:               viper.GetStringSlice("server.scrubbedAccounts"),
+		DynamicFilePath:                   viper.GetString("server.dynamicfilepath"),
```
minj131, line 105: double checking, is the key supposed to be all lower case?

cmd/aws-iam-authenticator/root.go
```diff
@@ -135,6 +136,20 @@ func getConfig() (config.Config, error) {
 		return cfg, errors.New("Invalid partition")
 	}
 
+	//add validation for DynamicFile BackendMode and DynamicFilePath, they should appear together
+	var dynamicFileModeSet bool
+	for _, mode := range cfg.BackendMode {
```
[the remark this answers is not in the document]
nnmin-aws: updated

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -0,0 +1,262 @@
...
+
+type ErrParsingMap struct {
+	errors []error
+}
+
+func (err ErrParsingMap) Error() string {
+	return fmt.Sprintf("error parsing dynamic file: %v", err.errors)
+}
+
+func waitUntilFileAvailable(filename string) error {
```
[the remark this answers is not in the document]
nnmin-aws: updated

cmd/aws-iam-authenticator/root.go
nnmin-aws: checked above code, actually there is no fixed style here.                
                Address:                           viper.GetString("server.address"),
		Kubeconfig:                     viper.GetString("server.kubeconfig"),
		Master:                            viper.GetString("server.master"),

PR #497, opened 2022-10-06
add a configurable option to switch yaml parser between yaml and utilyaml

nnmin-aws:
Add a configurable option to switch yaml parser between yaml and utilyaml. refer to https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/455

### DIFF
--- a/pkg/mapper/configmap/configmap.go
+++ b/pkg/mapper/configmap/configmap.go
@@ -115,15 +115,19 @@ func ParseMap(m map[string]string) (userMappings []config.UserMapping, roleMappi
 	rawUserMappings := make([]config.UserMapping, 0)
 	userMappings = make([]config.UserMapping, 0)
 	if userData, ok := m["mapUsers"]; ok {
-		userJson, err := utilyaml.ToJSON([]byte(userData))
+		if EKSYaml {
+			//use Lenient Yaml Parser
+			err = yaml.Unmarshal([]byte(userData), &rawUserMappings)
+		} else {
+			//use Strict Yaml Parser
+			userJson, err := utilyaml.ToJSON([]byte(userData))
+			if err == nil {
+				err = json.Unmarshal(userJson, &rawUserMappings)
+			}
+		}
 		if err != nil {
 			errs = append(errs, err)
 		} else {
-			err = json.Unmarshal(userJson, &rawUserMappings)
-			if err != nil {
-				errs = append(errs, err)
-			}
-
 			for _, userMapping := range rawUserMappings {
 				err = userMapping.Validate()
 				if err != nil {
@@ -138,15 +142,20 @@ func ParseMap(m map[string]string) (userMappings []config.UserMapping, roleMappi
 	rawRoleMappings := make([]config.RoleMapping, 0)
 	roleMappings = make([]config.RoleMapping, 0)
 	if roleData, ok := m["mapRoles"]; ok {
-		roleJson, err := utilyaml.ToJSON([]byte(roleData))
+		if EKSYaml {
+			//use Lenient Yaml Parser
+			err = yaml.Unmarshal([]byte(roleData), &rawRoleMappings)
+		} else {
+			//use Strict Yaml Parser
+			roleJson, err := utilyaml.ToJSON([]byte(roleData))
+			if err == nil {
+				err = json.Unmarshal(roleJson, &rawRoleMappings)
+			}
+		}
+
 		if err != nil {
 			errs = append(errs, err)
 		} else {
-			err = json.Unmarshal(roleJson, &rawRoleMappings)
-			if err != nil {
-				errs = append(errs, err)
-			}
-
 			for _, roleMapping := range rawRoleMappings {
 				err = roleMapping.Validate()
 				if err != nil {
--- a/pkg/mapper/configmap/mapper.go
+++ b/pkg/mapper/configmap/mapper.go
@@ -13,7 +13,10 @@ type ConfigMapMapper struct {
 
 var _ mapper.Mapper = &ConfigMapMapper{}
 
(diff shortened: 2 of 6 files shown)


pkg/mapper/configmap/configmap.go
```diff
@@ -115,15 +115,19 @@ func ParseMap(m map[string]string) (userMappings []config.UserMapping, roleMappi
 	rawUserMappings := make([]config.UserMapping, 0)
 	userMappings = make([]config.UserMapping, 0)
 	if userData, ok := m["mapUsers"]; ok {
-		userJson, err := utilyaml.ToJSON([]byte(userData))
+		if LenientYamlEKS {
```
nckturner, line 118: IMO one improvement would be push this logic into a utility.  We can have a `pkg/config/yaml` which includes our own implementation of yaml.Unmarshal which makes this decision based on a struct variable.  Then its set when the struct is initialized and we can do that once when authenticator starts up based on the flag value.

pkg/mapper/configmap/configmap.go
```diff
@@ -115,15 +115,19 @@ func ParseMap(m map[string]string) (userMappings []config.UserMapping, roleMappi
...
+			//use Lenient Yaml Parser
+			err = yaml.Unmarshal([]byte(userData), &rawUserMappings)
+		} else {
+			//use Strict Yaml Parser
```
wongma7, line 122: lenient/strict is not 100% accurate, the "lenient" parser is case insensitive, but it does not konw how to deal with `username: {{SessionName}}`, it expects quotes like`username: "{{SessionName}}"`

can we have a test case for the {{SessionName}} issue, I'll try to find a link to it. But basically we are using {{ in the golang template sense, but {{ also has a meaning in the yaml sense https://yaml.org/spec/1.2.2/#flow-mappings

nnmin-aws: thank you for the comments. there are test yamls https://github.com/kubernetes-sigs/aws-iam-authenticator/tree/master/pkg/mapper/configmap/yaml. do they cover the case you mentioned? this cr is to include https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/455 on master branch, it is only on release-0.5 branch now.

> lenient/strict is not 100% accurate, the "lenient" parser is case insensitive, but it does not konw how to deal with `username: {{SessionName}}`, it expects quotes like`username: "{{SessionName}}"`
> 
> can we have a test case for the {{SessionName}} issue, I'll try to find a link to it. But basically we are using {{ in the golang template sense, but {{ also has a meaning in the yaml sense https://yaml.org/spec/1.2.2/#flow-mappings

thank you for the comments. there are test yamls https://github.com/kubernetes-sigs/aws-iam-authenticator/tree/master/pkg/mapper/configmap/yaml. do they cover the case you mentioned? this cr is to include https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/455 on master branch, it is only on release-0.5 branch now.

pkg/mapper/configmap/configmap.go
nnmin-aws: > IMO one improvement would be push this logic into a utility. We can have a `pkg/config/yaml` which includes our own implementation of yaml.Unmarshal which makes this decision based on a struct variable. Then its set when the struct is initialized and we can do that once when authenticator starts up based on the flag value.

thank you for the comments. this can be a future goal to build a `pkg/config/yaml` if there are more changes about yaml parsing.

pkg/mapper/configmap/mapper.go
```diff
@@ -13,7 +13,10 @@ type ConfigMapMapper struct {
 
 var _ mapper.Mapper = &ConfigMapMapper{}
 
+var LenientYamlEKS = false
```
minj131, line 16: What's the reasoning for this global exported variable here?

pkg/mapper/configmap/configmap.go
minj131: +1 for this idea, also could make a case where any other need for it would just be able to import it from pkg

nnmin-aws: This is only short term fix to solve certain case in EKS configmap parsing.  We are working on long term solution for this. Given essentially it is a short-term fix, I'd like to constrain it only in configmap.go. Because if exposed in a new package, we can't control how the new package will be used potentially. It is harder to clean it up in future.

pkg/mapper/configmap/configmap.go
nnmin-aws: yep the testes cases are here https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/497/files#diff-8b827dc0fc13d692520fe80df4d6a52e15b065812a9f4a7b906f3ded58afd865R92

pkg/mapper/configmap/mapper.go
nnmin-aws: it is only meaningful in this file. it is a short term fix for EKS configmap

minj131: Thoughts on making it unexportable then if it's only meaningful in this package?

pkg/mapper/configmap/configmap.go
```diff
@@ -115,15 +115,19 @@ func ParseMap(m map[string]string) (userMappings []config.UserMapping, roleMappi
 	rawUserMappings := make([]config.UserMapping, 0)
 	userMappings = make([]config.UserMapping, 0)
 	if userData, ok := m["mapUsers"]; ok {
-		userJson, err := utilyaml.ToJSON([]byte(userData))
+		if EKSYaml {
+			//use Lenient Yaml Parser
```
nckturner, line 119: Remove or update comments

PR #501, opened 2022-10-19
add end to end test for dynamicfile backend mode

nnmin-aws:
to run the end to end test:

`make start-dev-dynamicfile`

The command will do the following steps:
(1)create an image of authenticator. 
(2)run a kind cluster and run the authenticator with above image in the container
(3)generate a kubeconfig to authenticate through authenticator token
(4)create an aws IAM role for testing and assumed this role
(5)run `kubectl get nodes` with the role

### DIFF
--- a/hack/e2e-dynamicfile.sh
+++ b/hack/e2e-dynamicfile.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+#
+# An easy way to quickly do end to end test for aws-iam-authenticator
+# after creating the kind cluster and running authenticator in docker.
+#
+
+REPO_ROOT="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
+OUTPUT="${OUTPUT:-${REPO_ROOT}/_output}"
+
+# Location of templates, config files, mounts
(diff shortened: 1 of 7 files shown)


hack/lib/dev-env.sh
```diff
@@ -111,6 +118,22 @@ function write_authenticator_config() {
         "${authenticator_config_template}" > "${authenticator_config}"
 }
 
+function write_authenticator_with_dynamicfile_mode_config() {
```
wongma7, line 121: not huge fan of this _with_dynamicfile_mode_config / _with_dynamic_file pattern. Some possible alternatives :
- have write_authenticator_config take argument
- have write_authenticator_with_dynamicfile_mode_config call write_authenticator_config, and then do dynamicfile specific things (to reduce code duplication)

hack/e2e-dynamicfile.sh
```diff
@@ -0,0 +1,106 @@
...
+access_entry_tmp="${OUTPUT}/dev/authenticator/access-entry/access-entries.tmp"
+access_entry_json="${OUTPUT}/dev/authenticator/access-entry/access-entries.json"
+client_dir="${OUTPUT}/dev/client"
+kubectl_kubeconfig="${client_dir}/kubeconfig.yaml"
+
+REGION=${AWS_REGION:-us-west-2}
+AWS_ACCOUNT=$(aws sts get-caller-identity --query "Account" --output text)
+AWS_TEST_ROLE=${AWS_TEST_ROLE-authenticator-dev-cluster-testrole}
+
+function e2e_dynamicfile(){
```
wongma7, line 39: what about writing this in golang? https://github.com/kubernetes-sigs/aws-iam-authenticator/tree/master/tests/e2e let's say there was a way for the test to modify the dynamic fijle. Then a lot of the code in https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/tests/e2e/tests.go could be reused: modify the file, try authentication, modify it again, try again, etc.

wongma7: the e2e test repo is lacking a README and there is a learning curve. But it is using ginkgo in the same style as k/k https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md, i think it is worth re-implementing this test in golang

wongma7:
/lgtm
/approve

two issues I think should be improved for future for integration with existing CI  but as it is the script is demonstrably useful so fine with merging as-is (that is what "hack" dir is for after all I guess)

PR #507, opened 2022-11-03
Add end to end test for mountfile mode

nnmin-aws:
Add end to end test for mountfile mode in kind
Update Makefile to support run e2e from either kind or kops.

make e2e RUNNER=kind #to run e2e on kind
make e2e RUNNER=kops #to run e2e on kops

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,8 +100,15 @@ integration:
 	./hack/test-integration.sh
 
 .PHONY: e2e
-e2e:
+e2e: bin
+ifeq ($(RUNNER),kops)
 	./hack/e2e/run.sh
+else ifeq ($(RUNNER),kind)
+	./hack/start-dev-env-dynamicfile.sh
+	./hack/stop-dev-env.sh
+else
+	echo "make e2e RUNNER=[kops|kind]"
+endif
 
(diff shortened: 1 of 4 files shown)


nckturner:
I don't think I reviewed the original e2e PR, but wondering if we can unify the e2e test runner under a single makefile command, like `make e2e WHAT=a,b,c RUNNER=[kops|kind]`?

Makefile
```diff
@@ -100,8 +100,15 @@ integration:
 	./hack/test-integration.sh
 
 .PHONY: e2e
-e2e:
+e2e: bin
+ifeq ($(RUNNER),kops)
 	./hack/e2e/run.sh
+else ifeq ($(RUNNER),kind)
+	./hack/start-dev-env-dynamicfile.sh
```
nckturner, line 107: Nit: so the "dev env" for dynamic file inculdes an automatic e2e test run, it would be nice if the UX was the same w.r.t the other dev env scripts.  I think I probably would have preferred modifying the other dev scripts so that there is just one way of creating a dev environment and some configuration options which determine whether or not dynamic file backend mode is used.  But we can clean up this experience as part of a future change.

PR #509, opened 2022-11-07
Revert "Add SSO Role suffix support (#416)"

nnmin-aws:
This reverts commit 55005e174cfca7eb5538b5dedfc409cca9be6d1a.


nckturner:
Please take a look at the failing tests.  Looks like we are using `Validate` in some tests but it is removed in this diff.

nnmin-aws:
> Please take a look at the failing tests. Looks like we are using `Validate` in some tests but it is removed in this diff.

thanks @nckturner. I have updated PR by cherry-picking a commitId. The testing cases pass.

PR #511, opened 2022-11-08
move e2e test from start dev script + minor fix for run.sh

nnmin-aws:
This PR move the e2e test from start-dev-dynamicfile script and do minor fix on run.sh so it can run in local repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,11 @@ BIN_ARCH_LINUX ?= amd64 arm64
 BIN_ARCH_WINDOWS ?= amd64
 BIN_ARCH_DARWIN ?= amd64
 
+#CI is defined in test-infra https://github.com/kubernetes/test-infra/blob/2e3dd84399745eb49cef69afc3ed5bded8a6580c/prow/pod-utils/downwardapi/jobspec.go#L89
+# and passed in when running on github prow
+CI ?= false
+RUNNER ?= kops
+
 ALL_LINUX_BIN_TARGETS = $(foreach arch,$(BIN_ARCH_LINUX),$(OUTPUT)/bin/aws-iam-authenticator_$(VERSION)_linux_$(arch))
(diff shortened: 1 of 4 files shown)


hack/e2e/run.sh
```diff
@@ -39,15 +39,16 @@ INSTANCE_TYPE=${INSTANCE_TYPE:-c5.large}
...
+REPO_NAME=${REPO_NAME:-aws-iam-authenticator}
 ECR_URL=${AWS_ACCOUNT_ID}.dkr.ecr.${REGION}.amazonaws.com
 IMAGE_NAME=${IMAGE_NAME:-${ECR_URL}/${REPO_NAME}}
 IMAGE_TAG=${IMAGE_TAG:-${TEST_ID}}
 
 K8S_VERSION=${K8S_VERSION:-1.22.10}
 
 KOPS_VERSION=${KOPS_VERSION:-1.23.0}
-KOPS_STATE_BUCKET=${KOPS_STATE_BUCKET:-k8s-kops-auth-e2e}
+KOPS_USERID=$(whoami)
```
nckturner, line 50: This might mess up the e2e tests that run in github.  We may need to detect if we are running in github and use `k8s-kops-auth-e2e`, and otherwise use the bucket with `$(whoami)`.

nnmin-aws: this's a good catch. thanks @nckturner. I will fix this

nckturner:
I think we should be able to use the `CI` env var to detect if we are running in prow (triggered by github PRs).  This is what kops uses.  Here is where it should get injected: https://github.com/kubernetes/test-infra/blob/2e3dd84399745eb49cef69afc3ed5bded8a6580c/prow/pod-utils/downwardapi/jobspec.go#L89  I'm not sure why it didn't show up in our prowjob though, do we need to add a preset or something?

Makefile
```diff
@@ -19,6 +19,9 @@ BIN_ARCH_LINUX ?= amd64 arm64
 BIN_ARCH_WINDOWS ?= amd64
 BIN_ARCH_DARWIN ?= amd64
 
+CI ?= false
```
nckturner, line 22: Can we add a comment with a link to the test-infra code where this is set?

nckturner:
Add PR description

PR #474, opened 2022-07-19
solve the problem that client-go cannot find v1alpha1 after upgrading to 0.24

fengshunli:
reference link https://github.com/fengshunli/aws-iam-authenticator/tree/main

PR #516, opened 2022-11-23
Cherry pick: solve the problem that client-go cannot find v1alpha1 after upgrading to 0.24

xdu31:
Cherry pick:
solve the problem that client-go cannot find v1alpha1 after upgrading to 0.24 #474


xdu31:
ok

nckturner:
Overriding merge because the kind e2e test failure is known and requires a cherry-pick which we can do separately but it doesn't need to block this.

Issue #514, opened 2022-11-17
sigs.k8s.io/aws-iam-authenticator v0.6.0


fengshunli:
why does this module introduce the dependency of kubernetes, which makes it impossible to compile now
[sigs.k8s.io/aws-iam-authenticator](http://sigs.k8s.io/aws-iam-authenticator) v0.6.0 Why does this module introduce the dependency of kubernetes, which makes it impossible to compile now

nnmin-aws:
apology for your inconvenience @fengshunli. we are checking this, will get back to you soon.

fengshunli:
the tag of 0.5.10 does not have the pr of the merged submission, and only the tag of v0.6.0 can be used. The dependency of kubernetes was introduced in v0.6.0, which caused the compilation to fail @nnmin-aws

fengshunli:
github.com/fengshunli/aws-iam-authenticator v0.0.0-20220719090955-bc28918e01f1  @pmyjavec  Just replace the 0.5.9 version and use this version of mine

micahhausler:
Can you post full repro information? 
* Failing command
* commit you're building from
* Go version
* OS/arch

nckturner:
Taking a look at this, its correct that including kubernetes as a module dependency breaks external projects which are consuming it.  Looking into whether or not having a separate go module for tests helps anything.

fengshunli:
Failing command
the version introduced by aws-iam-authenticator is lower than the one referenced by the project itself. Is it possible to remove the built-in kubernetes version, or upgrade the same version to solve it?
<img width="1052" alt="image" src="https://user-images.githubusercontent.com/18587688/203694763-6f740ab2-1c46-40a8-bf67-74f879d422e6.png">
Go version and arch
<img width="693" alt="image" src="https://user-images.githubusercontent.com/18587688/203695224-ac82d4aa-b2f2-4b50-a706-25cb11755afe.png">

@nckturner @micahhausler

nnmin-aws:
@fengshunli the new release v0.5.12 is available, please kindly check and it should fix this. https://github.com/kubernetes-sigs/aws-iam-authenticator/releases/tag/v0.5.12

PR #518, opened 2022-11-23
Add go mod for E2E

### LINKED ISSUE #512: aws-iam-authenticator v0.6.0 breaks imports
We use `sigs.k8s.io/aws-iam-authenticator/pkg/token` to allow some of our dev tooling to use IAM credentials for authentication, which means we import `sigs.k8s.io/aws-iam-authenticator` as a module.

In past versions, this works well. Unfortunately, the inclusion of `k8s.io/kubernetes` in the go.mod for aws-iam-authenticator v0.6.0 means it's no longer straight-forward to import, because of all the [local path replacements](https://github.com/kubernetes/kubernetes/blob/v1.22.0/go.mod#L489-L520). This inclusion was introduced by #476, which is a great addition in itself.

For example, with v0.5.4, the following works just fine:

```
❯ mkdir foobar
❯ cd foobar
❯ git init
Initialized empty Git repository in /Users/ripta/projects/foobar/.git/
❯ go mod init github.com/ripta/foobar
go: creating new go.mod: module github.com/ripta/foobar

❯ go get sigs.k8s.io/aws-iam-authenticator@v0.5.4
go: added sigs.k8s.io/aws-iam-authenticator v0.5.4
```

but in v0.6.0, the `go get` step fails with:

```
❯ go get sigs.k8s.io/aws-iam-authenticator@v0.6.0
go: sigs.k8s.io/aws-iam-authenticator@v0.6.0 requires
	k8s.io/kubernetes@v1.22.0 requires
	k8s.io/api@v0.0.0: reading k8s.io/api/go.mod at revision v0.0.0: unknown revision v0.0.0
```

I'm trying to figure out a long-term solution. Right now, I think we'd be stuck having to replicate the [explicit replacements you do](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/v0.6.0/go.mod#L28-L57) in our own go.mod.

I know we're [not the only ones](https://pkg.go.dev/sigs.k8s.io/aws-iam-authenticator/pkg/token?tab=importedby) that import this repo to get access to `pkg/token`.

Thanks!

nckturner:

* Add separate go mod for E2E
* Remove the kubernetes dependency from the main go.mod
* This allows other projects to import aws-iam-authenticator as a go module without including the same replace directives as kubernetes.

Fixes #512, #514


go.mod
```diff
@@ -1,57 +1,84 @@
 module sigs.k8s.io/aws-iam-authenticator
 
-go 1.16
```
nnmin-aws, line 3: this defines lowest go version. will user have to upgrade from 1.16/1.17/1.18?

nckturner: I think we should upgrade the version (at least to the last supported version by the go team, which is 1.18 right now I believe) but I'll do it in a separate PR

micahhausler (approved):
/lgtm
/approve

PR #528, opened 2022-12-01
Tag release on update to version.txt

nckturner:
* This PR will allow tags to be created by users editing the version.txt file. 

Fixes #


version.txt
```diff
@@ -0,0 +1 @@
+0.6.0
```
wongma7, line 1: want to test it with an alpha/beta tag first?

nckturner: Good idea

PR #532, opened 2022-12-06
Bump kubernetes dependency to 1.24.3

nckturner:
Bump Kubernetes dependency to 1.24.8

Fixes #


tamalsaha:
Could you please update this to 1.26 deps, since that is already out?

dims:
Nick, i have one for 1.26.1 k8s here - https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/547

do we want to jump right to that or stay with 1.24.x?

PR #537, opened 2022-12-14
Add automatic release creation

nckturner:

With this change, we will run goreleaser on tag creation to automatically create a release with binaries.  We are also reducing build duplication by remove `Dockerfile.goreleaser` and the docker image build in goreleaser.  Only `Dockerfile` will be used and we will no longer be building images and pushing the 602401143452 account which relied on a manual build step.  Instead, we will be using cloudbuild and promoting images via the CNCF artifact publishing machinery. 

Fixes #

Test release generated here: https://github.com/nckturner/aws-iam-authenticator/releases/tag/v9.9.10 (The admin still needs to click publish after goreleaser runs).


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v3
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v4
+        with:
+          distribution: goreleaser
```
nnmin-aws, line 25: this is awesome! 
Could you please kindly add a comment here or in Readme.md where to locate the new release image?

nckturner: Yes, I'll add something to the README.md but if its ok I'll do it in a follow up PR.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,29 @@
...
+          version: latest
```
jaypipes, line 26: Do we have a process for using goreleaser for non-latest tags?

jaypipes: Never mind... I think this `latest` is referring to *goreleaser*, is that right?

nckturner: Yes, correct, just saying use the latest goreleaser binary... which maybe we should change.

PR #540, opened 2022-12-20
Build s390x/ppc64le binaries

upodroid:

This PR adds support for releasing s390x/ppc64le binaries. I need it for https://github.com/kubernetes/test-infra/pull/28277 as we do multi-arch builds.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -12,6 +12,8 @@ builds:
     goarch:
       - amd64
       - arm64
+      - ppc64le
+      - s390x
     ignore:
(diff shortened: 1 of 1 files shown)


upodroid:
/cc @dims @nckturner 

Can I get this merged please and a new release cut after the holidays? I need it to allow Prow to authenticate with EKS clusters.

dims:
/approve
/lgtm

( i don't have enough karma @upodroid )

jaypipes:
Seems perfectly reasonable to me!

/lgtm

PR #542, opened 2023-01-17
Migrate away from google.com gcp project k8s-testimages

cpanato:

- Migrate away from google.com gcp project k8s-testimages

Related to https://github.com/kubernetes/k8s.io/issues/1523

/assign @ameukam @micahhausler

Fixes #

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,15 +1,15 @@
 options:
   substitution_option: ALLOW_LOOSE
 steps:
-  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
+  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20221214-1b4dd4d69a
     entrypoint: /buildx-entrypoint
     args:
-    - build
-    - --tag=gcr.io/$PROJECT_ID/aws-iam-authenticator:$_GIT_TAG
-    - --tag=gcr.io/$PROJECT_ID/aws-iam-authenticator:latest
-    - --platform=linux/amd64,linux/arm64
-    - --output=type=registry
-    - .
+      - build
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -1,15 +1,15 @@
 options:
   substitution_option: ALLOW_LOOSE
 steps:
-  - name: gcr.io/k8s-testimages/gcb-docker-gcloud
+  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20211118-2f2d816b90
```
dims, line 4: @cpanato looks like newer ones are available? https://cs.github.com/?scopeName=All+repos&scope=&q=gcr.io%2Fk8s-staging-test-infra%2Fgcb-docker-gcloud%3Av2022+org%3Akubernetes 

can you please update?

cloudbuild.yaml
```diff
@@ -1,15 +1,15 @@
...
     entrypoint: /buildx-entrypoint
     args:
-    - build
-    - --tag=gcr.io/$PROJECT_ID/aws-iam-authenticator:$_GIT_TAG
-    - --tag=gcr.io/$PROJECT_ID/aws-iam-authenticator:latest
-    - --platform=linux/amd64,linux/arm64
-    - --output=type=registry
-    - .
+      - build
```
dims, line 7: Do we need these whitespace changes?

cloudbuild.yaml
cpanato: done, ptal
thanks for the catch

cloudbuild.yaml
cpanato: there no whitespace it is just indentation for better yaml format

PR #547, opened 2023-01-19
Bump dependencies and go version (in go.mod)

dims:
We have new version of k8s, let's switch to the v1.26.x and the dependencies that version uses as well. If it is not in k8s go.mod then pick new versions directly from the dependencies.

- update to v2 for ginkgo
- adapt to the changes in k8s.io test framework

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

Fixes #


nnmin-aws:
apology for your inconvenience @dims. I will check on the failure and get back

tests/integration/server/main_test.go
```diff
@@ -29,10 +30,21 @@ var (
 	testArtifactsDir        string
 )
 
+//go:linkname startEtcd k8s.io/kubernetes/test/integration/framework.startEtcd
+func startEtcd() (func(), error)
```
dims, line 34: `framework.EtcdMain` has a check for leaking go routines which keeps tripping us up. so we should just use the `startEtcd` directly until we want to spend time hunting down those leaks unrelated to our code here in this repository.

NOTE `startEtcd` is not a public method in the `framework` package, so we have to hack around.

tests/integration/server/server_test.go
```diff
@@ -13,11 +13,8 @@ import (
 )
 
 func TestServer(t *testing.T) {
-	stopCh := make(chan struct{})
-	defer close(stopCh)
-
-	adminClient, execClient := testutils.StartAuthenticatorTestFramework(
```
dims, line 19: StartAuthenticatorTestFramework no longer takes a `stopCh` parameter

tests/integration/testutils/testserver.go
```diff
@@ -56,7 +56,7 @@ func StartAuthenticatorTestFramework(t *testing.T, stopCh <-chan struct{}, setup
 		t.Fatal(err)
 	}
 
-	adminClient, kubeAPIServerClientConfig := framework.StartTestServer(t, stopCh, framework.TestServerSetup{
+	adminClient, kubeAPIServerClientConfig, tearDownFn := framework.StartTestServer(t, framework.TestServerSetup{
```
dims, line 59: `TestServerSetup` now returns a tear down function we need to call for cleaning up

tests/integration/testutils/testserver.go
```diff
@@ -114,7 +115,11 @@ func StartAuthenticatorTestFramework(t *testing.T, stopCh <-chan struct{}, setup
 		t.Fatal(err)
 	}
 
-	return adminClient, clientWithExecAuthenticator
+	return adminClient, clientWithExecAuthenticator, func() {
```
dims, line 118: wrap the tear down method returned by `StartTestServer` with a couple of things we need to clean up related to things we started here in this method

pkg/server/server.go
```diff
@@ -159,10 +160,19 @@ func (c *Server) Run(stopCh <-chan struct{}) {
 		http.ListenAndServe(":21363", &healthzHandler{})
 	}()
 	if err := c.httpServer.Serve(c.listener); err != nil {
-		logrus.WithError(err).Fatal("http server exited")
+		logrus.WithError(err).Warning("http server exited")
 	}
 }
 
+func (c *Server) Close() {
```
dims, line 167: adding a new method to try and clean up goroutines that leak.

tests/integration/server/main_test.go
dims: fixing this in https://github.com/kubernetes/kubernetes/pull/115254

tests/integration/testutils/testserver.go
nckturner: Definitely more user-friendly.

tests/integration/server/main_test.go
nckturner: Some/all of this info is worth adding to a code comment.  I would definitely have to dig up this PR if I came across this code without knowing the context.

nnmin-aws: will we change back to framework.EtcdMain when the fixing in kubernetes are in a new release?

pkg/server/server.go
nckturner: This is good, and I think we should call this when the stopCh which is passed to `Run` is closed.  That way `Close` is called on SIGTERM and SIGINT ([[1]](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/ca0331a3b80448bf432aec17bbce7c1e730c4b10/cmd/aws-iam-authenticator/server.go#L54), [[2]](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager/signals#SetupSignalHandler)).

tests/integration/server/main_test.go
dims: `framework.EtcdMain` YESSSS!!

pkg/server/server.go
dims: @nckturner Could we do this in a follow please?

dims: Let's do this in a follow up

PR #554, opened 2023-02-02
Enable Mapping on UserID

nnmin-aws:
1. this PR add UserId lookup in the RoleMapping for dynamic file
2. this PR fix the User mapping failure introduced by https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/55005e174cfca7eb5538b5dedfc409cca9be6d1a#diff-4dfc68b8a5ed36d024d6f6626d8aa66132e3c5702e882182396fbbcb1299cf01R40

bug explaination:

https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/55005e174cfca7eb5538b5dedfc409cca9be6d1a#diff-4dfc68b8a5ed36d024d6f6626d8aa66132e3c5702e882182396fbbcb1299cf01R40 didn't update m.UserARN to canonicalizedARN. When it compares the UserARN, it only lower case the m.UserARN, but didn't call canonicalize either. https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/55005e174cfca7eb5538b5dedfc409cca9be6d1a#diff-7f58544dd05343616eb8d9d780c0989085cb17cbd650375495ce8152bc872642R118

This caused lookup failure because the server always pass in canonicalizedARN. https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/server/server.go#L379
 

Fixes #

### DIFF
--- a/pkg/config/types.go
+++ b/pkg/config/types.go
@@ -57,6 +57,9 @@ type RoleMapping struct {
 	// Groups is a list of Kubernetes groups this role will authenticate
 	// as (e.g., `system:masters`). Each group name can include placeholders.
 	Groups []string `json:"groups" yaml:"groups"`
+
+	// UserId is the AWS PrincipalId of the role. (e.g., "ABCXSOTJDDV").
+	UserId string `json:"userid,omitempty" yaml:"userid,omitempty"`
 }
 
 // UserMapping is a static mapping of a single AWS User ARN to a
@@ -70,6 +73,9 @@ type UserMapping struct {
 
 	// Groups is a list of Kubernetes groups this role will authenticate as (e.g., `system:masters`)
 	Groups []string `json:"groups" yaml:"groups"`
+
+	// UserId is the AWS PrincipalId of the user. (e.g., "ABCXSOTJDDV").
+	UserId string `json:"userid,omitempty" yaml:"userid,omitempty"`
 }
 
 // SSOARNMatcher contains fields used to match Role ARNs that
@@ -171,4 +177,6 @@ type Config struct {
 	EC2DescribeInstancesBurst int
 	//Dynamic File Path for DynamicFile BackendMode
 	DynamicFilePath string
+	//use UserId for mapping, IdentityArn is not used any more when DynamicFileUserIDStrict=true
+	DynamicFileUserIDStrict bool
 }
--- a/pkg/mapper/dynamicfile/dynamicfile.go
+++ b/pkg/mapper/dynamicfile/dynamicfile.go
@@ -4,16 +4,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
-	"strings"
-	"sync"
-	"time"
-
 	"github.com/fsnotify/fsnotify"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"os"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
+	"strings"
+	"sync"
+	"time"
 )
 
 type DynamicFileMapStore struct {
@@ -21,8 +20,9 @@ type DynamicFileMapStore struct {
 	users map[string]config.UserMapping
 	roles map[string]config.RoleMapping
 	// Used as set.
-	awsAccounts map[string]interface{}
-	filename    string
+	awsAccounts  map[string]interface{}
+	filename     string
+	userIDStrict bool
 }
 
 type DynamicFileData struct {
@@ -66,7 +66,7 @@ func (m *DynamicFileMapStore) loadDynamicFile() error {
 	}
 	logrus.Infof("LoadDynamicFile: %v is available. loading", m.filename)
 	// load the initial file content into memory
-	userMappings, roleMappings, awsAccounts, err := ParseMap(m.filename)
+	userMappings, roleMappings, awsAccounts, err := ParseMap(m)
 	if err != nil {
 		logrus.Errorf("LoadDynamicFile: There was an error parsing the dynamic file: %+v. Map is not updated. Please correct dynamic file", err)
 		return err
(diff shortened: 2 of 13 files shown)


pkg/server/server.go
```diff
@@ -378,15 +381,25 @@ func (h *handler) doMapping(identity *token.Identity) (string, []string, error)
...
-			if err != mapper.ErrNotMapped {
-				errs = append(errs, fmt.Errorf("mapper %s Map error: %v", m.Name(), err))
-			}
-
-			if m.IsAccountAllowed(identity.AccountID) {
-				return identity.CanonicalARN, []string{}, nil
+		}
+		if h.userIDInUse {
+			//doMapping once more with UserID
+			mapping, err = m.Map(identity.UserID)
```
nckturner, line 387: Why do you do the ARN mapping AND the principal ID mapping?  Given there is an opt in switch, I'd think its one or the other.  In what case do we want both?

nckturner: I also don't understand how this is expected to work, you are overloading the parameter to Map which is supposed to be an ARN, and passing that to the RoleMapping lookup.  RoleMapping compares the ARN to the stored ARN loaded from the mapping.  I was expecting a new field for principal ID, which we can add to our structs in the Authenticator.

nckturner: What if we pass the entire `identity` to m.Map()? Then we can have logic that treats the ARN and the principal ID separately in the Map function based on whether principal ID mapping is enabled.

nnmin-aws: thanks! I have updated by modifying the interface and passing the entire identity to m.Map

pkg/mapper/dynamicfile/mapper.go
```diff
@@ -29,14 +30,14 @@ func (m *DynamicFileMapper) Start(stopCh <-chan struct{}) error {
 	return nil
 }
 
-func (m *DynamicFileMapper) Map(canonicalARN string) (*config.IdentityMapping, error) {
-	canonicalARN = strings.ToLower(canonicalARN)
+func (m *DynamicFileMapper) Map(identity *token.Identity) (*config.IdentityMapping, error) {
+	canonicalARN := strings.ToLower(identity.CanonicalARN)
 
-	rm, err := m.RoleMapping(canonicalARN)
+	rm, err := m.RoleMapping(identity.UserID)
```
nckturner, line 36: Can we support mapping by UserId and ARN? For example: if only UserId is set, use it.  If only ARN is set, use it.  If both are set, use UserId and compare ARN to ensure a match.

nnmin-aws: dynamic file is new mode and we have decided to use aws UserId because it is global unique. assuming we do not go back to IdentityARN any more since we already know the weakness of IdentityARN for non-exist role and recreated role?

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -4,22 +4,31 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/wait"
+	"os"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
+	"strings"
+	"sync"
+	"time"
 )
 
+type RoleMappingV2 struct {
```
nckturner, line 18: I would expect adding a field to be backwards compatible.  Are there any situations that we need a new struct for?

nnmin-aws: Yep I can add the new field into existing RoleMapping struct. But existing mode user might wonder what this field is and what value shall they put in when they create new entries for existing mode. Hence I introduced a new struct for dynamic file and minimize the impact.

pkg/mapper/dynamicfile/mapper.go
nnmin-aws: the new requirement is to replace the IdentityArn with PrincipleId because IdentityArn has weakness in non-exist role and recreated-role, is it?

nckturner: Yes, but its also a decision that we can leave up to the user, and we can do this in a backwards compatible way for other backend modes.

nckturner:
After discussion, we will limit the flag to affect dynamic file backend mode, with the intention that we can add flags for other backend modes.

pkg/config/types.go
```diff
@@ -70,6 +73,9 @@ type UserMapping struct {
 
 	// Groups is a list of Kubernetes groups this role will authenticate as (e.g., `system:masters`)
 	Groups []string `json:"groups" yaml:"groups"`
+
+	// UserId is the AWS PrincipleId of the user. (e.g., "ABCXSOTJDDV").
```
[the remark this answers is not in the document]
nnmin-aws: updated, good catch. thank you

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -184,12 +185,18 @@ func (ms *DynamicFileMapStore) saveMap(
 	ms.awsAccounts = make(map[string]interface{})
 
 	for _, user := range userMappings {
-		canonicalizedARN, _ := arn.Canonicalize(strings.ToLower(user.UserARN))
-		ms.users[canonicalizedARN] = user
+		key, _ := arn.Canonicalize(strings.ToLower(user.UserARN))
+		if ms.userIDStrict {
+			key = user.UserId
```
wongma7, line 190: what happens if the config has entries without userId set, since it is optional regardless of whether `userIDStrict` is true/false?

should we throw error or warning?

For example, let's say i have 2 entries
- userArn:a
  userId:a-id
  username: a
- userArn:b
  username:b

the second entry will be useless if `userIDStrict` is true. Should user be warned that they have useless entries?

wongma7: i think logging a warning in case a useless entry is detected would be nice. But IDK how often this gets called so if it's annoying then never mind.


the inverse is also an issue but a much smaller one. If they have `userIDStrict` false and they have an entry with userId in it, actually userId will not be respected. So i don't think in this case warning is needed

nnmin-aws: that's a good point. I will add the check for UserId when userIDStrict is true

pkg/config/types.go
```diff
@@ -171,4 +177,6 @@ type Config struct {
 	EC2DescribeInstancesBurst int
 	//Dynamic File Path for DynamicFile BackendMode
 	DynamicFilePath string
+	//use UserId for mapping strictly
```
nckturner, line 180: Define what "strictly" means in the comment.  As in, does not support mapping by ARN.

pkg/mapper/dynamicfile/dynamicfile.go
nckturner: yeah I think its worth logging.  Also we can skip the lookup if its not set?

pkg/config/types.go
nnmin-aws: thank you. I have updated

nckturner:
I think everything is resolved.

/lgtm
/approve

PR #555, opened 2023-02-08
chore: Update dependency golang.org/x/net to fix CVE-2022-41717

tembleking:

The golang.org/x/net dependency in use is affected by https://nvd.nist.gov/vuln/detail/CVE-2022-41717, this updates the dependency to the latest version so this binary is not affected


go.mod
```diff
@@ -16,15 +16,14 @@ require (
 	k8s.io/api v0.26.1
 	k8s.io/apimachinery v0.26.1
 	k8s.io/client-go v0.26.1
-	k8s.io/code-generator v0.26.1
-	k8s.io/component-base v0.26.1
-	k8s.io/sample-controller v0.26.1
+	k8s.io/code-generator v0.23.0-alpha.0
```
dims, line 19: Sorry NO, this is going backwards.

nckturner:
This was supposed to be fixed in https://github.com/kubernetes/kubernetes/commit/afe5378db9d17b1e16ea0028ecfab432475f8e25, which is included in our current dependency version.  Either that was a patched version of 0.3 (vulnerability database said > 0.4), or @liggitt made a mistake and updated to a version that didn't contain the fix.

Considering this dependency version included due to code-generator and sample-controller, I don't think its high risk and we can wait until we get the fix from upstream in the if the version didn't include the fix.

nckturner:
Actually let me look at our usage of sample-controller, that might be an issue.

liggitt:
> This was supposed to be fixed in https://github.com/kubernetes/kubernetes/commit/afe5378db9d17b1e16ea0028ecfab432475f8e25, which is included in our current dependency version. Either that was a patched version of 0.3 (vulnerability database said > 0.4), or @liggitt made a mistake and updated to a version that didn't contain the fix.

It is fixed in that version, https://github.com/golang/net/commit/1e63c2f08a10a150fa02c50ece89b340ae64efe4 is exactly the content/commit tagged as v0.4.0 (https://github.com/golang/net/commits/v0.4.0), that tag just didn't exist at the time we bumped, so go generated a prerelease pseudo-version of "v0.3.1-0.20221206200815-1e63c2f08a10"

nckturner:
Thanks, closing. 

/close

PR #556, opened 2023-02-14
Add default instance region in sts hostname

kmala:
This PR will validate if the token hostname if it matches the instances metadata region even though if the endpoint is not part of the aws-sdk-go static config. This will help in removing the dependency on updating the sdk-go every time a new AWS region comes up. We are doing similar fallback check both in the [aws cloud provider](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v1/aws.go#L1484) and in [k8s  code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1435).

Fixes #

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -27,6 +27,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/ec2provider"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/mapper"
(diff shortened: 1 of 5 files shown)


pkg/server/server.go
```diff
@@ -185,10 +187,16 @@ func (c *Server) getHandler(mappers []mapper.Mapper, ec2DescribeQps int, ec2Desc
 			panic(fmt.Sprintf("describeinstancesrole %s is not a valid arn", c.ServerEC2DescribeInstancesRoleARN))
 		}
 	}
+	sess := session.Must(session.NewSession())
+	ec2metadata := ec2metadata.New(sess)
+	instanceRegion, err := ec2metadata.Region()
+	if err != nil {
```
nckturner, line 193: Since this falls through on error, is passing empty/nil `instanceRegion` to `token.NewVerifier()` and `ec2provider.New()` safe?

kmala: I am trying to keep the current behavior as it is. EC2 provider currently updates the region from metadata on a best effort. Same goes for NewVerifier where it skips if it finds any error when trying to get the region sts endpoint.

PR #593, opened 2023-04-21
fix semantic error

dominik-ba:
If I'm not mistaken the comment should state that a IAM User is mapped to a Kubernetes user at the changed spot. roles are described in the lines before that part

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,7 +482,7 @@ server:
     groups:
     - system:masters
 
-  # each mapUsers entry maps an IAM role to a static username and set of groups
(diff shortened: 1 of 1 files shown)


dominik-ba:
@nckturner since I only changed that one comment I highly doubt the failing check is related to my commit. Can I get some guidance what to do here?

dominik-ba:
/remove-lifecycle stale

Honestly I don't get why a one word readme change PR can get stale... why is no one looking at this?

dims:
@dominik-ba thank you and apologies. since this is not a code issue, i will merge without waiting for the job to pass for now.

PR #595, opened 2023-04-26
Enable cross-compilation in Dockerfile

chewong:

Enable cross-compilation in Dockerfile. We need to add `ARG TARGETOS TARGETARCH` for them to be populated by buildx and we need to add `--platform=$TARGETPLATFORM` in `FROM` statements to specify which platform we use to pull those images

Fixes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,12 +18,13 @@ FROM --platform=$BUILDPLATFORM $golang_image AS builder
 WORKDIR /go/src/github.com/kubernetes-sigs/aws-iam-authenticator
 COPY . .
 RUN go mod download
+ARG TARGETOS TARGETARCH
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,12 +18,13 @@ FROM --platform=$BUILDPLATFORM $golang_image AS builder
 WORKDIR /go/src/github.com/kubernetes-sigs/aws-iam-authenticator
 COPY . .
 RUN go mod download
+ARG TARGETOS TARGETARCH
```
nckturner, line 21: Can we get a makefile target similar to [build-all-bins](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/Makefile#L81) for building a set of supported targets?

chewong: Done

PR #601, opened 2023-05-19
reduce token expiration if credentials will expire

### LINKED ISSUE #590: API Authorization from Outside EKS Cluster throws `Unauthorized` error after 20+ minutes
We have tried to implement similar methods defined in section https://github.com/kubernetes-sigs/aws-iam-authenticator#api-authorization-from-outside-a-cluster in Golang

```go
func GetEKSToken(ctx context, clusterName string) (*KubeToken, error) {

	request, _ := NewStsClientFrom(ctx).GetCallerIdentityRequest(&sts.GetCallerIdentityInput{})
	# clusterName in the request header
        request.HTTPRequest.Header.Add("x-k8s-aws-id", clusterName)

	presignUrl, err := request.Presign(60) # presign 60 seconds
	if err != nil {
            .....
	}

	return &KubeToken{
		Kind:       "ExecCredential",
		ApiVersion: "client.authentication.k8s.io/v1beta1",
		Status: &KubeTokenStatus{
			ExpirationTimestamp: time.Now().Local().Add(time.Hour * time.Duration(1)).Format("2006-01-02T15:04:05Z"),
			Token:               "k8s-aws-v1." + base64.RawURLEncoding.EncodeToString([]byte(presignUrl)),
		},
	}, nil
}
```

It works fine.

However, we are having intermittent issue that the Kubernetes Client created with this token is throwing `Unauthorized` error when performing kubernetes operation, for example, equivalent to `kubectl get nodes`

We are using https://kubernetes.io/docs/reference/config-api/client-authentication.v1beta1/#client-authentication-k8s-io-v1beta1-ExecCredential instead of using similar `headers = {'Authorization': 'Bearer ' + get_bearer_token('my_cluster', 'us-east-1')}` as the tutorial.

I am wondering what could be the potential reason for this `Unauthorized` error ? The API successfully ran for almost 20 minutes and then suddenly this error is thrown back.

I am thinking:

1. Is the `bearerToken` expired exactly at the timestamp defined in `ExpirationTimestamp` or after some magical time delta ? Currently we configured the `ExpirationTimestamp` to be 1 hour after the token is generated. Does this conflict with the sts presign 60 seconds ?
2. I noticed that `Something to note though is that the IAM Authenticator explicitly omits base64 padding to avoid any = characters thus guaranteeing a string safe to use in URLs.` is mentioned and the python code example

```
    # remove any base64 encoding padding:
```

is explicitly replace `=` with ` ` empty string, which is absent from our golang methods, how so far everything related to kubernetes operation is working fine.

Also trying to get some feedback on this if there is anything else that I am missing.

iamnoah:

An issue comes up when using a long-lived `token.Generator` instance where the underlying assume-role session expires,  which invalidates the token. This only changes the expiration date so e.g., `oauth2.ReuseTokenSource` will know to regenerate the token when the session expires.

**Which issue(s) this PR fixes** **:

fixes #590

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -41,6 +41,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/pkg/apis/clientauthentication"
 	clientauthv1beta1 "k8s.io/client-go/pkg/apis/clientauthentication/v1beta1"
+
 	"sigs.k8s.io/aws-iam-authenticator/pkg"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/metrics"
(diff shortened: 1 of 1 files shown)


iamnoah:
👋🏼 
still here. Anything this PR needs to be considered?

PR #604, opened 2023-06-06
Use release key

nckturner:

Github by default does not allow a workflow to be triggered from a workflow.  However, we would like our release workflow to be triggered from our tag workflow.  In order to workaround this limitation, we create a dedicated encrypted secret containing a deploy key called `RELEASE_KEY`.  

/cc @nnmin-aws

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -15,6 +15,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ssh-key: "${{ secrets.RELEASE_KEY }}"
       -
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release.yml
```diff
@@ -15,6 +15,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          ssh-key: "${{ secrets.RELEASE_KEY }}"
```
nnmin-aws, line 18: where is secrets.RELEASE_KEY defined? forgive me if it is a silly q

PR #607, opened 2023-06-20
Keep original aws-auth configuration when the new value cannot be parsed

suzaku:
In the current implementation, a value like `rokeMappings` is always
reinitialized to an empty slice for any `Added` or `Modified` event:

`roleMappings = make([]config.RoleMapping, 0)`

And then, even when there's parse errors detected, it's passed to
`saveMap` with other probably valid data, and inside `saveMap` the field
`ms.roles` is, again, always reinitialized to an empty slice.

## It took only one typo in `aws-auth` to break a whole EKS cluster

We have introduced an error in the `mapRoles` configuration recently
like the following:

`time="2023-06-13T09:35:43Z" level=warning **msg="Errors parsing
configmap: [yaml: line 13: did not find expected key]"**`

And this change alone broke the whole EKS cluster because, as explained
above, `ms.roles` is always reinitialized to an empty slice and now even
the nodes in cluster get `access denied` because of the `ARN is not
mapped` error.

*To make EKS clusters less fragile*, we should keep the original
configuration if the new one is malformed.

### DIFF
--- a/pkg/mapper/configmap/configmap.go
+++ b/pkg/mapper/configmap/configmap.go
@@ -110,60 +110,74 @@ func (err ErrParsingMap) Error() string {
 	return fmt.Sprintf("error parsing config map: %v", err.errors)
 }
 
+// ParseMap parses a ConfigMap for mapUsers, mapRoles and mapAccounts.
+// If a configuration is not correctly encoded in YAML, nil will be returned for the corresponding value.
 func ParseMap(m map[string]string) (userMappings []config.UserMapping, roleMappings []config.RoleMapping, awsAccounts []string, err error) {
 	errs := make([]error, 0)
 	rawUserMappings := make([]config.UserMapping, 0)
 	userMappings = make([]config.UserMapping, 0)
+	var isMapUsersMalformed bool
 	if userData, ok := m["mapUsers"]; ok {
-		userJson, err := utilyaml.ToJSON([]byte(userData))
-		if err != nil {
+		if userJson, err := utilyaml.ToJSON([]byte(userData)); err != nil {
 			errs = append(errs, err)
+			isMapUsersMalformed = true
 		} else {
-			err = json.Unmarshal(userJson, &rawUserMappings)
-			if err != nil {
+			if err := json.Unmarshal(userJson, &rawUserMappings); err != nil {
 				errs = append(errs, err)
-			}
-
-			for _, userMapping := range rawUserMappings {
-				err = userMapping.Validate()
-				if err != nil {
-					errs = append(errs, err)
-				} else {
-					userMappings = append(userMappings, userMapping)
+				isMapUsersMalformed = true
+			} else {
+				for _, userMapping := range rawUserMappings {
+					err = userMapping.Validate()
+					if err != nil {
+						errs = append(errs, err)
+					} else {
+						userMappings = append(userMappings, userMapping)
+					}
 				}
 			}
 		}
+	}
+	if isMapUsersMalformed {
+		userMappings = nil
 	}
 
 	rawRoleMappings := make([]config.RoleMapping, 0)
 	roleMappings = make([]config.RoleMapping, 0)
+	var isMapRolesMalformed bool
 	if roleData, ok := m["mapRoles"]; ok {
-		roleJson, err := utilyaml.ToJSON([]byte(roleData))
-		if err != nil {
+		if roleJson, err := utilyaml.ToJSON([]byte(roleData)); err != nil {
 			errs = append(errs, err)
+			isMapRolesMalformed = true
 		} else {
-			err = json.Unmarshal(roleJson, &rawRoleMappings)
-			if err != nil {
+			if err := json.Unmarshal(roleJson, &rawRoleMappings); err != nil {
 				errs = append(errs, err)
-			}
-
-			for _, roleMapping := range rawRoleMappings {
-				err = roleMapping.Validate()
-				if err != nil {
-					errs = append(errs, err)
-				} else {
-					roleMappings = append(roleMappings, roleMapping)
+				isMapRolesMalformed = true
+			} else {
+				for _, roleMapping := range rawRoleMappings {
+					err = roleMapping.Validate()
+					if err != nil {
+						errs = append(errs, err)
+					} else {
+						roleMappings = append(roleMappings, roleMapping)
+					}
 				}
 			}
 		}
 	}
+	if isMapRolesMalformed {
+		roleMappings = nil
+	}
 
 	awsAccounts = make([]string, 0)
+	var isMapAccountsMalformed bool
 	if accountsData, ok := m["mapAccounts"]; ok {
-		err := yaml.Unmarshal([]byte(accountsData), &awsAccounts)
-		if err != nil {
+		if err := yaml.Unmarshal([]byte(accountsData), &awsAccounts); err != nil {
 			errs = append(errs, err)
+			isMapAccountsMalformed = true
 		}
+	}
+	if isMapAccountsMalformed {
+		awsAccounts = nil
 	}
 
 	if len(errs) > 0 {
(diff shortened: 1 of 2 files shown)


suzaku:
@nckturner PTAL

jyotimahapatra:
From a functionality perspective, how useful is it to allow authenticator to keep running on stale value? When the authenticator process/pod restarts, the new authenticator cannot really parse the values and eventually fail. If there is a problem, authenticator should rather fail immediately. 
Can you describe a bit about how the change handles authenticator restarts?

suzaku:
> From a functionality perspective, how useful is it to allow authenticator to keep running on stale value? When the authenticator process/pod restarts, the new authenticator cannot really parse the values and eventually fail. If there is a problem, authenticator should rather fail immediately. Can you describe a bit about how the change handles authenticator restarts?

If an essential component like `authenticator` fails as I described, you can't even use `kubectl` because all access will be denied.
For example, someone might accidentally introduce extra whitespace when adding a new role, I think it's much safer to keep the original value instead of suddenly disconnecting all kubelets with the API server by clearing the existing values.

There are cases where we should fail immediately, for example, when writing a command line program, you can validate the user input and fail immediately with some meaningful message, but `authenticator` configuration update is not such a scenario IMHO. 

I agree that on restarts, the version with my change would still read the incorrect configuration and the `roles` would be empty because there's no `original value` to keep, but it's the same case with the current implementation, right? 

> 
> Also, can you specify which EKS version is being used?

It's `eks.7` with `Kubernetes 1.24`.

jyotimahapatra:
> suddenly disconnecting all kubelets with the API server by clearing the existing values

I reason with this but this could go both ways. We can choose to let the older values still work, but then there is no indication to the user that a mistake has been made. Finding the error abruptly later is only delaying it.

There may be other usecases related to security that may have to be evaluated before letting stale values be allowed. EKS is trying out a webhook to disallow such erroneous updates to the configmap. 
Could I request you to provide the cluster arn here or in a support ticket where you are able to still update an erroneous change to the configmap? The EKS team will look into the reason why the webhook is not working as expected.

suzaku:
> > suddenly disconnecting all kubelets with the API server by clearing the existing values
> 
> I reason with this but this could go both ways. We can choose to let the older values still work, but then there is no indication to the user that a mistake has been made. Finding the error abruptly later is only delaying it.
> 
> There may be other usecases related to security that may have to be evaluated before letting stale values be allowed. EKS is trying out a webhook to disallow such erroneous updates to the configmap. Could I request you to provide the cluster arn here or in a support ticket where you are able to still update an erroneous change to the configmap? The EKS team will look into the reason why the webhook is not working as expected.

Since `ConfigMap` itself doesn't prevent such updates, it would be great if EKS can provide such validations and prevent dirty data in the first place.
But it seems to not be working correctly, after understanding the root cause of the incident, one of our SRE engineers tried to create a new EKS cluster and updated`aws-auth` to include an invalid `map-roles` and he managed to reproduce the issue.
Can you try creating a new EKS cluster and see if you can successfully update the `map-roles` section of `aws-auth` to include invalid `YAML` data?

ivelichkovich:
New clusters should have validation in place that prevents this. Can you please open a support ticket and on the support ticket share specific details for how to reproduce and information on which clusters you were able to reproduce on so we can investigate further.

jyotimahapatra:
> Can you try creating a new EKS cluster and see if you can successfully update the map-roles section of aws-auth to include invalid YAML data?

Can i have a sample yaml that fails. The webhook just uses [this](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/mapper/configmap/configmap.go#L113) function from authenticator to validate. Its not very advanced because of backward compat considerations. We're handling the most common errors we see.

suzaku:
> > Can you try creating a new EKS cluster and see if you can successfully update the map-roles section of aws-auth to include invalid YAML data?
> 
> Can i have a sample yaml that fails. The webhook just uses [this](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/mapper/configmap/configmap.go#L113) function from authenticator to validate. Its not very advanced because of backward compat considerations. We're handling the most common errors we see, which will not break live clusters.

Cool, I think calling `ParseMap` is good enough for detecting malformed YAML contents.
I guess my colleague used `eks.7` for the testing cluster, do you know in which version of EKS is the webhook introduced?

jyotimahapatra:
> do you know in which version of EKS is the webhook introduced?

The webhook is added for  new clusters. We'll make the changes next week such that its added for existing clusters too. I did check by creating a new cluster that the webhook will help prevent the error condition you're facing.

suzaku:
> > do you know in which version of EKS is the webhook introduced?
> 
> The webhook is added for new clusters. We'll make the changes next week such that its added for existing clusters too. I did check by creating a new cluster that the webhook will help prevent the error condition you're facing.

Do you mean clusters with latest version or any newly created EKS?

jyotimahapatra:
> Do you mean clusters with latest version or any newly created EKS?

Should be available for any newly created EKS 1.22+ cluster. (I tested on 1.26 cluster). Let me know how it goes

mohanpedala:
@jyotimahapatra I have a question even though the clusters are old, we keep updating them with the latest aws eks versions; at what point does this webhook being bootstrapped with the new clusters?

As per your comments, new clusters will have the webhook in place; we replicated the same issue in Version 1.24 in a new ephemeral cluster.

We are happy to discuss more on this issue.

ivelichkovich:
> @jyotimahapatra I have a question even though the clusters are old, we keep updating them with the latest aws eks versions; at what point does this webhook being bootstrapped with the new clusters?
> 
> As per your comments, new clusters will have the webhook in place; we replicated the same issue in Version 1.24 in a new ephemeral cluster.
> 
> We are happy to discuss more on this issue.

Can you please share the steps to reproduce this? What exact changes are you making to the configmap?

mohanpedala:
> > @jyotimahapatra I have a question even though the clusters are old, we keep updating them with the latest aws eks versions; at what point does this webhook being bootstrapped with the new clusters?
> > As per your comments, new clusters will have the webhook in place; we replicated the same issue in Version 1.24 in a new ephemeral cluster.
> > We are happy to discuss more on this issue.
> 
> Can you please share the steps to reproduce this? What exact changes are you making to the configmap?

* When we are adding a role to the configmap we accidentally delete a letter in the key and saved the config

### To replicate the issue, here are the steps.
* If you are spinning a new cluster, you have to remove the validatingwebhookconfiguration eks-aws-auth-configmap-validation-webhook

* If you are running on a cluster older than 3 to 4 weeks from the date of this comment. You most probably won't see the eks-aws-auth-configmap-validation-webhook

* Edit aws-auth configmap - See the below example ConfigMap the error is present in line 13 key "rolearn"
```yaml
apiVersion: v1
data:
  mapAccounts: |
    []
  mapRoles: |
    - "groups":
      - "system:bootstrappers"
      - "system:nodes"
      "olearn": "arn:aws:iam::xxxx:role/node-role"
      "username": "system:node:{{EC2PrivateDNSName}}"
    - "groups":
      - "system:masters"
      "rolearn": "arn:aws:iam::xxxxx:role/AWSReservedSSO_AdministratorAccess_1234567890"
      "username": "aws_sso_admin"
  mapUsers: |
    []
immutable: false
kind: ConfigMap
metadata:
  creationTimestamp: "2023-07-07T15:50:37Z"
  labels:
    app.kubernetes.io/managed-by: Terraform
    terraform.io/module: terraform-aws-modules.eks.aws
  name: aws-auth
  namespace: kube-system
```

mohanpedala:
> @jyotimahapatra I have a question even though the clusters are old, we keep updating them with the latest aws eks versions; at what point does this webhook being bootstrapped with the new clusters?
> 
> As per your comments, new clusters will have the webhook in place; we replicated the same issue in Version 1.24 in a new ephemeral cluster.
> 
> We are happy to discuss more on this issue.

@jyotimahapatra The cluster I created was a week before your change was introduced, I can now see the webhook in all newly created clusters.

PR #608, opened 2023-06-29
Allow running create release from github UI

dims:
We have a lot of tags with no corresponding release. So trying to figure out how to get them back.


dims:
Doing GH action surgery/diagnosis, so merging by hand

PR #612, opened 2023-07-17
add Dynamic Backend Mode support

nnmin-aws:
Authenticator will dynamically load backend mode and update mapper.
 
Fixes #

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -270,6 +270,11 @@ func BuildMapperChain(cfg config.Config, modes []string) (BackendMapper, error) 
 		if err := m.Start(backendMapper.mapperStopCh); err != nil {
 			logrus.Fatalf("start mapper %q failed", m.Name())
 		}
+		if backendMapper.currentModes != "" {
+			backendMapper.currentModes = backendMapper.currentModes + " " + m.Name()
+		} else {
+			backendMapper.currentModes = m.Name()
+		}
 	}
 	return backendMapper, nil
 }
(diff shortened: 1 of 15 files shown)


pkg/server/server.go
```diff
@@ -80,18 +84,11 @@ func New(cfg config.Config, stopCh <-chan struct{}) *Server {
 		Config: cfg,
 	}
 
-	mappers, err := BuildMapperChain(cfg)
+	backendMapper, err := BuildMapperChain(cfg, cfg.BackendMode)
```
wongma7, line 87: I think if DynamicBackendModePath is set then server startup should block until the server loads backend modes from the file for the first time instead of using cfg.BackendMode

Right now there is a brief period in which the server uses cfg.BackendMode until it loads backend modes from DynamicBackendModePath. I would prefer if I set cfg.DynamicBackendModePath that the server use cfg.DynamicBackendModePath exclusively, and ignore cfg.BackendMode, to avoid any potential race/misconfiguration.

For example let's say I leave BackendMode = MountedFile and assume that the server uses DynamicBackendModePath=DynamicFile instead. But if somebody accidentally deletes DynamicBackendModePath then the next time the server starts it will use BackendMode=MountedFile when that is not my intention.

nnmin-aws: Thank you for the comments! It is designed this way by the moment. cfg provides the default backend mode, and the dynamic file might not be available during startup.

pkg/server/server.go
```diff
@@ -457,3 +484,32 @@ func (h *handler) renderTemplate(template string, identity *token.Identity) (str
...
+
+func (h *handler) CallBackForFileLoad(dynamicContent []byte) error {
+	var backendModes BackendModeConfig
+	logrus.Infof("BackendMode dynamic file got changed to %s", string(dynamicContent))
+	err := json.Unmarshal(dynamicContent, &backendModes)
+	if err != nil {
+		logrus.Infof("CallBackForFileLoad: could not unmarshal dynamic file.")
+		return err
+	}
+	if h.backendMapper.currentModes != backendModes.BackendMode {
```
wongma7, line 496: this seems too sensitive to typos/whitespace in the config, what if it's "MountedFile DynamicFile" vs "MountedFile  DynamicFile"? since the file is json, what about json array? Or comma-separated string to be consistent with cfg.BackendMode

pkg/server/server.go
wongma7: ok thx. Please add this to the comment for `DynamicBackendModePath`. 

The other edge case is if DynamicBackendModePath gets deleted then the server just keeps using the last DynamicBackendModePath it read until it restarts since CallBackForFileDeletion is not implemented. Should it fallback to cfg.BackendMode?

nnmin-aws: this is a good point. updated to fallback to cfg.BackendMode

pkg/server/server.go
nnmin-aws: thank you for the comments! if there is typo in the dynamic backend json file, user can modify it quickly? I will add it in the release note for the format. does it sound good to you?

PR #618, opened 2023-08-04
Avoid parsing known empty inputs

dims:
We are seeing instances of the configmap in the wild with variations like below:

```
apiVersion: v1
data:
mapAccounts: |
    ''
mapUsers: |
    ""
```

We should tolerate these as the user literally means, there's no data they need in there.

Fixes #

### DIFF
--- a/pkg/mapper/configmap/configmap.go
+++ b/pkg/mapper/configmap/configmap.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


ivelichkovich:
We should backport this into v0.5.X

divyasri537:
The issue we encountered from HSBC aws-auth configmap looked like below. Can we cover this case? 
data:
 mapAccounts: |
        ''
    mapUsers: |
        ''
 The webhook does already allow request this way.
 data:
    mapAccounts: ""
    mapUsers: ""

PR #619, opened 2023-08-04
Avoid parsing known empty inputs

DanielCKennedy:
Cherry-pick of https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/618/ to add to 0.5 release

We are seeing instances of the configmap in the wild with variations like below:

```
apiVersion: v1
data:
mapAccounts: |
    ''
mapUsers: |
    ""
```
We should tolerate these as the user literally means, there's no data they need in there.

Fixes #

### DIFF
--- a/pkg/mapper/configmap/configmap.go
+++ b/pkg/mapper/configmap/configmap.go
@@ -3,7 +3,6 @@ package configmap
 import (
 	"context"
 	"errors"
-	"strings"
 	"sync"
 
 	"fmt"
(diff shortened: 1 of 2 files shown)


pkg/mapper/configmap/configmap.go
```diff
@@ -146,6 +152,11 @@ func ParseMap(m map[string]string) (userMappings []config.UserMapping, roleMappi
 	return userMappings, roleMappings, awsAccounts, err
 }
 
+func isSkippable(data string) bool {
+	trimmed := strings.TrimSpace(data)
+	return trimmed == "" || trimmed == "``" || trimmed == "\"\""
```
nckturner, line 157: What about single quotes?  Your example has one with single quotes in the PR description.

PR #629, opened 2023-09-15
Switch to new community owned account

dims:
related to https://github.com/kubernetes/k8s.io/issues/5194


dims:
got a green run https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-iam-authenticator/629/pull-aws-iam-authenticator-integration/1702786074834636800

PR #630, opened 2023-09-18
return 429 for STS throttling

nnmin-aws:
currently authenticator only returns 401 and it is hard to know if the request is really unauthorized or it is due to aws sts throttle.
This PR will create a new metric in authenticator to record the STS throttling count. 
This PR will return 429 to api-server when authenticator got STS throttling from AWS STS service

Fixes #

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -139,6 +139,20 @@ func NewSTSError(m string) STSError {
 	return STSError{message: m}
 }
 
+// STSThrottling is returned when there was STS Throttling.
+type STSThrottling struct {
+	message string
+}
+
+func (e STSThrottling) Error() string {
+	return "sts getCallerIdentity was throttled: " + e.message
+}
+
+// NewSTSError creates a error of type STS.
+func NewSTSThrottling(m string) STSThrottling {
+	return STSThrottling{message: m}
+}
+
 var parameterWhitelist = map[string]bool{
 	"action":               true,
 	"version":              true,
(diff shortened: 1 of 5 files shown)


pkg/token/token.go
```diff
@@ -139,6 +139,14 @@ func NewSTSError(m string) STSError {
 	return STSError{message: m}
 }
 
+// STSThrottling is returned when there was STS Throttling.
+type STSThrottling struct {
+}
+
+func (e STSThrottling) Error() string {
+	return "STSThrottling"
```
nckturner, line 147:
```suggestion
	return "sts getCallerIdentity was throttled: " + e.message
```

pkg/token/token.go
```diff
@@ -570,7 +578,11 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 
 	metrics.Get().StsResponses.WithLabelValues(fmt.Sprint(response.StatusCode)).Inc()
 	if response.StatusCode != 200 {
-		return nil, NewSTSError(fmt.Sprintf("error from AWS (expected 200, got %d). Body: %s", response.StatusCode, string(responseBody[:])))
+		responseStr := string(responseBody[:])
+		if strings.Contains(responseStr, "Throttling") {
```
nckturner, line 582: I assume you are referencing https://docs.aws.amazon.com/STS/latest/APIReference/CommonErrors.html, can you add a link to that in a comment?

mmerkes:
Can you add more context to the PR description? Would be good for people to understand why we're making this change.

mmerkes:
What testing have you done?

pkg/token/token.go
mmerkes: Are `getCallerIdentity` the only calls that will go down this path?

pkg/token/token.go
```diff
@@ -570,7 +578,11 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
...
+			return nil, STSThrottling{}
```
mmerkes, line 583: Should we need to add a message here for this?

nckturner: Yes, I think we need to pass through the error here.

pkg/token/token.go
nnmin-aws: yes, it is

pkg/token/token.go
nnmin-aws: that's a good point. thank you! I will add a link to it in comment

nnmin-aws:
> What testing have you done?

The unit test has been included in the PR, server_test.go and token_test.go

pkg/token/token.go
nnmin-aws: thank you! updated

pkg/token/token.go
```diff
@@ -139,6 +139,20 @@ func NewSTSError(m string) STSError {
 	return STSError{message: m}
 }
 
+// STSThrottling is returned when there was STS Throttling.
+type STSThrottling struct {
+	message string
+}
+
+func (e STSThrottling) Error() string {
+	return "sts getCallerIdentity was throttled:" + e.message
```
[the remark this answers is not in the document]
nnmin-aws: updated. thank you

mmerkes (approved):
LGTM

nnmin-aws:
retest

PR #632, opened 2023-09-27
Optionally return full role arn including path during token verification

### LINKED ISSUE #153: IAM roles with paths are only recognized without the path
Assuming I have the following role ARN:

```
arn:aws:iam::1234567890:role/iam-ss/some-path/actual-role-name
```

If I enter this under `mapRoles`, this will not be recognized. Instead I need to enter:

```
arn:aws:iam::1234567890:role/actual-role-name
```

### LINKED ISSUE #268: Roles with paths do not work when the path is included in their ARN in the aws-auth configmap
I have a role with an ARN that looks like this: `arn:aws:iam::XXXXXXXXXXXX:role/gitlab-ci/gitlab-runner`. My aws-auth configmap was as follow:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: aws-auth
  namespace: kube-system
data:
  mapRoles: |
    - rolearn: arn:aws:iam::XXXXXXXXXXXX:role/EKSWorkerNode
      username: system:node:{{EC2PrivateDNSName}}
      groups:
        - system:bootstrappers
        - system:nodes
    - rolearn: arn:aws:iam::XXXXXXXXXXXX:role/EKSServiceWorker
      username: kubernetes-admin
      groups:
        - system:masters
    - rolearn: arn:aws:iam::XXXXXXXXXXXX:role/gitlab-ci/gitlab-runner
      username: gitlab-admin
      groups:
        - system:masters
```

I repeated got unauthorized errors from the cluster until I updated the `rolearn` to `arn:aws:iam::XXXXXXXXXXXX:role/gitlab-runner`. After that change my access worked as expected.

If it makes a difference, I'm using assume-role on our gitlab-runner, and using `aws eks update-kubeconfig --region=us-east-1 --name=my-cluster` to get kubectl configured.

terrycain:
**What this PR does / why we need it**: When creating EKS nodegroups where the node's IAM role contains a path, the nodes fail to join the cluster. This adds the ability to look up a role's full ARN including the path and return that when verifying a token.

The reason this is needed is because the assume-role ARN that is returned by `sts.get-caller-idendentity` does not include the IAM role's path.

I've added a "role cache" which calles `iam.list-roles` and stores a mapping between the roles unique role ID and the full ARN, as the `UserID` returned by `sts.get-caller-identity` contains the roles unique ID therefore looking up the existing role's correct ARN is possible.

This does require the clusters IAM role to have the `iam:ListRoles` action. If the role cache gets an error when calling list roles and its not a transient error, it'll disable all future lookups and the functionality will revert to the original behaviour. I figured it would be better to make this more an opt in rather than a requirement as else it might just break nodes joining a cluster whenever its released.

**Which issue(s) this PR fixes**:
Fixes #268
Fixes #153

### DIFF
--- a/pkg/token/rolecache.go
+++ b/pkg/token/rolecache.go
@@ -1,0 +1,103 @@
+package token
+
+import (
+	"errors"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/sirupsen/logrus"
+	"sync"
+	"time"
+)
+
+type IAMClient interface {
+	ListRolesPages(input *iam.ListRolesInput, fn func(*iam.ListRolesOutput, bool) bool) error
+}
+
+// RoleCache Cache on first use IAM role cache which stores a map of Role ID -> ARN for lookup during verify
+// operations. This solves the issue of assume-role ARNs not including the assumed role's path.
+type RoleCache struct {
+	awsClient   IAMClient
+	searchRoles bool
+	lastUpdate  time.Time
+	idToFullARN map[string]string
+	mutex       sync.RWMutex
+}
+
+// NewRoleCache Creates a RoleCache and returns it.
+func NewRoleCache() *RoleCache {
+	sess, err := session.NewSessionWithOptions(session.Options{SharedConfigState: session.SharedConfigEnable})
+	var iamClient *iam.IAM
+	if err != nil {
+		logrus.WithError(err).Warn("failed to instantiate AWS session, disabling full role ARN lookup")
+	} else {
+		iamClient = iam.New(sess)
+	}
+
+	return &RoleCache{
+		awsClient:   iamClient,
+		searchRoles: iamClient != nil,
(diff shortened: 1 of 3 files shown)


terrycain:
@dims / @nckturner / @nnmin-aws can we get a review on this please?

terrycain:
Thanks @nnmin-aws the tests seem to have passed, what now?

terrycain:
@nnmin-aws @dims @nckturner what else is required for this PR? There's been no progress for nearly 2 months, some guidance would be appreciated.

dims:
cc @cartermckinnon

nnmin-aws:
@terrycain @dims apology for the delay! i am so sorry about it. We were hesitating at the change because it requires additional invocation of aws IAM and might increase latency of authN.

There is another PR actually merged https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/670/files#diff-5de38ab8d77d935766314beeb9a9e3b7165eb470e656dbb5e4fdf9bb60ae59de. which removed the path from the arn comparison.

Could you please kindly pull latest release and verify if it works for your scenario? If it doesn't, we can check this PR again.

Thank you!

PR #633, opened 2023-10-14
Stricter dependency/security review / Do not use AWS internal image(s)

dims:
- Add new GH action for dependency review
- Add new GH action for govulncheck
- Use community distroless image
- It's ok to use the golang image proxied via public ecr


nckturner:
What's the difference between the eks-distro and the community image?

PR #644, opened 2023-11-17
Fix federated user ID parsing

nckturner:
* refactor the parsing logic to make it easier

This PR changes the way the federated user ID is treated.  Instead of splitting at a colon like we do for assumed-roles, we take the whole value.  This is because if we split at the colon, the user ID is just the account number and can't be differentiated from the root user.

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -600,29 +600,42 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 		return nil, NewSTSError(err.Error())
 	}
 
-	// parse the response into an Identity
 	id := &Identity{
-		ARN:         callerIdentity.GetCallerIdentityResponse.GetCallerIdentityResult.Arn,
-		AccountID:   callerIdentity.GetCallerIdentityResponse.GetCallerIdentityResult.Account,
 		AccessKeyID: accessKeyID,
 	}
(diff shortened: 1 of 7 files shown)


pkg/token/token.go
```diff
@@ -600,29 +600,42 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
...
-	userIDParts := strings.Split(callerIdentity.GetCallerIdentityResponse.GetCallerIdentityResult.UserID, ":")
-	if len(userIDParts) == 2 {
-		id.UserID = userIDParts[0]
-		id.SessionName = userIDParts[1]
-	} else if len(userIDParts) == 1 {
-		id.UserID = userIDParts[0]
+	// The user ID is one of:
+	// 1. UserID:SessionName (for assumed roles)
+	// 2. UserID (for IAM User principals).
+	// 3. AWSAccount:CallerSpecifiedName (for federated users)
```
SaranBalaji90, line 625: How reliable it is, if this is a caller specified custom name?

nckturner: Similar to assumed role session name, although this matches the username I think.  But in the case of session name it depends how its used, if you control how a role is assumed and give the credentials to someone else then you can control the session name.

dims (approved):
LGTM

PR #638, opened 2023-11-09
Bump go to 1.21.4

micahhausler:
**What this PR does / why we need it**:

Go version bump, fixes govulncheck test failure

PR #646, opened 2023-11-21
Go vunlncheck fix

nckturner:
* Replaces https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/638
* Go version bump, fixes govulncheck test failure


cmd/aws-iam-authenticator/server.go
```diff
@@ -52,15 +53,16 @@ var serverCmd = &cobra.Command{
 		var err error
 		fmt.Printf("Authenticator Version: %q, %q\n", pkg.Version, pkg.CommitID)
 		metrics.InitMetrics(prometheus.DefaultRegisterer)
-		stopCh := signals.SetupSignalHandler()
+		cmd.Context()
```
nckturner, line 56: @micahhausler what are we doing with this?  https://github.com/spf13/cobra/blob/main/command.go#L257-L266

pkg/mapper/configmap/configmap.go
```diff
@@ -50,16 +50,16 @@ func New(masterURL, kubeConfig string) (*MapStore, error) {
...
 		for {
 			select {
-			case <-stopCh:
-				logrus.Info("stopCh is closed in startLoadConfigMap")
+			case <-ctx.Done():
+				logrus.Info("Context is complete in startLoadConfigMap")
 				return
 			default:
-				watcher, err := ms.configMap.Watch(context.TODO(), metav1.ListOptions{
+				watcher, err := ms.configMap.Watch(context.Background(), metav1.ListOptions{
```
nckturner, line 62: @micahhausler originally you had this set to the same context that was passed in to startLoadConfigMap(), but this was causing the context to be closed when the watch call failed.  I'm not sure what the right thing to do here is though.

cmd/aws-iam-authenticator/server.go
micahhausler: I think we can drop it

pkg/mapper/configmap/configmap.go
micahhausler: ah ok good catch

nckturner:
Need this PR for integration tests: https://github.com/kubernetes/test-infra/pull/31295

PR #649, opened 2023-11-28
Remove govulncheck on test deps temporarily

nckturner:
We have failing test dependecies and we need to remove these checks until we can fix them.


dims (approved):
LGTM

PR #660, opened 2023-12-07
Check ARN for user ID strict check

nckturner:

Check ARN when mapping based on User ID.  This is necessary because IAM User names/ARNs can change, and without this check, it would be difficult to determine which IAM users currently have access to a cluster.

### DIFF
--- a/pkg/mapper/dynamicfile/mapper.go
+++ b/pkg/mapper/dynamicfile/mapper.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
+	"sigs.k8s.io/aws-iam-authenticator/pkg/errutil"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/fileutil"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/mapper"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/token"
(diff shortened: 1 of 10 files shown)


pkg/mapper/dynamicfile/mapper.go
```diff
@@ -37,31 +38,39 @@ func (m *DynamicFileMapper) Start(stopCh <-chan struct{}) error {
...
-	return nil, mapper.ErrNotMapped
+	return nil, errutil.ErrNotMapped
+}
+
+func (m *DynamicFileMapper) match(token *token.Identity, mappedUserID, mappedARN string) error {
+	if m.userIDStrict {
+		// ARN must be validated along with UserID.  This avoids having to support IAM user
+		// name/ARN changes. Without preventing this the mapping would look invalid but still
+		// work and auditing would be difficult/impossible.
+		if token.ARN != mappedARN {
```
wongma7, line 69: is arn required if userid is set? if arn is not required and not set then this will not work

pkg/mapper/dynamicfile/mapper.go
```diff
@@ -37,31 +38,39 @@ func (m *DynamicFileMapper) Start(stopCh <-chan struct{}) error {
...
-	rm, err := m.RoleMapping(key)
-	// TODO: Check for non Role/UserNotFound errors
-	if err == nil {
-		return &config.IdentityMapping{
-			IdentityARN: canonicalARN,
-			Username:    rm.Username,
-			Groups:      rm.Groups,
-		}, nil
+	if roleMapping, err := m.RoleMapping(key); err == nil {
+		if err := m.match(identity, roleMapping.RoleARN, roleMapping.UserId); err != nil {
```
micahhausler, line 48: Is there a test we can add a new case for to cover this?

pkg/mapper/dynamicfile/mapper.go
nckturner: I've modified to allow the UserARN to be empty, so that the user can choose whether or not they want that check, and it won't fail if its not provided.

pkg/mapper/dynamicfile/mapper.go
```diff
@@ -37,31 +39,40 @@ func (m *DynamicFileMapper) Start(stopCh <-chan struct{}) error {
...
+	return nil, errutil.ErrNotMapped
+}
+
+func (m *DynamicFileMapper) match(token *token.Identity, mappedARN, mappedUserID string) error {
+	if m.userIDStrict {
+		// If ARN is provided, ARN must be validated along with UserID.  This avoids having to
+		// support IAM user name/ARN changes. Without preventing this the mapping would look
+		// invalid but still work and auditing would be difficult/impossible.
+		if mappedARN != "" && token.ARN != mappedARN {
+			logrus.Warnf("errorororororororor: %v, %v", token.ARN, mappedARN)
```
micahhausler, line 71: 🤣

nckturner: aaaaaand fixed :)

micahhausler (approved):
/lgtm
/approve

PR #669, opened 2023-12-09
Fix role ARN comparison for user ID strict check

DanielCKennedy:

Fixes UserID strict ARN comparison for roles by comparing mapped arn against the identity's canonical arn (role representation) rather than the identity's arn (assume role format). Additionally strips the role path off of the dynamic file role mapping for this check since assumed roles don't have path info.

Fixes #

### DIFF
--- a/pkg/mapper/dynamicfile/mapper.go
+++ b/pkg/mapper/dynamicfile/mapper.go
@@ -3,6 +3,7 @@ package dynamicfile
 import (
 	"strings"
 
+	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/errutil"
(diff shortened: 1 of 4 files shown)


pkg/mapper/dynamicfile/mapper.go
```diff
@@ -66,7 +67,8 @@ func (m *DynamicFileMapper) match(token *token.Identity, mappedARN, mappedUserID
 		// If ARN is provided, ARN must be validated along with UserID.  This avoids having to
 		// support IAM user name/ARN changes. Without preventing this the mapping would look
 		// invalid but still work and auditing would be difficult/impossible.
-		if mappedARN != "" && token.ARN != mappedARN {
+		strippedArn, _ := arn.StripPath(mappedARN)
```
nckturner, line 70: Can you add a `// TODO:` that says something along the lines of: this should potentially be moved to canonicalize or somehow shared by other mappers because it could fix the path issue but needs to be tested with other mappers first, e.g. in a cluster with existing custom resources for the CRD mapper.

PR #674, opened 2023-12-10
fix arn comparison for IAM ARN

nnmin-aws:
this PR fix the issue in PR 669. The original fix doesn't work.

Fixes #

### DIFF
--- a/pkg/mapper/dynamicfile/mapper.go
+++ b/pkg/mapper/dynamicfile/mapper.go
@@ -3,6 +3,7 @@ package dynamicfile
 import (
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/errutil"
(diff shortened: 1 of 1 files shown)


pkg/mapper/dynamicfile/mapper.go
```diff
@@ -46,31 +47,35 @@ func (m *DynamicFileMapper) Map(identity *token.Identity) (*config.IdentityMappi
 	}
 
 	if roleMapping, err := m.RoleMapping(key); err == nil {
-		if err := m.match(identity, roleMapping.RoleARN, roleMapping.UserId); err != nil {
+		if err := m.match(canonicalARN, roleMapping.RoleARN); err != nil {
```
nckturner, line 50: So should we remove the match call to role since only user ARNs can change?

pkg/mapper/dynamicfile/mapper.go
```diff
@@ -46,31 +47,35 @@ func (m *DynamicFileMapper) Map(identity *token.Identity) (*config.IdentityMappi
...
 
-func (m *DynamicFileMapper) match(token *token.Identity, mappedARN, mappedUserID string) error {
+func (m *DynamicFileMapper) match(canonicalARN string, mappingARN string) error {
 	if m.userIDStrict {
 		// If ARN is provided, ARN must be validated along with UserID.  This avoids having to
 		// support IAM user name/ARN changes. Without preventing this the mapping would look
 		// invalid but still work and auditing would be difficult/impossible.
-		strippedArn, _ := arn.StripPath(mappedARN)
-		if strippedArn != "" && token.CanonicalARN != strings.ToLower(strippedArn) {
+		// compare arn additionally for IAM user if principalId is used in mapping
```
nckturner, line 71: This comment is the same as the text above, I dont think it needs to be added.

pkg/mapper/dynamicfile/mapper.go
nnmin-aws: thank you for the comments! yeah we can refactor it in future

pkg/mapper/dynamicfile/mapper.go
nnmin-aws: thank you for the comments! yeah we can refactor it in future

PR #690, opened 2024-01-30
add support for e2e latency for dynamic mode and adoption rate metrics

minj131:

This adds logic to support e2e latency for dynamic mode and dynamic backend mode files.

Fixes #

### DIFF
--- a/pkg/fileutil/util.go
+++ b/pkg/fileutil/util.go
@@ -1,6 +1,7 @@
 package fileutil
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 	"time"
@@ -109,16 +110,20 @@ func StartLoadDynamicFile(filename string, callBack FileChangeCallBack, stopCh <
 func CalculateTimeDeltaFromUnixInSeconds(from, to string) (float64, error) {
 	parsedFrom, err := strconv.ParseInt(from, 10, 64)
 	if err != nil {
-		return 0, err
+		return 0, fmt.Errorf("failed to parse 'from' string: %v", err)
 	}
 
 	parsedTo, err := strconv.ParseInt(to, 10, 64)
 	if err != nil {
-		return 0, err
+		return 0, fmt.Errorf("failed to parse 'to' string: %v", err)
 	}
 
 	timeFrom := time.Unix(parsedFrom, 0).UTC()
 	timeTo := time.Unix(parsedTo, 0).UTC()
+
+	if timeFrom.After(timeTo) {
+		return 0, fmt.Errorf("start timestamp is after end timestamp")
+	}
 
 	return timeTo.Sub(timeFrom).Seconds(), nil
 }
--- a/pkg/fileutil/util.go
+++ b/pkg/fileutil/util.go
@@ -2,6 +2,7 @@ package fileutil
 
 import (
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -103,4 +104,21 @@ func StartLoadDynamicFile(filename string, callBack FileChangeCallBack, stopCh <
 			}
 		}
 	}, time.Second, stopCh)
+}
+
+func CalculateTimeDeltaFromUnixInSeconds(from, to string) (float64, error) {
+	parsedFrom, err := strconv.ParseInt(from, 10, 64)
+	if err != nil {
+		return 0, err
+	}
+
+	parsedTo, err := strconv.ParseInt(to, 10, 64)
+	if err != nil {
+		return 0, err
+	}
+
+	timeFrom := time.Unix(parsedFrom, 0).UTC()
+	timeTo := time.Unix(parsedTo, 0).UTC()
+
+	return timeTo.Sub(timeFrom).Seconds(), nil
 }
--- a/pkg/fileutil/util.go
+++ b/pkg/fileutil/util.go
@@ -107,23 +107,17 @@ func StartLoadDynamicFile(filename string, callBack FileChangeCallBack, stopCh <
 	}, time.Second, stopCh)
 }
 
-func CalculateTimeDeltaFromUnixInSeconds(from, to string) (float64, error) {
-	parsedFrom, err := strconv.ParseInt(from, 10, 64)
+func CalculateTimeDeltaFromUnixInSeconds(from string) (int64, error) {
+	startTime, err := strconv.ParseInt(from, 10, 64)
 	if err != nil {
-		return 0, fmt.Errorf("failed to parse 'from' string: %v", err)
+		return 0, fmt.Errorf("failed to parse 'startTime' string: %v", err)
 	}
 
-	parsedTo, err := strconv.ParseInt(to, 10, 64)
-	if err != nil {
-		return 0, fmt.Errorf("failed to parse 'to' string: %v", err)
-	}
+	endTime := time.Now().Unix()
 
-	timeFrom := time.Unix(parsedFrom, 0).UTC()
-	timeTo := time.Unix(parsedTo, 0).UTC()
-
-	if timeFrom.After(timeTo) {
+	if startTime > endTime {
 		return 0, fmt.Errorf("start timestamp is after end timestamp")
 	}
 
-	return timeTo.Sub(timeFrom).Seconds(), nil
+	return endTime - startTime, nil
 }
--- a/pkg/fileutil/util_test.go
+++ b/pkg/fileutil/util_test.go
@@ -122,3 +122,34 @@ func TestDeleteDynamicFile(t *testing.T) {
 	}
 	testA.mutex.Unlock()
 }
+
+func TestCalculateTimeDeltaFromUnixInSeconds(t *testing.T) {
+	type args struct {
+		from string
+		to   string
+	}
+	cases := []struct {
+		input args
+		want  float64
+	}{
+		{
+			args{"1706648530", "1706648539"},
+			9.0,
+		},
+		{
+			args{"1706648520", "1706648539"},
+			19.0,
+		},
+	}
+
+	for _, c := range cases {
+		out, err := CalculateTimeDeltaFromUnixInSeconds(c.input.from, c.input.to)
+		if err != nil {
+			t.Errorf("error is not expected")
+		}
+
+		if out != c.want {
+			t.Errorf("unexpected result: got %v but expected %v", out, c.want)
+		}
+	}
+}
--- a/pkg/fileutil/util_test.go
+++ b/pkg/fileutil/util_test.go
@@ -1,6 +1,7 @@
 package fileutil
 
 import (
+	"fmt"
 	"os"
 	"sync"
 	"testing"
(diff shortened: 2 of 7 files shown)


pkg/metrics/metrics.go
```diff
@@ -98,5 +99,14 @@ func createMetrics(reg prometheus.Registerer) Metrics {
 				Help:      "Number of EC2 describe instances calls.",
 			},
 		),
+		E2ELatency: factory.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:      "dynamic_e2e_latency_seconds",
+				Namespace: Namespace,
+				Help:      "End to end latency in seconds partitioned by type.",
+				Buckets:   []float64{0.25, 0.5, 0.7, 1, 1.5, 3, 5, 10},
```
nnmin-aws, line 107: what's the data point in testing? is 0.25, 0.5, 0.75 too small

pkg/server/types.go
```diff
@@ -39,7 +39,20 @@ type BackendMapper struct {
 	currentModes string
 }
 
+// Meta is the collection of fields which should be included in all top-level
+// objects which are propagated for dynamic file mode
+type Meta struct {
+	APIVersion string `json:"ApiVersion"`
```
jyotimahapatra, line 45: how is APIVersion used? Would it have constraint on being a semver?

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -48,6 +68,7 @@ func NewDynamicFileMapStore(cfg config.Config) (*DynamicFileMapStore, error) {
 	ms := DynamicFileMapStore{}
 	ms.filename = cfg.DynamicFilePath
 	ms.userIDStrict = cfg.DynamicFileUserIDStrict
+	ms.skip = true
```
jyotimahapatra, line 71: what does `skip` mean? Could we use a name that tells what is being skipped?

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -21,9 +25,25 @@ type DynamicFileMapStore struct {
 	filename                  string
 	userIDStrict              bool
 	usernamePrefixReserveList []string
+
+	skip bool
+}
+
+// Meta is the collection of fields which should be included in all top-level
+// objects which are propagated for dynamic file mode
+type Meta struct {
```
jyotimahapatra, line 34: is this struct a duplicate? I notice another in `pkg/server/types.go` .

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -165,6 +186,24 @@ func (ms *DynamicFileMapStore) CallBackForFileLoad(dynamicContent []byte) error
 		return err
 	}
 	ms.saveMap(userMappings, roleMappings, awsAccounts)
+
+	// emit load latency and if first time, skip emit and set to false to avoid instance bounce issue
```
jyotimahapatra, line 190: could we describe more about what does `instance bounce` mean ?

pkg/server/server.go
```diff
@@ -513,6 +516,24 @@ func (h *handler) CallBackForFileLoad(dynamicContent []byte) error {
...
+	// emit load latency and if first time, skip emit and set to false to avoid instance bounce issue
+	if h.skip {
+		h.skip = false
+	} else {
+		latency, err := fileutil.CalculateTimeDeltaFromUnixInSeconds(backendModes.LastUpdatedDateTime, strconv.FormatInt(time.Now().Unix(), 10))
+		if err != nil {
+			return fmt.Errorf("error parsing latency for dynamic backend mode file: %v", err)
+		}
+		metrics.Get().E2ELatency.WithLabelValues("dynamic_backend_mode").Observe(latency)
+		logrus.WithFields(logrus.Fields{
```
jyotimahapatra, line 529: instead of logging these, would it be possible to add dimensions on the metrics itself?

pkg/mapper/dynamicfile/dynamicfile.go
minj131: addressed, updated the comment

pkg/server/types.go
minj131: As of right now, it is not being used. It would have a constraint on being semvar though.

pkg/mapper/dynamicfile/dynamicfile.go
minj131: skip here means we ignore the first time the metric is emitted to solve the authenticator/instance restart issue. Updated the name to be more descriptive.

pkg/mapper/dynamicfile/dynamicfile.go
minj131: removed the duplicate

pkg/metrics/metrics.go
```diff
@@ -108,5 +110,19 @@ func createMetrics(reg prometheus.Registerer) Metrics {
...
+		AdoptionRate: factory.NewGauge(
+			prometheus.GaugeOpts{
+				Name:      "adoption_rate",
+				Namespace: Namespace,
+				Help:      "Dynamic file in backend mode is enabled",
+			},
+		),
+		AdoptionRateCamOnly: factory.NewGauge(
+			prometheus.GaugeOpts{
+				Name:      "adoption_rate_cam_only",
```
nnmin-aws, line 122: can we update to dynamic_file_enabled and dynamic_file _only?

minj131: good idea, fixed!

pkg/fileutil/util_test.go
```diff
@@ -122,3 +122,34 @@ func TestDeleteDynamicFile(t *testing.T) {
...
+			args{"1706648530", "1706648539"},
+			9.0,
+		},
+		{
+			args{"1706648520", "1706648539"},
+			19.0,
+		},
+	}
+
+	for _, c := range cases {
```
nnmin-aws, line 145: can you add a negative test case that invalid format?

minj131: added

pkg/server/server.go
```diff
@@ -513,6 +516,34 @@ func (h *handler) CallBackForFileLoad(dynamicContent []byte) error {
...
+			logrus.WithFields(logrus.Fields{
+				"ClusterId": backendModes.ClusterID,
+				"Version":   backendModes.Version,
+				"Type":      "dynamic_backend_mode",
+				"Latency":   latency,
+			}).Infof("logging latency metric")
+		}
+	}
+
+	if strings.Contains(h.backendMapper.currentModes, mapper.ModeDynamicFile) {
```
nnmin-aws, line 541: will this logic work? dynamic_file_only will be counted as dynamic_file_enabled

pkg/server/server.go
```diff
@@ -513,6 +516,34 @@ func (h *handler) CallBackForFileLoad(dynamicContent []byte) error {
 	} else {
 		logrus.Infof("BackendMode dynamic file got changed, but same with current mode, skip rebuild mapper")
 	}
+
+	// when instance or container restarts, the dynamic file is (re)loaded and the latency metric is calculated
+	// regardless if there was a change upstream, and thus can emit an incorrect latency value
+	// so a workaround is to skip the first time the metric is calculated, and only emit metris after
+	// as we know any subsequent calculations are from a valid change upstream
```
jyotimahapatra, line 523: if this path is always setting the value to false, why do we need the if condition

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -21,9 +25,25 @@ type DynamicFileMapStore struct {
...
 	userIDStrict              bool
 	usernamePrefixReserveList []string
+
+	skipFirstTimeMetric bool
+}
+
+// Meta is the collection of fields which should be included in all top-level
+// objects which are propagated for dynamic file mode
+type Meta struct {
+	APIVersion string `json:"ApiVersion"`
```
jyotimahapatra, line 35: i'd remove APIVersion if its not used. If this field is not persisted, i can't imagine what it will be used for

pkg/server/types.go
```diff
@@ -41,5 +42,6 @@ type BackendMapper struct {
 
 // AccessConfig represents the configuration format for cluster access config via backend mode.
 type BackendModeConfig struct {
+	dynamicfile.Meta
```
jyotimahapatra, line 45: i dont' feel that this config needs teh extra fields. Could you tell more why the new fields are relevant for other backend modes

minj131: The access config file is propagated via CAM, and so the extra fields are needed to calculate the latency and version of that change, regardless if dynamic file was enabled or not. I did remove the nested struct field as it's not specific to dynamicfile.

pkg/server/server.go
```diff
@@ -513,6 +516,31 @@ func (h *handler) CallBackForFileLoad(dynamicContent []byte) error {
...
+	if h.initDone {
```
nnmin-aws, line 524: did you set h.initDone=true?

pkg/server/server.go
```diff
@@ -513,6 +516,32 @@ func (h *handler) CallBackForFileLoad(dynamicContent []byte) error {
...
```
jyotimahapatra, line 524: rename this to `backendModeConfigInitialized`

pkg/server/server.go
```diff
@@ -513,6 +516,32 @@ func (h *handler) CallBackForFileLoad(dynamicContent []byte) error {
 	} else {
 		logrus.Infof("BackendMode dynamic file got changed, but same with current mode, skip rebuild mapper")
 	}
+
+	// when instance or container restarts, the dynamic file is (re)loaded and the latency metric is calculated
```
jyotimahapatra, line 520: s/dynamic file/access config file/

pkg/fileutil/util.go
```diff
@@ -104,3 +106,19 @@ func StartLoadDynamicFile(filename string, callBack FileChangeCallBack, stopCh <
 		}
 	}, time.Second, stopCh)
 }
+
+func CalculateTimeDeltaFromUnixInSeconds(from string, to int64) (float64, error) {
```
jyotimahapatra, line 110: We are converting the data type from int64 to string and converting it back to int64 for comparison.
Please make it more focused by passing (string, int64) and only converting the first argument to int64.

jyotimahapatra: change the param names to startTime and endTime

pkg/fileutil/util.go
```diff
@@ -104,3 +106,19 @@ func StartLoadDynamicFile(filename string, callBack FileChangeCallBack, stopCh <
...
+	parsedFrom, err := strconv.ParseInt(from, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse 'from' string: %v", err)
+	}
+
+	timeFrom := time.Unix(parsedFrom, 0).UTC()
```
jyotimahapatra, line 116: if the time is already UTC, why do we convert it here back to UTC?

pkg/fileutil/util.go
jyotimahapatra: can we drop the to parameter completely? The callers are passing time.Now anyway.

pkg/mapper/dynamicfile/dynamicfile.go
```diff
@@ -21,13 +24,18 @@ type DynamicFileMapStore struct {
 	filename                  string
 	userIDStrict              bool
 	usernamePrefixReserveList []string
+
+	dynamicFileInitDone bool
 }
 
 type DynamicFileData struct {
+	// Time that the object takes from update time to load time
+	LastUpdatedDateTime string `json:"LastUpdatedDateTime"`
```
jyotimahapatra, line 33: This being a string is going to cause problems in the future. We have to migrate this to the right representation(epoch time is always int64) in a future change.

pkg/server/types.go
```diff
@@ -41,5 +41,9 @@ type BackendMapper struct {
 
 // AccessConfig represents the configuration format for cluster access config via backend mode.
 type BackendModeConfig struct {
+	// Time that the object takes from update time to load time
+	LastUpdatedDateTime string `json:"LastUpdatedDateTime"`
```
jyotimahapatra, line 45: The var name does not tell whethere it expects epoch time or mm-yy format. In a future change this var should be made more readable.

pkg/mapper/dynamicfile/dynamicfile_test.go
```diff
@@ -133,6 +144,10 @@ var origFileContent = `
 
 var updatedFileContent = `
 {
+  "ApiVersion": "1",
+  "Version": "1",
```
jyotimahapatra, line 148: The ApiVersion should be removed?

pkg/mapper/dynamicfile/dynamicfile.go
minj131: Will take an AI, this requires changes across all components unfortunately

pkg/fileutil/util_test.go
```diff
@@ -122,3 +122,35 @@ func TestDeleteDynamicFile(t *testing.T) {
...
+			false,
+		},
+		{
+			args{"foo"},
+			true,
+		},
+	}
+
+	for _, c := range cases {
+		_, err := CalculateTimeDeltaFromUnixInSeconds(c.input.startTime)
```
jyotimahapatra, line 149: test the variable thats ignored as `_` here

jyotimahapatra (approved):
/lgtm

PR #693, opened 2024-02-08
feat: add support for EC2 Instance Identity roles

### LINKED ISSUE #696: [Feature request]: support EC2 instance identities
### What would you like to be added?

EC2 instance identities (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-identity-roles.html) are unique ad-hoc IAM roles assigned to EC2 Instances. They are not currently supported by aws-iam-authenticator

### Why is this needed?

opened to match pull request [https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/693]

First and foremost because they're there and support can be enabled. As implemented, aws-iam-authenticator throws an incorrect error.

I envision two possible use cases: cluster admission control and limited pre-access.

1. cluster admission control - in this scenario, a node candidate will be unable to connect as a node to the cluster until authorized by some other means (let's say an integrity check or security audit). A single IAM role shared by many nodes is unsuitable for this purpose, but ad-hoc identities are. The authorizing mechanism will add the relevant credentials to the auth-map once the node has been vetted.

2. limited pre-access. while candidate nodes are assumed to have system:nodes / system:bootstrapper privileges which are elevated, using them directly may be undesirable security-wise, for two reasons:

a. it may rightfully trigger a violation from monitoring tools 
b. it entails using a superuser for what may be better served by a user with limited access

thus, using a scoped user for AWS identities may allow e.g. read-only access for data that might be useful by the node to configure/tune itself or other such customizations.

_No response_

gilbahat:

This PR enables aws-iam-authenticator to support EC2 Instance Identity roles - https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-identity-roles.html , by parsing their UserID format. This enables use cases such as limited dynamic ad-hoc access to a kubernetes cluster based on AWS identity.

as of now, this prints an  error: "sts getCallerIdentity failed: malformed UserID \"5553582100:aws:ec2-instance:i-1234567890\"" method=POST path=/authenticate , this error is incorrect as the UserID is well formed (just not what was expected)

fixes #696

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -623,6 +623,7 @@ func getIdentityFromSTSResponse(id *Identity, wrapper getCallerIdentityWrapper) 
 	// 1. UserID:SessionName (for assumed roles)
 	// 2. UserID (for IAM User principals).
 	// 3. AWSAccount:CallerSpecifiedName (for federated users)
+	// 4. AWSAccount:aws:ec2-instance:instance-id (for EC2 instance identity roles)
 	// We want the entire UserID for federated users because otherwise,
 	// its just the account ID and is indistinguishable from the UserID
 	// of the root user.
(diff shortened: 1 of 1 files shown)


micahhausler:
Thanks for this PR! Can you open a GH issue so we can discuss the use case? I want to better understand why this is needed as an alternative to the EC2 instance role

orsenthil:
>  This enables use cases such as limited dynamic ad-hoc access to a kubernetes cluster based on AWS identity.

Under what scenarios would be this useful ?

PR #714, opened 2024-04-26
fix: Add random string to e2e test role to avoid pipeline run conflicts

bryantbiggs:
- Add random string to e2e test role to avoid pipeline run conflicts. When running multiple e2e tests concurrently across PRs, using the same role name means that the first process to finish will delete the role and the other processes will lose permission and fail, or the role will not be found for use

Reference: https://github.com/search?q=repo%3Akubernetes-sigs%2Faws-iam-authenticator%20aws-iam-authenticator-test-role-&type=code

Fixes #

image


bryantbiggs:
cc @wongma7 / @jaidevmane for the various test failures due to the IAM role not found - there is a conflict when multiple e2e tests are run concurrently across workflows since the same role name is used "globally"

wongma7:
/lgtm
/approve

thank you. 

it may be hardcoded to avoid roles leaking or something like that. But that is easy to solve in other ways i.e. sweeping them up with ttl.

PR #733, opened 2024-07-10
skip service validation to get the default regions endpoint

kmala:

We had previously made changes to validate token if the hostname matches the instance metadata region in #556 . However it won't work if this is a new partition as the new partition may not have the sdk updated with the services. Hence extending the same logic to validate token even if the service doesn't exist in the sdk. This should be fine as we are only validating the instance metadata region and just ignoring if the service exists in the partition in the sdk which is the done by the default sdk client also https://github.com/aws/aws-sdk-go/blob/main/aws/session/session.go#L936 

Fixes #


pkg/token/token.go
```diff
@@ -392,6 +392,18 @@ type tokenVerifier struct {
 	validSTShostnames map[string]bool
 }
 
+func getDefaultHostNameForRegion(partition *endpoints.Partition, region string) (string, error) {
+	rep, err := partition.EndpointFor(stsServiceID, region, endpoints.STSRegionalEndpointOption, endpoints.ResolveUnknownServiceOption)
```
jyotimahapatra, line 396: qs for familiarity:  Looks like `endpoints.ResolveUnknownServiceOption` will return an error [here[(https://github.com/aws/aws-sdk-go/blob/fbe26bf118cf10a5c6aef4a105e2a5609c617b1a/aws/endpoints/v3model.go#L218-L221). What was the current behavior of `partition.EndpointFor` when `endpoints.ResolveUnknownServiceOption` was false?

pkg/token/token.go
```diff
@@ -410,6 +422,14 @@ func stsHostsForPartition(partitionID, region string) map[string]bool {
 	stsSvc, ok := partition.Services()[stsServiceID]
 	if !ok {
 		logrus.Errorf("STS service not found in partition %s", partitionID)
+		// Add the host of the current instances region if the service doesn't already exists in the partition
+		// so we don't fail if the service is not present in the go sdk but matches the instances region.
+		stsHostName, err := getDefaultHostNameForRegion(partition, region)
+		if err != nil {
+			logrus.WithError(err).Error("Error getting default hostname")
+		} else {
+			validSTShostnames[stsHostName] = true
```
jyotimahapatra, line 431: qs for familiarity: can you describe how this helps? when the sdk does not have the partition, the else block will not be taken. How does the rest of the flow succeed?

pkg/token/token.go
kmala: currently it would throw service not found https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go#L412-L413 as the services are not yet updated in the sdk go

pkg/token/token.go
kmala: when the sdk doesn't have partition we fail as we are now aware of what a default dns hostname should be. This would help where there is partition but the services are not yet defined under the partition.

pkg/token/token.go
nnmin-aws: discussed offline, will add an unit test here

PR #737, opened 2024-08-06
fix: check credential expiration timestamp when generating tokens

### LINKED ISSUE #590: API Authorization from Outside EKS Cluster throws `Unauthorized` error after 20+ minutes

alvaroaleman:
What this PR does / why we need it:

There are two expirations which must be considered when using a signed EKS token:

* 15 minutes after the point in time when the AWS STS request has been signed
* The underlying AWS credentials can expire at which point the token won't be accepted

The second case is particularly common when making frequent requests while using AssumeRole or AssumeRoleWithWebRequest as mentioned in #590 as the default session timeout is 1 hour.

This PR adds an additional check fetching the AWS credential expiration and using that as the returned expiration if it is before the 15 minute token expiration.

Which issue(s) this PR fixes

Fixes #590

Fixes #

Replaces https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/724
/assign @micahhausler

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -335,8 +335,16 @@ func (g generator) GetWithSTS(clusterID string, stsAPI stsiface.STSAPI) (Token, 
 		return Token{}, err
 	}
 
+	// Fetch the timestamp when the credentials we're going to use for signing will not be valid anymore
+	// This operation is potentially racey, but the worst case is that we expire a token early
+	// Not all credential providers support this, so we ignore any returned errors
+	credentialsExpiration, _ := request.Config.Credentials.ExpiresAt()
+
 	// Set token expiration to 1 minute before the presigned URL expires for some cushion
 	tokenExpiration := g.nowFunc().Local().Add(presignedURLExpiration - 1*time.Minute)
+	if !credentialsExpiration.IsZero() && credentialsExpiration.Before(tokenExpiration) {
+		tokenExpiration = credentialsExpiration.Add(-1 * time.Minute)
+	}
 	// TODO: this may need to be a constant-time base64 encoding
 	return Token{v1Prefix + base64.RawURLEncoding.EncodeToString([]byte(presignedURLString)), tokenExpiration}, nil
 }
--- a/pkg/token/token_test.go
+++ b/pkg/token/token_test.go
@@ -590,6 +590,10 @@ func Test_getDefaultHostNameForRegion(t *testing.T) {
 func TestGetWithSTS(t *testing.T) {
 	clusterID := "test-cluster"
 
+	// Example non-real credentials
+	decodedAkid, _ := base64.StdEncoding.DecodeString("QVNJQVIyVEc0NFY2QVMzWlpFN0M=")
+	decodedSk, _ := base64.StdEncoding.DecodeString("NEtENWNudEdjVm1MV1JkRjV3dk5SdXpOTDVReG1wNk9LVlk2RnovUQ==")
+
 	cases := []struct {
 		name    string
 		creds   *credentials.Credentials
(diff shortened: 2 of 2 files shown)


micahhausler:
I'll check on this, but I think this PR is unnecessary. Most AWS temporary creds' advertised expiration is well before (as much as 50% of the cred lifetime) the actual credential is invalid. This is meant to increase the stability of an application in the event that the credential distribution method is ever impaired. Ex: If you have a 6 hour cred, but the advertised expiration is after 4 hours and normal issuance of new creds is every 4 hours, you can ride out a new credential distribution disruption for up to 2 hours.

alvaroaleman:
> I'll check on this, but I think this PR is unnecessary. Most AWS temporary creds' advertised expiration is well before (as much as 50% of the cred lifetime) the actual credential is invalid. This is meant to increase the stability of an application in the event that the credential distribution method is ever impaired. Ex: If you have a 6 hour cred, but the advertised expiration is after 4 hours and normal issuance of new creds is every 4 hours, you can ride out a new credential distribution disruption for up to 2 hours.

This change is definitely necessary, we made this after we saw frequent auth failures which this fixes, which is why we are currently using a fork with this change. You might be right that this is less of a problem with longer-lived creds, but we for example have a max session validity of 1h for compliance reason. Without this change, it happens regularly that the creds used to sign this request expire and the returned token stops working.

micahhausler:
I'm ok adding this behavior with a flag (default off) because of the above situation where a cred might be reported as expired, but is still valid, we wouldn't want to create tokens with a reported expiration in the past. 

I'll get #741 merged first which includes a test you should be able to extend for the new behavior

micahhausler:
Ok, I've merged #741, that test should be extensible for this change, with the addition of a flag

alvaroaleman:
@micahhausler thanks for that, I've updated the PR, please have another look

alvaroaleman:
@micahhausler any chance you could have a look?

alvaroaleman:
It would be great if this could get a review
/remove-lifecycle stale

PR #738, opened 2024-08-06
Update dependency to v1.31.0-rc

shizuchanw:
Version release for upstream 1.31.0-rc.0
N/A


bryantbiggs:
I think we should wait for the GA release version and avoid using RC versions for stability

shizuchanw:
> I think we should wait for the GA release version and avoid using RC versions for stability

Makes sense, closing the PR

Issue #736, opened 2024-07-26
switch to aws-sdk-go-v2 as v1 is going away


dims:
Context: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/
Migration guide: https://aws.github.io/aws-sdk-go-v2/docs/migrating/

Basically maintenance mode on July 31, 2024 and reach end-of-support on July 31, 2025. The SDK will not 
receive API updates for new or existing services, or be updated to support new regions.

bored-engineer:
For folks that are only using this package to generate client-side tokens (i.e. [github.com/kubernetes-sigs/aws-iam-authenticator/pkg/token](https://pkg.go.dev/github.com/kubernetes-sigs/aws-iam-authenticator/pkg/token)), I have published a small Golang package that uses the AWS v2 SDK instead: https://github.com/bored-engineer/aws-eks-auth

dims:
cc @prateekgogia @kmala

micahhausler:
In working on the V2 upgrade, one key difference is that the new SDK does _not_ expose the internal list of known endpoints for a particular partition. 

* aws/aws-sdk-go-v2#1026
* aws/aws-sdk-go-v2#2336

The current behavior is that the server builds up a list of known regions from SDK v1 for a given partition (specified via a server flag), and ensures that the domain of incoming pre-signed requests is one of those known endpoints. The reason we don't want to allow cross-partition STS tokens, is that AWS treats partitions (`aws`, `aws-cn`, `aws-us-gov`, etc) [as fully isolated](https://docs.aws.amazon.com/whitepapers/latest/aws-fault-isolation-boundaries/partitions.html) fault and trust domains. The tradeoff we currently make is that a new region requires an SDK update to aws-iam-authenticator. Given that we no longer can get this list of STS endpoints, I'll need to add new functionality for region discovery. 

Similar to how we support multiple backend mappers for ARN to username discovery, I'll start out with region discovery in the following formats:
* SDK v1 based (deprecated)
* [`account:ListRegions`](https://docs.aws.amazon.com/accounts/latest/reference/API_ListRegions.html) based discovery (with some periodic cache update)
* File based input. 

We'll default to SDKv1 based behavior to preserve backward compatibility, and likely drop support around the time SDKv1 reaches end of life on July 31, 2025. At that point, we'll change the default to `account:ListRegions`.

[...]

PR #741, opened 2024-08-15
Add configurable Now time for signature generation

micahhausler:
This behavior and test is prep for an AWS SDK update, to ensure that the generated token (signature) matches when we update the SDK.

Preparation for #736

### DIFF
--- a/pkg/token/token_test.go
+++ b/pkg/token/token_test.go
@@ -14,7 +14,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/google/go-cmp/cmp"
 	"github.com/prometheus/client_golang/prometheus"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/token/token_test.go
```diff
@@ -582,3 +586,61 @@ func Test_getDefaultHostNameForRegion(t *testing.T) {
...
+					string(decodedSk),
+					"",
+				)
+			}(),
+			time.Unix(1682640000, 0),
+			Token{
+				Token:      "k8s-aws-v1.aHR0cHM6Ly9zdHMudXMtd2VzdC0yLmFtYXpvbmF3cy5jb20vP0FjdGlvbj1HZXRDYWxsZXJJZGVudGl0eSZWZXJzaW9uPTIwMTEtMDYtMTUmWC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BU0lBUjJURzQ0VjZBUzNaWkU3QyUyRjIwMjMwNDI4JTJGdXMtd2VzdC0yJTJGc3RzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyMzA0MjhUMDAwMDAwWiZYLUFtei1FeHBpcmVzPTAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JTNCeC1rOHMtYXdzLWlkJlgtQW16LVNpZ25hdHVyZT00ZDdhYmZkZTk2NzI1ZWI4YTc3MzgyNDg0MTZlNGI1ZDA4ZDlkYmQ3MThiNGY2ZGQ2OTBmOGZiNzUwMTMyOWQ1",
+				Expiration: time.Unix(1682640000, 0).Local().Add(time.Minute * 14),
+			},
+			nil,
```
jyotimahapatra, line 617: do we need a test case that has an error?

micahhausler: The only possible error is a `request.Presign()` error, which I'm not really concerned with for the purposes of this test

PR #742, opened 2024-08-16
Inject log level via CLI flag; add debug logging for incoming requests

modulitos:

Allow users to inject the log level via a `--loglevel` flag. It is `info` by default, which is also the default for the logrus logger.

Adds a debug log statement to log incoming HTTP requests.

Fixes # N/A

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/http/httputil"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -303,6 +304,17 @@ func (h *handler) authenticateEndpoint(w http.ResponseWriter, req *http.Request)
...
 	})
 
+	if logrus.IsLevelEnabled(logrus.DebugLevel) {
+		reqDump, err := httputil.DumpRequest(req, true)
+		if err != nil {
+			log.Info(err, "Failed to dump http request")
+		}
+
+		log.WithFields(logrus.Fields{
+			"request": string(reqDump),
```
micahhausler, line 314: are we sure we want this? It'll dump all tokens to logs

modulitos: Good point... I'll update this to only log safe fields from the request

PR #743, opened 2024-08-16
add a namespaced field

modulitos:
**What this PR does / why we need it**:

Adds a namespaced `principalId` field.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes # N/A

PR #747, opened 2024-08-20
Update 0.6 branch with new namespaced user info field

modulitos:

Cherry-pick changes from: https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/743/files

Fixes # N/A


version.txt
```diff
@@ -1 +1 @@
-0.6.24
```
xdu31, line 1: It looks we have the convention to cut release in a separate PR, not sure if it can work if we do it together with other changes

modulitos: Thank for the heads up - I removed the change from the PR.

Once this PR is merged, I'll send a separate PR to bump the version again.

modulitos: PR to only bump the version: https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/748

PR #749, opened 2024-08-23
Add sourceArn to sts through headers

haoranleo:
Add cluster details (clusters_arn and account_id) as headers to STS when assuming a role to make requests. When sourceARN is not empty, it will panic if it fails to parse account id from the sourceARN (same behavior as verifying roleARN).

The reason to add sourceArn and sourceAccount to STS request headers is to avoid confusion deputy issue. AWS IAM supports these global conditional key in IAM roles trust policy https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html, and we want to include the caller identity into the STS requests when assuming the role. We have did same change for CCM in https://github.com/kubernetes/cloud-provider-aws/pull/649

Fixes #

### DIFF
--- a/pkg/ec2provider/ec2provider.go
+++ b/pkg/ec2provider/ec2provider.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
@@ -35,6 +36,11 @@ const (
 	// Maximum time in Milliseconds to wait for a new batch call this also depends on if the instance size has
 	// already become 100 then it will not respect this limit
 	maxWaitIntervalForBatch = 200
+
+	// Headers for STS request for source ARN
+	headerSourceArn = "x-amz-source-arn"
+	// Headers for STS request for source account
+	headerSourceAccount = "x-amz-source-account"
 )
 
 // Get a node name from instance ID
@@ -60,7 +66,7 @@ type ec2ProviderImpl struct {
 	instanceIdsChannel chan string
 }
 
-func New(roleARN, region string, qps int, burst int) EC2Provider {
+func New(roleARN, sourceARN, region string, qps int, burst int) EC2Provider {
 	dnsCache := ec2PrivateDNSCache{
 		cache: make(map[string]string),
 		lock:  sync.RWMutex{},
@@ -70,7 +76,7 @@ func New(roleARN, region string, qps int, burst int) EC2Provider {
 		lock: sync.RWMutex{},
 	}
 	return &ec2ProviderImpl{
-		ec2:                ec2.New(newSession(roleARN, region, qps, burst)),
+		ec2:                ec2.New(newSession(roleARN, sourceARN, region, qps, burst)),
 		privateDNSCache:    dnsCache,
 		ec2Requests:        ec2Requests,
 		instanceIdsChannel: make(chan string, maxChannelSize),
@@ -81,7 +87,7 @@ func New(roleARN, region string, qps int, burst int) EC2Provider {
 // the environment, shared credentials (~/.aws/credentials), or EC2 Instance
 // Role.
 
-func newSession(roleARN, region string, qps int, burst int) *session.Session {
+func newSession(roleARN, sourceARN, region string, qps int, burst int) *session.Session {
 	sess := session.Must(session.NewSession())
 	sess.Handlers.Build.PushFrontNamed(request.NamedHandler{
 		Name: "authenticatorUserAgent",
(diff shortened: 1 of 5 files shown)


haoranleo:
Thanks @bryantbiggs, could you review when you have a chance?

bryantbiggs:
code wise things look good - but I am not an IAM expert so I'll let Micah or someone else weigh in for final approval

/lgtm

micahhausler:
I don't think we can merge this. The additional header is not signed, and can be changed by a middleman. If you did add it to the signed header, thats not backward compatible. We do have an existing scoping/aud check with the cluster ID

micahhausler:
Ok, I was misunderstanding, this is about the _server side_ role assumption for EC2 instance verification. Yea I think this check is probably fine to add.

haoranleo:
@micahhausler would you like to review when you have a chance?

pkg/ec2provider/ec2provider.go
```diff
@@ -277,3 +285,41 @@ func (p *ec2ProviderImpl) getPrivateDnsAndPublishToCache(instanceIdList []string
 		p.unsetRequestInFlightForInstanceId(id)
 	}
 }
+
+func applySTSRequestHeaders(stsClient *sts.STS, roleARN, sourceARN string) *sts.STS {
+	logrus.Infof("Using AWS assumed role %v", roleARN)
+	sourceAcct, err := getSourceAccount(roleARN)
```
micahhausler, line 291: Shouldn't the account ID be pulled from the sourceARN and not the role ARN?

pkg/ec2provider/ec2provider.go
```diff
@@ -60,7 +66,7 @@ type ec2ProviderImpl struct {
 	instanceIdsChannel chan string
 }
 
-func New(roleARN, region string, qps int, burst int) EC2Provider {
+func New(roleARN, sourceARN, region string, qps int, burst int) EC2Provider {
```
micahhausler, line 69: This new field should be optional, can we make it optional?

micahhausler (changes requested):

pkg/ec2provider/ec2provider.go
```diff
@@ -277,3 +285,41 @@ func (p *ec2ProviderImpl) getPrivateDnsAndPublishToCache(instanceIdList []string
 		p.unsetRequestInFlightForInstanceId(id)
 	}
 }
+
+func applySTSRequestHeaders(stsClient *sts.STS, roleARN, sourceARN string) *sts.STS {
+	logrus.Infof("Using AWS assumed role %v", roleARN)
+	sourceAcct, err := getSourceAccount(roleARN)
+	if err != nil {
+		logrus.Errorf("failed to parse source account from role ARN %v: %v", roleARN, err)
+		return stsClient
```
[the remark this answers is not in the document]
haoranleo: Good callout, found we have check in the server lifecycle earlier https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/server/server.go#L195-L200. I would prefer still keep the check here for robustness regardless in case new caller invokes EC2 provider function, but this check will be no-op currently

pkg/ec2provider/ec2provider.go
haoranleo: Yeah it is already optional, by default it will be empty string if the value is not set in the config. https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/749/files#diff-8098996ac6450091ed1b73d6860e512c51ed35c8e3fdf67dd2d0f100e1968f7bR299 - it will ignore this header if the value is empty.

pkg/ec2provider/ec2provider.go
micahhausler: They _should_ be the same, but in the event of a config mix up, which one would we want to set?

pkg/ec2provider/ec2provider.go
```diff
@@ -277,3 +285,41 @@ func (p *ec2ProviderImpl) getPrivateDnsAndPublishToCache(instanceIdList []string
 		p.unsetRequestInFlightForInstanceId(id)
 	}
 }
+
+func applySTSRequestHeaders(stsClient *sts.STS, roleARN, sourceARN string) *sts.STS {
+	logrus.Infof("Using AWS assumed role %v", roleARN)
+	sourceAcct, err := getSourceAccount(roleARN)
+	if err != nil {
+		logrus.Errorf("failed to parse source account from role ARN %v: %v", roleARN, err)
```
micahhausler, line 293:
```suggestion
		logrus.WithError(err).WithField("roleARN", roleARN).Error("failed to parse source account from role ARN)
```

pkg/ec2provider/ec2provider.go
haoranleo: We will use the roleARN (ec2DescribeInstancesRoleARN) because this is the role we are going to assume. We did same in CCM https://github.com/kubernetes/cloud-provider-aws/pull/649

pkg/ec2provider/ec2provider.go
haoranleo: Updated

pkg/ec2provider/ec2provider.go
haoranleo: Discussed offline, we are gonna parse source account id from sourceARN to avoid confusion deputy issue in the misconfiguration scenario. Also panic the authenticator when the sourceARN is ill-formatted

micahhausler (approved):
/lgtm
/approve

haoranleo:
@micahhausler could you remove the hold label?

PR #750, opened 2024-08-26
Remove parameterized AWS session from token.go

micahhausler:

This simplifies the API, and removes the unnecessary `GetWithRoleForSession()` method. This also simplifies migration to aws-sdk-go-v2 by allowing both `Generator` and `TokenOptions` to be not bound to a specific SDK version.

Preparation for #736 migration


bryantbiggs:
nice clean-up & simplification!
/lgtm

PR #756, opened 2024-08-29
Filecache sdk upgrade

micahhausler:
Update filecache to use AWS SDK Go V2 with wrappers
    
This changes updates filecache's internal types to use the AWS SDK Go  v2's types, while preserving the external interface used by /pkg/token.This will simplify the future project-wide change for AWS SDK Go v2.

Depends on #755 to be merged first

First change toward #736

### DIFF
--- a/pkg/filecache/converter.go
+++ b/pkg/filecache/converter.go
@@ -1,0 +1,55 @@
+package filecache
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+)
+
+type v2 struct {
+	creds *credentials.Credentials
+}
+
+var _ aws.CredentialsProvider = &v2{}
+
+func (p *v2) Retrieve(ctx context.Context) (aws.Credentials, error) {
+	val, err := p.creds.GetWithContext(ctx)
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+	resp := aws.Credentials{
+		AccessKeyID:     val.AccessKeyID,
+		SecretAccessKey: val.SecretAccessKey,
+		SessionToken:    val.SessionToken,
+		Source:          val.ProviderName,
+		CanExpire:       false,
+		// Don't have account ID
+	}
+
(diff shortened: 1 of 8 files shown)


pkg/token/token.go
```diff
@@ -248,7 +248,11 @@ func (g generator) GetWithOptions(options *GetTokenOptions) (Token, error) {
 			profile = session.DefaultSharedConfigProfile
 		}
 		// create a cacheing Provider wrapper around the Credentials
-		if cacheProvider, err := filecache.NewFileCacheProvider(options.ClusterID, profile, options.AssumeRoleARN, sess.Config.Credentials); err == nil {
+		if cacheProvider, err := filecache.NewFileCacheProvider(
+			options.ClusterID,
+			profile,
+			options.AssumeRoleARN,
+			filecache.V1CredentialToV2Provider(sess.Config.Credentials)); err == nil {
```
jyotimahapatra, line 255: for my education: is the filecache intended to be v1 format until some future time?
reason for asking: at a future point(or in this PR) invalidate the filecache and use v2 format directly?

pkg/filecache/converter.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+	resp := aws.Credentials{
+		AccessKeyID:     val.AccessKeyID,
+		SecretAccessKey: val.SecretAccessKey,
+		SessionToken:    val.SessionToken,
+		Source:          val.ProviderName,
+		CanExpire:       false,
+		// Don't have account ID
```
jyotimahapatra, line 27: for my recollection: does the struct not have accountId? or having account ID breaks any usecase?

pkg/token/token.go
jyotimahapatra: Talked to Micah offline. The change as is is backward compatible. We can invalidate the cache if v1 creds are found in a later tag. The change will be backward incompatible because rollback will not work when filecache is sdkv2 creds.

pkg/filecache/converter.go
micahhausler: The v1 struct does not have an account ID, v2 does. It won't break anything to not include it

PR #757, opened 2024-09-04
Abstract region discovery

micahhausler:

With the deprecation of AWS SDK Go v1, and the removal of the endpoints package, we need a new way to discover valid regions. This change preserves the previous behavior, while adding a new method that calls accounts:ListRegions, or looking up endpoints from a file.

See https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/736#issuecomment-2321734713

Further progress toward #736


pkg/regions/file_discoverer_test.go
```diff
@@ -0,0 +1,89 @@
...
+			fs: func() fs.FS {
+				return aferoFSWrapper{afero.NewMemMapFs()}
+			}(),
+			want:    nil,
+			wantErr: fs.ErrNotExist,
+		},
+		{
+			name: "InvalidJSON",
+			fs: func() fs.FS {
+				tmpfs := afero.NewMemMapFs()
```
modulitos, line 41: I noticed `afero` in the last PR too - this looks pretty useful!

pkg/regions/api_discoverer.go
```diff
@@ -0,0 +1,74 @@
...
+	partition string
+}
+
+func (d *apiDiscoverer) Find(ctx context.Context) (map[string]bool, error) {
+	stsHostnames := map[string]bool{}
+
+	var tlds []string
+	serviceNames := []string{"sts", "sts-fips"}
+	switch d.partition {
+	case "aws":
```
modulitos, line 35: Similar to my enums question above: is it worthwhile to use enums here?

PR #760, opened 2024-09-16
use protobuf content type instead of json for k8s client

bhavi-koduru:

This MR is a part of effort to elevate single eks cluster performance by migrating the EKS components to use protobuf instead of json.

Modify kubeconfig type to use content type application/vnd.kubernetes.protobuf instead of json for performance gain.

Fixes #


bryantbiggs:
FYI - only these folks can approve currently https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/90beff7155038ab8861b7ea9f2c15d1f4c75bddb/OWNERS#L1-L5

PR #767, opened 2024-10-10
add logs and metrics to find sts call success/failures on global/regional endpoints

sushanth0910:
This PR adds few logs to find if the Verify token presigned URL is sent to Global/Regional end point.
This PR also adds new metrics dimention to STS metrics to get requests failed/throttled on regional/global endpoint.

Fixes #

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -78,6 +78,9 @@ type Identity struct {
 	// in conjunction with CloudTrail to determine the identity of the individual
 	// if the individual assumed an IAM role before making the request.
 	AccessKeyID string
+
+	// ASW STS endpoint typ used to authenticate (sts_global/sts_regional)
+	STSEndpointType string
 }
 
 const (
@@ -569,8 +572,6 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 		stsEndpointType = metrics.STSGlobal
 	}
 
-	logrus.Infof("Sending request to %s endpoint, host: %s", stsEndpointType, parsedURL.Host)
-
 	response, err := v.client.Do(req)
 	if err != nil {
 		metrics.Get().StsConnectionFailure.WithLabelValues(stsEndpointType).Inc()
@@ -606,7 +607,8 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 	}
 
 	id := &Identity{
-		AccessKeyID: accessKeyID,
+		AccessKeyID:     accessKeyID,
+		STSEndpointType: stsEndpointType,
 	}
 	return getIdentityFromSTSResponse(id, callerIdentity)
 }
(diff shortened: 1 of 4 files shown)


pkg/token/token.go
```diff
@@ -564,14 +564,23 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 	req.Header.Set(clusterIDHeader, v.clusterID)
 	req.Header.Set("accept", "application/json")
 
+	stsEndpointType := metrics.InvalidSTSEndpoint
```
kmala, line 567: we already validate the endpoint above, is there a chance for it to be invalid?

pkg/token/token.go
```diff
@@ -564,14 +564,23 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
...
+	if parsedURL.Host == "sts.amazonaws.com" {
```
kmala, line 568: do we have global endpoint for non aws partitions?

pkg/token/token.go
sushanth0910: no. i will remove that

pkg/token/token.go
sushanth0910: no, for non aws partition there is no separate global endpoint

pkg/token/token.go
```diff
@@ -78,6 +78,9 @@ type Identity struct {
 	// in conjunction with CloudTrail to determine the identity of the individual
 	// if the individual assumed an IAM role before making the request.
 	AccessKeyID string
+
+	// ASW STS endpoint type(global/regional) used to authenticate (expected values sts_global/sts_regional)
+	STSEndpointType string
```
micahhausler, line 83: Rather than a type with "global" or "regional", lets put the region the request was signed to here. It'll be more informative

micahhausler (changes requested):

pkg/token/token.go
sushanth0910: updated code to log endpoint instead of global/regional

pkg/token/token.go
```diff
@@ -564,14 +567,16 @@ func (v tokenVerifier) Verify(token string) (*Identity, error) {
 	req.Header.Set(clusterIDHeader, v.clusterID)
 	req.Header.Set("accept", "application/json")
 
+	stsRegion := getStsRegion(parsedURL.Host)
```
micahhausler, line 570: This is not safe to parse before executing the request. We need to validate this is only parsed after a successful signature response.

micahhausler: ah I see we've already done host verification. Move this call to just after that to make it more clear.

Also:
* Can you make your function more robust against bad inputs? (ex: hosts without a period ) I know there is host verification before this, but bad things can happen and your code here should not depend on  ordering if we're going to log/metric this stuff before request execution
* Can you write a test for your `getStsRegion` function?

micahhausler (approved):
/lgtm 
/approve

PR #772, opened 2024-10-24
Log level setting + refactor filecache logs to use logrus

### LINKED ISSUE #771: [Feature request]: Quiet mode
### What would you like to be added?

It would be nice to give the user the ability to hide common messages like these:
```
No cached credential available.  Refreshing...
Updated cached credential
```

### Why is this needed?

I use `kubectl foreach` a lot and I'm spammed by those messages I don't really care about as long as it works fine.

I think we should be able to change the verbosity without modifying the `~/.kube/config`, so it should be configured from an environment variable like `AWS_IAM_AUTHENTICATOR_QUIET`.

gdlx:
<del>It adds a `--quiet` global flag, defaulted to the `AWS_IAM_AUTHENTICATOR_QUIET` boolean environment variable.</del>

<del>It currently only removes non error (but still written into stderr) filecache related logs.
Maybe some other logs could be concerned.</del>

It adds a `--log-verbosity` flag to configure the minimum log level to display and fixes the filecache raw prints into using logrus like other logs.

<del>I've also enabled the viper AutomaticEnv mode, which cleanly handles env vars with a common prefix.
The undocumented `KUBERNETES_AWS_AUTHENTICATOR_CLUSTER_ID` env flag should probably be refactored, but I didn't want to introduce a breaking change.</del>

I've also moved the following cache logs to the debug level, because the default cache output should be clean when no problem occurs. One still can display them by using the new verbosity flag when debugging the cache behavior:

```
DEBU[2025-06-02T15:01:32+02:00] No cached credential available. Refreshing...
DEBU[2025-06-02T15:01:33+02:00] Updated cached credential
```

Fixes #771

### DIFF
--- a/cmd/aws-iam-authenticator/init.go
+++ b/cmd/aws-iam-authenticator/init.go
@@ -21,6 +21,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"sigs.k8s.io/aws-iam-authenticator/pkg"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
@@ -75,6 +76,7 @@ var initCmd = &cobra.Command{
 func init() {
 	viper.AutomaticEnv()
 	viper.SetEnvPrefix("aws_iam_authenticator")
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
 	initCmd.Flags().String(
 		"hostname",
--- a/cmd/aws-iam-authenticator/init.go
+++ b/cmd/aws-iam-authenticator/init.go
@@ -73,6 +73,9 @@ var initCmd = &cobra.Command{
 }
 
 func init() {
+	viper.AutomaticEnv()
+	viper.SetEnvPrefix("aws_iam_authenticator")
+
 	initCmd.Flags().String(
 		"hostname",
 		"localhost",
--- a/cmd/aws-iam-authenticator/init.go
+++ b/cmd/aws-iam-authenticator/init.go
@@ -21,7 +21,6 @@ package main
 import (
 	"fmt"
 	"os"
-	"strings"
 
 	"sigs.k8s.io/aws-iam-authenticator/pkg"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
@@ -74,10 +73,6 @@ var initCmd = &cobra.Command{
 }
 
 func init() {
-	viper.AutomaticEnv()
-	viper.SetEnvPrefix("aws_iam_authenticator")
-	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
-
 	initCmd.Flags().String(
 		"hostname",
 		"localhost",
--- a/cmd/aws-iam-authenticator/root.go
+++ b/cmd/aws-iam-authenticator/root.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/mapper"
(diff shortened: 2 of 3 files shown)


gdlx:
Hi @jyotimahapatra @wongma7 !
Happy New Year !
Any chance the PR could be reviewed soon ? 😬

pkg/filecache/filecache.go
```diff
@@ -211,11 +212,15 @@ func (f *FileCacheProvider) Retrieve() (credentials.Value, error) {
...
 // with an expiration time.
 func (f *FileCacheProvider) RetrieveWithContext(ctx context.Context) (credentials.Value, error) {
+	var quiet bool = viper.GetBool("quiet")
+
 	if !f.cachedCredential.Expired() && f.cachedCredential.HasKeys() {
 		// use the cached credential
 		return V2CredentialToV1Value(f.cachedCredential), nil
 	} else {
-		_, _ = fmt.Fprintf(os.Stderr, "No cached credential available.  Refreshing...\n")
+		if !quiet {
```
kmala, line 221: i am not sure if we should be adding if else conditions and may be we should use klog or something similar and have these at default log level. If quiet we can set the klog to just print errors? Because if we add more logs, with this approach we need to add more if conditions. what do you think?

gdlx: I opted for the simplest option since only two strings were involved.
I can try to switch to klog, but it's a wider change.
Would you accept a PR that only migrates those 2 strings into klog with possible further refactoring of other logs? Because I don't feel comfortable testing all the log strings.

gdlx: I've just seen sirupsen/logrus is already used in the code. I'll try to use it instead of klog.
https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/cmd/aws-iam-authenticator/root.go#L28C23-L28C29

gdlx: I've updated the PR by using Logrus, which is the most obvious choice as it's used everywhere else in the code.

I should have seen that the actual issue was that filecache wasn't using Logrus.

I'm not sure if it can be considered a breaking change, but filecache logs will now include the level and timestamp from the Logrus formatter:

```
aws-iam-authenticator token --cache -i mycluster
INFO[2025-04-16T16:31:52+02:00] No cached credential available. Refreshing... 
INFO[2025-04-16T16:31:53+02:00] Updated cached credential                    
{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1beta1","spec":{
```

gdlx: Hi @kmala,
Does my change from last month fit your recommendation ?
Thanks !

cmd/aws-iam-authenticator/root.go
```diff
@@ -59,6 +60,9 @@ func init() {
 
 	rootCmd.PersistentFlags().StringP("log-format", "l", "text", "Specify log format to use when logging to stderr [text or json]")
 
+	rootCmd.PersistentFlags().String("log-level", "info", "Specify log level to use when logging to stderr ["+strings.Join(getValidLogLevels(), ", ")+"]")
```
kmala, line 63: nit: can we use StringP with shorthand as `v` ?

cmd/aws-iam-authenticator/init.go
```diff
@@ -73,6 +74,10 @@ var initCmd = &cobra.Command{
 }
 
 func init() {
+	viper.AutomaticEnv()
+	viper.SetEnvPrefix("aws_iam_authenticator")
```
kmala, line 78: i think this will make https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/cmd/aws-iam-authenticator/root.go#L68 not to work. can we not have the automatic env and use bind for the ones we need ?

cmd/aws-iam-authenticator/root.go
gdlx: Sure I can add a shorthand, but wouldn't it be clearer to rename the parameter `log-verbosity` or `verbosity`for clarity ?

cmd/aws-iam-authenticator/init.go
gdlx: I'll test if `KUBERNETES_AWS_AUTHENTICATOR_CLUSTER_ID` doesn't work anymore.
As I understand it, automatic env doesn't affect paramters with hard coded env var name.

kmala: i think automatic env overrides the specific one https://github.com/spf13/viper/blob/1508a7ba4478e59a385419023ed91eeeb8d522b4/viper.go#L1207-L1218

gdlx:
Hi @kmala,

I've just added the fixes you asked.

I've also changed (in a distinct commit) the log level to debug for the non problematic cache logs I initially opened the PR for:
```
DEBU[2025-06-02T15:01:32+02:00] No cached credential available. Refreshing...
DEBU[2025-06-02T15:01:33+02:00] Updated cached credential
```

I think normal cache behavior should be to have a clean output when no problem occur, unless we actually need to debug the cache behavior, which is now possible with the new verbosity flag.

Tell me if you want to revert this, but I think this is the default behavior expected by a large majority of the users.

Thanks !

gdlx:
> /lgtm /approve

Thanks !

PR #778, opened 2024-12-11
handle scenario when the file is created but doesn't have content

kmala:

If the dynamic file is present but does't have the correct/no data, then authenticator would be stuck in for loop waiting for any file event and until that happens would not have the latest changes. If this happens during the start then it would be failing requests until the file is updated and new changes are reloaded.

Fixes #

### DIFF
--- a/pkg/fileutil/util_test.go
+++ b/pkg/fileutil/util_test.go
@@ -1,11 +1,15 @@
 package fileutil
 
 import (
+	"context"
+	"fmt"
 	"os"
 	"strconv"
 	"sync"
 	"testing"
 	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var origFileContent = `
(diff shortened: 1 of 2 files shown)


pkg/fileutil/util_test.go
```diff
@@ -75,27 +95,30 @@ func TestLoadDynamicFile(t *testing.T) {
...
-				"Unexpected result: TestLoadDynamicFile: got: %s, wanted %s",
-				testA.expectedContent,
-				c.want,
-			)
-		}
-		testA.mutex.Unlock()
+		updateFile(f.Name(), c.input, t)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+		go wait.Until(func() {
+			testA.mutex.Lock()
```
wongma7, line 101: what does the goroutine do, won't it race with the next test case? like a) update abcd; b) poll every 100ms for 1s for DyamicFile to read abc;  c) update fgh: won't b and c overlap?

(since the test passes I think I'm misunderstanding something)

wongma7: or maybe it's passing because the goroutine gets canceled immediately by cancel()

wongma7:
/lgtm
ty

PR #785, opened 2024-12-13
deps: Update `golang.org/x/crypto` library to remediate high CVE

bryantbiggs:
- High CVE per https://github.com/advisories/GHSA-v778-237x-gjrc
- Causing vulnerability check to fail dep update PR https://github.com/kubernetes-sigs/aws-iam-authenticator/actions/runs/12321586343/job/34393314475?pr=783

Fixes #


kmala:
we shouldn't be impacted as we are not using the PublicKeyCallback but we should be okay to update the dependencies.
FYI on k/k repo https://github.com/kubernetes/kubernetes/issues/129164

PR #812, opened 2025-01-31
release authenticator from mainline with 0.7.0

sushanth0910:
release authenticator from mainline with tag 0.7.0

Fixes #


version.txt
```diff
@@ -1 +1 @@
-dev
+0.7.0
```
xdu31, line 1: need a fix on https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/.github/workflows/create-release.yml#L7

PR #844, opened 2025-04-25
v1.33.0 Release

Ganiredi:

- Added endpointslices as an indirect dependency  
- Updated v1.32.0/v0.32.0 to v1.33.0/v0.33.0  
- Ran tests locally

Fixes #


go.mod
```diff
@@ -7,20 +7,20 @@ require (
...
-	k8s.io/apimachinery v0.32.0
-	k8s.io/client-go v0.32.0
-	k8s.io/code-generator v0.32.0
-	k8s.io/component-base v0.32.0
+	k8s.io/api v0.33.0
+	k8s.io/apimachinery v0.33.0
+	k8s.io/client-go v0.33.0
+	k8s.io/code-generator v0.33.0
+	k8s.io/component-base v0.33.0
 	k8s.io/sample-controller v0.26.1
```
bryantbiggs, line 24: if we update this dependency, then dependabot will be able to automatically handle most of the dependency updates. updating `k8s.io/sample-controller` is not trivial, but it needs to be taken care of given how far behind its falling

Ganiredi: https://pkg.go.dev/k8s.io/sample-controller/pkg/apis/samplecontroller?tab=versions
Updated to match with latest.

bryantbiggs: thanks for updating that @Ganiredi !

PR #856, opened 2025-05-28
Partial upgrade of AWS SDK Go packages from V1 to V2

gargipanatula:
Upgrades some packages of the AWS SDK Go from V1 to V2. The AWS SDK Go V1 is being deprecated on July 31, 2025: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/. 

This PR upgrades all packages except the `github.com/aws/aws-sdk-go/aws/endpoints` package, which is used in partition verification. Given that upgrading that particular functionality will require larger changes, its changes will be pushed out in a separate PR. 

Testing:
Ran unit and e2e tests via `make test` and `make e2e`. Manually checked that token caching worked by checking that the value of `X-Amz-Security-Token` remained unchanged when using cached tokens within the 15 minute expiry window. Some output has been truncated.
```
(25-06-05 22:07:08) <0> [/aws-iam-authenticator]  
panatula % AWS_PROFILE=dev /aws-iam-authenticator/_output/bin/aws-iam-authenticator token --cache -i panatulacontrol --token-only

k8s-aws-v1.aHR0...QxZQ

(25-06-05 22:07:14) <0> [/aws-iam-authenticator]  
panatula % echo "aHR0...zQxZQ"
 | base64 -d
https://sts.us-west-2.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=XXXXXXXXXXXXXXXX%2F20250605%2Fus-west-2%2Fsts%2Faws4_request&X-Amz-Date=20250605T220714Z&X-Amz-Expires=60&X-Amz-Security-Token=IQoJb3JpZ2lu...FxIB7vxw7hxSINB5PTuY%3D&X-Amz-SignedHeaders=host%3Bx-k8s-aws-id&X-Amz-Signature=98d0...741ebase64: invalid input

(25-06-05 22:07:25) <1> [/aws-iam-authenticator]  
panatula % AWS_PROFILE=dev /aws-iam-authenticator/_output/bin/aws-iam-authenticator token --cache -i panatulacontrol --token-only 
k8s-aws-v1.aHR0...YwZg

(25-06-05 22:07:28) <0> [/aws-iam-authenticator]  
panatula % echo "aHR0...YwZg"
| base64 -d
https://sts.us-west-2.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=XXXXXXXXXXXXXXXX%2F20250605%2Fus-west-2%2Fsts%2Faws4_request&X-Amz-Date=20250605T220728Z&X-Amz-Expires=60&X-Amz-Security-Token=IQoJb3JpZ2lu...FxIB7vxw7hxSINB5PTuY%3D&X-Amz-SignedHeaders=host%3Bx-k8s-aws-id&X-Amz-Signature=8ffe...4f0fbase64: invalid input
```

Addresses #736, but does not fully close it.


pkg/token/token.go
```diff
@@ -226,20 +227,19 @@ func (g generator) GetWithOptions(options *GetTokenOptions) (Token, error) {
 
 	// create a session with the "base" credentials available
 	// (from environment variable, profile files, EC2 metadata, etc)
-	sess, err := session.NewSessionWithOptions(session.Options{
-		AssumeRoleTokenProvider: StdinStderrTokenProvider,
-		SharedConfigState:       session.SharedConfigEnable,
```
gargipanatula, line 231: No longer setting `SharedConfigState`, as its behavior when set to `SharedConfigEnable` is now default in V2:

From the V1 docs (https://docs.aws.amazon.com/sdk-for-go/api/aws/session/): 
By default NewSession will only load credentials from the shared credentials file (~/.aws/credentials). If the AWS_SDK_LOAD_CONFIG environment variable is set to a truthy value the Session will be created from the configuration values from the shared config (~/.aws/config) and shared credentials (~/.aws/credentials) files. Using the NewSessionWithOptions with SharedConfigState set to SharedConfigEnable will create the session as if the AWS_SDK_LOAD_CONFIG environment variable was set.

V2 docs (https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config):
The config package will load configuration from environment variables, AWS shared configuration file (~/.aws/config), and AWS shared credentials file (~/.aws/credentials).
Use the LoadDefaultConfig to load configuration from all the SDK's supported sources, and resolve credentials using the SDK's default credential chain.

pkg/ec2provider/ec2provider.go
```diff
@@ -296,12 +305,23 @@ func applySTSRequestHeaders(stsClient *sts.STS, sourceARN string) *sts.STS {
...
+			return stack.Build.Add(smithymiddleware.BuildMiddlewareFunc("STSHeaderMiddleware", func(
+				ctx context.Context, in smithymiddleware.BuildInput, next smithymiddleware.BuildHandler,
+			) (smithymiddleware.BuildOutput, smithymiddleware.Metadata, error) {
+				req, ok := in.Request.(*smithyhttp.Request)
+				if ok {
+					req.Header.Set(headerSourceAccount, reqHeaders[headerSourceAccount])
+					req.Header.Set(headerSourceArn, reqHeaders[headerSourceArn])
+				}
+				return next.HandleBuild(ctx, in)
+			}), smithymiddleware.Before)
```
gargipanatula, line 319: Note: V2 now uses middleware to add headers, added a test for this.

pkg/filecache/converter.go
```diff
@@ -1,55 +0,0 @@
...
-	"github.com/aws/aws-sdk-go/aws/credentials"
-)
-
-type v2 struct {
-	creds *credentials.Credentials
-}
-
-var _ aws.CredentialsProvider = &v2{}
-
-func (p *v2) Retrieve(ctx context.Context) (aws.Credentials, error) {
```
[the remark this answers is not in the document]
gargipanatula: This was used to convert between V1 and V2 credentials, and now that we've fully switched to V2, we can get rid of it.

Initial commit that added this file: https://github.com/kubernetes-sigs/aws-iam-authenticator/commit/9cdf38dc8f74fbf32f85a4d3241657e2868eabb6

pkg/ec2provider/ec2provider.go
```diff
@@ -76,26 +81,24 @@ func New(roleARN, sourceARN, region string, qps int, burst int) EC2Provider {
...
-	sess := session.Must(session.NewSession())
-	sess.Handlers.Build.PushFrontNamed(request.NamedHandler{
-		Name: "authenticatorUserAgent",
-		Fn: request.MakeAddToUserAgentHandler(
-			"aws-iam-authenticator", pkg.Version),
-	})
-	if aws.StringValue(sess.Config.Region) == "" {
-		sess.Config.Region = aws.String(region)
+func newEC2Client(roleARN, sourceARN, region string, qps int, burst int) EC2API {
+	ctx := context.Background()
```
kmala, line 92: can you get this as input param as the caller stack has it

pkg/ec2provider/ec2provider.go
```diff
@@ -109,16 +112,22 @@ func newSession(roleARN, sourceARN, region string, qps int, burst int) *session.
 			logrus.Errorf("Getting error = %s while creating rate limited client ", err)
 		}
 
-		stsClient := applySTSRequestHeaders(sts.New(sess, aws.NewConfig().WithHTTPClient(rateLimitedClient).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)), sourceARN)
-		ap := &stscreds.AssumeRoleProvider{
-			Client:   stsClient,
-			RoleARN:  roleARN,
-			Duration: time.Duration(60) * time.Minute,
+		stsCfg, err := config.LoadDefaultConfig(ctx,
+			config.WithRegion(region),
```
kmala, line 116: this should be cfg.region right?

kmala: also can we copy/duplicate the cfg and override the http client instead of LoadDefaultConfig again as if something is changed in the original cfg, it has to be done here also?

pkg/ec2provider/ec2provider.go
```diff
@@ -197,17 +206,17 @@ func (p *ec2ProviderImpl) GetPrivateDNSName(id string) (string, error) {
 	logrus.Infof("Calling ec2:DescribeInstances for the InstanceId = %s ", id)
 	metrics.Get().EC2DescribeInstanceCallCount.Inc()
 	// Look up instance from EC2 API
-	output, err := p.ec2.DescribeInstances(&ec2.DescribeInstancesInput{
-		InstanceIds: aws.StringSlice([]string{id}),
+	output, err := p.ec2.DescribeInstances(context.Background(), &ec2.DescribeInstancesInput{
```
kmala, line 209: can we get ctx as param?

pkg/ec2provider/ec2provider_test.go
```diff
@@ -191,3 +195,97 @@ func TestGetSourceAcctAndArn(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		args    map[string]string
+		want    map[string]string
+		wantErr bool
+	}{
+		{
+			name: "header with source arn",
+			args: map[string]string{
+				headerSourceAccount: "123456789012",
```
kmala, line 209: i don't see `headerSourceAccount` is used, can we remove it?

pkg/server/server.go
```diff
@@ -482,7 +489,7 @@ func (h *handler) renderTemplate(template string, identity *token.Identity) (str
 		if !instanceIDPattern.MatchString(identity.SessionName) {
 			return "", fmt.Errorf("SessionName did not contain an instance id")
 		}
-		privateDNSName, err := h.ec2Provider.GetPrivateDNSName(identity.SessionName)
+		privateDNSName, err := h.ec2Provider.GetPrivateDNSName(context.Background(), identity.SessionName)
```
kmala, line 492: can we get the ctx passed as a param since it called from the handler which already has it?

pkg/token/token.go
```diff
@@ -219,27 +221,34 @@ func StdinStderrTokenProvider() (string, error) {
...
-			"aws-iam-authenticator", pkg.Version),
-	})
+	cfg.APIOptions = append(cfg.APIOptions,
+		middleware.AddUserAgentKeyValue("aws-iam-authenticator", pkg.Version),
+	)
 	if options.Region != "" {
-		sess = sess.Copy(aws.NewConfig().WithRegion(options.Region).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint))
+		cfg.Region = options.Region
+	}
+	if cfg.Region == "" {
```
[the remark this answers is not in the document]
kmala: isn't the behavior same before in v1 that we set it as empty?

pkg/filecache/filecache.go
```diff
@@ -204,23 +203,23 @@ func NewFileCacheProvider(clusterID, profile, roleARN string, provider aws.Crede
 // Retrieve() implements the Provider interface, returning the cached credential if is not expired,
 // otherwise fetching the credential from the underlying Provider and caching the results on disk
 // with an expiration time.
-func (f *FileCacheProvider) Retrieve() (credentials.Value, error) {
-	return f.RetrieveWithContext(context.Background())
+func (f *FileCacheProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
```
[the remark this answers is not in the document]
kmala: did you get a chance to test the cache file provider feature?

pkg/ec2provider/ec2provider_test.go
```diff
@@ -191,3 +195,93 @@ func TestGetSourceAcctAndArn(t *testing.T) {
...
+		want    map[string]string
+		wantErr bool
+	}{
+		{
+			name: "header with source arn",
+			args: map[string]string{
+				headerSourceArn: "arn:aws:eks:us-east-1:123456789012:MyCluster/res1",
+			},
+			want: map[string]string{
+				headerSourceArn: "arn:aws:eks:us-east-1:123456789012:MyCluster/res1",
```
kmala, line 212: this should have the headerSourceAccount also right?

pkg/ec2provider/ec2provider_test.go
```diff
@@ -191,3 +195,93 @@ func TestGetSourceAcctAndArn(t *testing.T) {
...
+func TestApplySTSRequestHeaders(t *testing.T) {
+	tests := []struct {
+		name    string
+		args    map[string]string
+		want    map[string]string
+		wantErr bool
+	}{
+		{
+			name: "header with source arn",
+			args: map[string]string{
```
kmala, line 208: this can just be a string right?

pkg/ec2provider/ec2provider.go
```diff
@@ -258,8 +266,8 @@ func (p *ec2ProviderImpl) getPrivateDnsAndPublishToCache(instanceIdList []string
 	// Look up instance from EC2 API
 	logrus.Infof("Making Batch Query to DescribeInstances for %v instances ", len(instanceIdList))
 	metrics.Get().EC2DescribeInstanceCallCount.Inc()
-	output, err := p.ec2.DescribeInstances(&ec2.DescribeInstancesInput{
-		InstanceIds: aws.StringSlice(instanceIdList),
+	output, err := p.ec2.DescribeInstances(context.Background(), &ec2.DescribeInstancesInput{
```
kmala, line 269: Can we pass the context as the function param

pkg/ec2provider/ec2provider.go
```diff
@@ -109,16 +111,22 @@ func newSession(roleARN, sourceARN, region string, qps int, burst int) *session.
...
 
-		stsClient := applySTSRequestHeaders(sts.New(sess, aws.NewConfig().WithHTTPClient(rateLimitedClient).WithSTSRegionalEndpoint(endpoints.RegionalSTSEndpoint)), sourceARN)
-		ap := &stscreds.AssumeRoleProvider{
-			Client:   stsClient,
-			RoleARN:  roleARN,
-			Duration: time.Duration(60) * time.Minute,
+		stsCfg := cfg
+		stsCfg.HTTPClient = rateLimitedClient
+
+		if err != nil {
```
kmala, line 117: this err check is not needed right?

pkg/token/token.go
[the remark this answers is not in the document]
gargipanatula: After digging a bit, it looks like this actually stems from STS. In AWS SDK Go V1, if a region was not configured, it would fallback to the global endpoint. In V2, it causes the request to fail (documentation [here](https://docs.aws.amazon.com/sdkref/latest/guide/feature-sts-regionalized-endpoints.html#:~:text=typical%20global%20requests.-,SDK,-Supports%20setting))

Given this, I'm thinking we should panic if we can't find a region through imds, since it means that `GetCallerIdentity` will fail.

pkg/filecache/filecache.go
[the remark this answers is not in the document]
gargipanatula: I was just relying on the existing tests since there weren't major changes to the functionality. I might be missing something though, so if there's something we should test then I can take a look and add that in.

pkg/ec2provider/ec2provider_test.go
gargipanatula: Since we're working with multiple values, I personally find it easier to parse/use them if they're in a data structure. However, if it's more status quo to use a string for the args then I can take a look into that.

pkg/token/token.go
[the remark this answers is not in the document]
kmala: This is used as a CLI and hence i am not sure what would happen for the imds call if we run on our laptops. I think we should use global endpoint or fail the request in this case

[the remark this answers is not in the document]
kmala: how would imds work on your laptop because it has to make an api call to get the details https://github.com/aws/aws-sdk-go-v2/blob/feature/ec2/imds/v1.16.30/feature/ec2/imds/api_client.go#L59-L60?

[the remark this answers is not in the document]
gargipanatula: Oh got it - it's succeeding on my remote desktop but not my Mac. This is probably because my remote is running on an EC2 instance somewhere, so the IMDS will indeed fail for people on their local machine.

So, as you said, we can either error out here or use a default region, since otherwise we'll get errors from the GetCallerIdentity call. If we return an error, we risk downgrading the customer experience since in V1, it just fell back `us-east-1`. So, I'm leaning towards manually setting a region. Let me know if you have any objections, or use cases we should keep in mind, otherwise I'll try to get this working in the next commit.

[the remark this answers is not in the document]
kmala: if it fell to us-east-1 if no region is set, can we set the region to us-east-1 if not set or use global endpoint 
instead of using imds?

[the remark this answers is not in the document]
gargipanatula: Yeah I think we can do that. Going to try and validate on my end, will push up changes once I get that working.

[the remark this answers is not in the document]
gargipanatula: Update: added a fallback to use the global region (us-east-1) in case `cfg` doesn't have a region. Configuring usage of the global endpoint during endpoint resolution doesn't work since `cfg` requires a region to make requests, so I went with this option.

Let me know if there's a cleaner/less hardcoded way to go about this.
Commit: https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/856/commits/136628b42edf991e03eafb0f4e5edaef293a707d

PR #873, opened 2025-06-23
chore: Move observability dependencies to separate dependabot update group

bryantbiggs:
- Move observability dependencies to separate dependabot update group

Fixes #

- Not a fix per se, but breaks up the `misc` group a bit by separating out the observability packages which tend to cause disruptions/issues more frequently than one would hope https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_aws-iam-authenticator/870/pull-aws-iam-authenticator-integration/1937162696222838784

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -33,11 +33,17 @@ updates:
         patterns:
           - 'k8s.io*'
           - 'sigs.k8s.io*'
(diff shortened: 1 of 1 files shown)


bryantbiggs:
cc @kmala

PR #875, opened 2025-06-25
[do not merge] Upgrade token verification to AWS SDK Go V2

gargipanatula:
Updates the way that the server discovers valid partitions and regions during token verification. In essence, the host of a token is valid if it 1) comes from the same partition as the server and 2) contains a valid region.

Functionality changes
* This weakens the verification check, as it does not check that host regions are supported by STS. 

Implementation changes
* Listing partitions through the  `github.com/aws/aws-sdk-go/aws/endpoints` package → using a hardcoded list of partitions
* Discovering regions through the `github.com/aws/aws-sdk-go/aws/endpoints` package → discovering regions through the `ec2.DescribeRegions` API. 
    * The call yields all available regions to an account, regardless of opt-in status.

Testing
* Tested that the verify command still a) correctly verifies a token from the same partition and b) correctly denies tokens from different partitions

Created a token using credentials from an account in the `aws` partition (`US_TOKEN`) and an account in the `aws-cn` partition (`CHINA_TOKEN`)

Ran `verify` on each token using accounts from within their partition, one with the same and one with a different region than the account that created the token. I was able to see that the token was correctly successfully verified when an account is in the same partition as the one that generated the token, regardless of region.

Output of this test for `CHINA_TOKEN` (similar for `US_TOKEN`):
```
&{ARN:arn:aws-cn:sts::<account id>:assumed-role/Admin/<user> CanonicalARN:arn:aws-cn:iam::<account id>:role/<role> AccountID:<account id> UserID:<user-id> SessionName:<user> AccessKeyID:<access-key-id> STSEndpoint:sts.cn-northwest-1.amazonaws.com.cn}
```

Then, I ran `verify` on each token with both the correct and incorrect partition (`aws` for `CHINA_TOKEN` and `aws-cn` for `US_TOKEN`) to replicate conditions where the client is in a different partition from the server. I saw that the command correctly rejected the tokens when coming from a different partition.

Output for `US_TOKEN` (similar for `CHINA_TOKEN`):
```
# same partition (aws)
(25-06-24 23:47:21) <0> [/aws-iam-authenticator]
dev % /aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=<id> —token=$US_TOKEN —partition=aws
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
&{ARN:arn:aws:sts::<account-id>:assumed-role/admin/<role> CanonicalARN:arn:aws:iam::<account-id>:role/<role> AccountID:<account-id> UserID:<user-id> SessionName:<session-nmae> AccessKeyID:<key-id> STSEndpoint:sts.us-west-2.amazonaws.com}

# different partition (aws-cn)
(25-06-24 23:47:33) <0> [/aws-iam-authenticator]
dev % /aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=<id> —token=$US_TOKEN —partition=aws-cn
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
could not verify token: input token was not properly formatted: unexpected hostname "sts.us-west-2.amazonaws.com" in pre-signed URL
```
Note that the warning about ec2imds comes from the fact that the `AWS_EC2_METADATA_DISABLED` env var is set to true, if set to false the output is the same but without the warning.

Fixes #


pkg/server/server.go
```diff
@@ -215,9 +216,10 @@ func (c *Server) getHandler(ctx context.Context, backendMapper BackendMapper, ec
 	} else {
 		instanceRegion = instanceRegionOutput.Region
 	}
+	ec2Client := ec2.NewFromConfig(cfg)
```
kmala, line 219: are we creating new client instead of using ec2provider because we need to make call against the account where the authenticator is running instead of ServerEC2DescribeInstancesRoleARN ?

cmd/aws-iam-authenticator/verify.go
```diff
@@ -79,6 +84,24 @@ var verifyCmd = &cobra.Command{
 	},
 }
 
+// Uses EC2 metadata to get the region. Returns "" if no region found.
+func getInstanceRegion(ctx context.Context) string {
+	cfg, err := config.LoadDefaultConfig(ctx)
+	if err != nil {
+		fmt.Printf("[Warn] Unable to create metadata client, err: %v", err)
```
kmala, line 91: nit: the error message is not correct right as we just generating config here  and not creating client?

cmd/aws-iam-authenticator/verify.go
```diff
@@ -54,14 +56,17 @@ var verifyCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		sess := session.Must(session.NewSession())
```
kmala, line 57: Here we are expecting to have the session object right? But in getInstanceRegion function its best effort , any specific reason?

pkg/server/server.go
gargipanatula: Yeah, I figured configuring a client directly would be more straightforward than creating `ec2provider` with the host account's ARN.

cmd/aws-iam-authenticator/verify.go
gargipanatula: My mistake, this requirement shouldn't have been relaxed. I added a panic to `getInstanceRegion` that occurs when `config.LoadDefaultConfig` returns an error, to replicate the panic that `session.Must()` would've sent if `session.New()` had an error. ([session code](https://github.com/aws/aws-sdk-go/blob/v1.55.7/aws/session/session.go#L378)).

pkg/token/token.go
```diff
@@ -420,77 +422,66 @@ type tokenVerifier struct {
...
-		}
-	}
-	if partition == nil {
-		logrus.Errorf("Partition %s not valid", partitionID)
+	var tlds []string
+	serviceNames := []string{"sts", "sts-fips"}
+	switch partitionID {
+	case "aws":
+		tlds = []string{"amazonaws.com", "api.aws"}
+		validSTShostnames["sts.amazonaws.com"] = true
```
gargipanatula, line 433: todo: support global for all partitions

gargipanatula: update: it's only supported for the commercial partition, so this is actually not necessary

PR #876, opened 2025-06-27
Migrate STS hostname verification to AWS SDK Go V2

gargipanatula:
This PR removes calls to the `github.com/aws/aws-sdk-go/aws/endpoints` package. Due to the Go SDK making details about valid endpoints private, and to avoid hardcoding a list of supported regions, hostname verification will now use `sts.EndpointResolverV2` to verify hostnames. Note that this will make the verification process weaker, turning the check into making sure that a hostname was valid to simply making sure that the hostname follows the correct format and belongs to a valid AWS partition. However, since Amazon owns its partition's domains, any malformed hostnames will be rejected and bad actors should not be able to get their hostnames through. 

This is meant to move the repository away from the AWS SDK Go V1, which is set to be deprecated on July 31, 2025.

Tagging @kmala @yue9944882 and @micahhausler as this approach has been discussed with them.

Testing:
First, I tested that an account can generate a token and verify it. Checked that the token is verified when checked against the same partition, and rejected when not. Tested in accounts across two partitions (`aws` and `aws-cn`) to ensure this code maintains the general approach of partition verification.

In the US account:
```
export US_TOKEN=$(aws-iam-authenticator/_output/bin/aws-iam-authenticator token --cluster-id testcluster --token-only)

# same partition - expect success
(25-07-22 17:20:55) <0> [~/workplace/aws-iam-authenticator] 
dev-dsk % /aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=cluster —token=$US_TOKEN —partition=aws
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
&{ARN:arn:aws:sts::{account-id}:assumed-role/{role}/{session} CanonicalARN:arn:aws:iam::{account-id}:role/admin AccountID:{account-id} UserID:{user-id} SessionName:{session} AccessKeyID:{access-key-id} STSEndpoint:sts.us-west-2.amazonaws.com}

# different partition - expect failure
(25-07-22 17:20:59) <0> [~/workplace/aws-iam-authenticator] 
dev-dsk % /aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=cluster —token=$US_TOKEN —partition=aws-cn
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
could not verify token: input token was not properly formatted: partition {aws-cn} does not support hostname sts.us-west-2.amazonaws.com
```

In the China account:
```
export US_TOKEN=$(aws-iam-authenticator/_output/bin/aws-iam-authenticator token --cluster-id testcluster --token-only)

# different partition - expect failure
(25-07-22 17:20:55) <0> [~/workplace/aws-iam-authenticator] 
dev-dsk % aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=testcluster —token=$US_TOKEN —partition=aws
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
&{ARN:arn:aws:sts::{account-id}:assumed-role/{role}/{session} CanonicalARN:arn:aws:iam::{account-id}:role/admin AccountID:{account-id} UserID:{user-id} SessionName:{session} AccessKeyID:{access-key-id} STSEndpoint:sts.us-west-2.amazonaws.com}

# same partition - expect success
(25-07-22 17:20:59) <0> [~/workplace/aws-iam-authenticator] 
dev-dsk % aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=testcluster —token=$US_TOKEN —partition=aws-cn
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
could not verify token: input token was not properly formatted: partition {aws-cn} does not support hostname sts.us-west-2.amazonaws.com
```

Then, I tested verification of tokens that were not generated in the same partition. I verified that tokens can still be verified, provided that the verifier is given the correct region.

In the US account:
```
# partition doesn't match token's - expect failure
(25-07-22 17:26:02) <1> [~/workplace/aws-iam-authenticator]
dev-dsk %aws-iam-authenticator/_output/bin/aws-iam-authenticator verify --cluster-id=testcluster --token=$CHINA_TOKEN --partition=aws
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
could not verify token: input token was not properly formatted: partition {aws} does not support hostname sts.cn-north-1.amazonaws.com.cn

# partition matches token's - expect success
(25-07-22 17:26:04) <1> [~/workplace/aws-iam-authenticator] 
dev-dsk % aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=testcluster —token=$CHINA_TOKEN —partition=aws-cn
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
&{ARN:arn:aws-cn:sts::{cn-account-id}:assumed-role/ReadOnly/{session} CanonicalARN:arn:aws-cn:iam::{cn-account-id}:role/ReadOnly AccountID:{cn-account-id} UserID:{user-id} SessionName:{session} AccessKeyID:{access-key-id} STSEndpoint:sts.cn-north-1.amazonaws.com.cn}
```
In the China account:
```
# partition matches token's - expect success
(25-07-22 17:25:07) <0> [~/workplace/aws-iam-authenticator] 
dev-dsk % aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=testcluster —token=$US_TOKEN —partition=aws
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
&{ARN:arn:aws:sts::{account-id}:assumed-role/{role}/{session} CanonicalARN:arn:aws:iam::{account-id}:role/admin AccountID:{account-id} UserID:{user-id} SessionName:{session} AccessKeyID:{access-key-id} STSEndpoint:sts.us-west-2.amazonaws.com}

# partition doesn't match token's - expect failure
(25-07-22 17:25:09) <0> [~/workplace/aws-iam-authenticator] 
dev-dsk % aws-iam-authenticator/_output/bin/aws-iam-authenticator verify —cluster-id=testcluster —token=$US_TOKEN —partition=aws-cn
[Warn] Region not found in instance metadata, err: operation error ec2imds: GetRegion, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable
could not verify token: input token was not properly formatted: partition {aws-cn} does not support hostname sts.us-west-2.amazonaws.com
```

Fixes #

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -25,8 +25,11 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
+	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -36,7 +39,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
-	"github.com/aws/aws-sdk-go/aws/endpoints"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -46,6 +48,7 @@ import (
 	clientauthv1beta1 "k8s.io/client-go/pkg/apis/clientauthentication/v1beta1"
 	"sigs.k8s.io/aws-iam-authenticator/pkg"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/arn"
+	"sigs.k8s.io/aws-iam-authenticator/pkg/endpoints"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/filecache"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/metrics"
 )
@@ -418,75 +421,40 @@ type tokenVerifier struct {
 	client            *http.Client
 	clusterID         string
 	validSTShostnames map[string]bool
+	partition         string
+	region            string
+	mutex             sync.RWMutex
 }
 
-func getDefaultHostNameForRegion(partition *endpoints.Partition, region, service string) (string, error) {
-	rep, err := partition.EndpointFor(service, region, endpoints.STSRegionalEndpointOption, endpoints.ResolveUnknownServiceOption)
-	if err != nil {
-		return "", fmt.Errorf("Error resolving endpoint for %s in partition %s. err: %v", region, partition.ID(), err)
+// Returns the hostnames (regular and dualstack) for a service given a certain region and partition.
+// This hostname is not validated, but follows the format of what the hostname would be given
+// these parameters.
+func getDefaultHostNamesForRegion(partition, region, service string) ([]string, error) {
+	if !validateInputRegion(region) {
+		return []string{}, fmt.Errorf("invalid region identifier format provided: %s", region)
 	}
-	parsedURL, err := url.Parse(rep.URL)
+
+	hostnames := []string{}
+
+	partitionDomain, err := endpoints.GetSTSPartitionDomain(partition)
 	if err != nil {
-		return "", fmt.Errorf("Error parsing STS URL %s. err: %v", rep.URL, err)
+		return []string{}, fmt.Errorf("couldn't get domain for partition %s, %w", partition, err)
 	}
-	return parsedURL.Hostname(), nil
+	hostnames = append(hostnames, fmt.Sprintf("%s.%s.%s", service, region, partitionDomain))
+
+	dualStackPartitionDomain := endpoints.GetSTSDualStackPartitionDomain(partition)
+	if dualStackPartitionDomain != "" {
+		hostnames = append(hostnames, fmt.Sprintf("%s.%s.%s", service, region, dualStackPartitionDomain))
+	}
+
+	return hostnames, nil
 }
 
-func stsHostsForPartition(partitionID, region string) map[string]bool {
-	validSTShostnames := map[string]bool{}
-
-	var partition *endpoints.Partition
-	for _, p := range endpoints.DefaultPartitions() {
-		if partitionID == p.ID() {
-			partition = &p
-			break
-		}
-	}
-	if partition == nil {
-		logrus.Errorf("Partition %s not valid", partitionID)
-		return validSTShostnames
-	}
-
-	stsSvc, ok := partition.Services()[stsServiceID]
-	if !ok {
-		logrus.Errorf("STS service not found in partition %s", partitionID)
-		// Add the host of the current instances region if the service doesn't already exists in the partition
-		// so we don't fail if the service is not present in the go sdk but matches the instances region.
-		stsHostName, err := getDefaultHostNameForRegion(partition, region, stsServiceID)
-		if err != nil {
-			logrus.WithError(err).Error("Error getting default hostname")
-		} else {
-			validSTShostnames[stsHostName] = true
-		}
-		return validSTShostnames
-	}
-	stsSvcEndPoints := stsSvc.Endpoints()
-	for epName, ep := range stsSvcEndPoints {
-		rep, err := ep.ResolveEndpoint(endpoints.STSRegionalEndpointOption)
-		if err != nil {
-			logrus.WithError(err).Errorf("Error resolving endpoint for %s in partition %s", epName, partitionID)
-			continue
-		}
-		parsedURL, err := url.Parse(rep.URL)
-		if err != nil {
-			logrus.WithError(err).Errorf("Error parsing STS URL %s", rep.URL)
-			continue
-		}
-		validSTShostnames[parsedURL.Hostname()] = true
-	}
-
-	// Add the host of the current instances region if not already exists so we don't fail if the region is not
-	// present in the go sdk but matches the instances region.
-	if _, ok := stsSvcEndPoints[region]; !ok {
-		stsHostName, err := getDefaultHostNameForRegion(partition, region, stsServiceID)
-		if err != nil {
-			logrus.WithError(err).Error("Error getting default hostname")
-			return validSTShostnames
-		}
-		validSTShostnames[stsHostName] = true
-	}
-
-	return validSTShostnames
+// Ported over from the AWS SDK Go V1 endpoints package, which validated regions as part of endpoint resolution
+// https://github.com/aws/aws-sdk-go/blob/163aada692ed32951f979aacf452ded4c03b8a7c/aws/endpoints/v3model.go#L592
+func validateInputRegion(region string) bool {
+	regionValidationRegex := regexp.MustCompile(`^[[:alnum:]]([[:alnum:]\-]*[[:alnum:]])?$`)
+	return regionValidationRegex.MatchString(region)
 }
 
 // NewVerifier creates a Verifier that is bound to the clusterID and uses the default http client.
@@ -497,7 +465,7 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
 		metrics.InitMetrics(prometheus.NewRegistry())
 	}
 
-	return tokenVerifier{
+	return &tokenVerifier{
 		client: &http.Client{
 			CheckRedirect: func(req *http.Request, via []*http.Request) error {
 				return http.ErrUseLastResponse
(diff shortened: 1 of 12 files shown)


pkg/endpoints/hostList.go
```diff
@@ -0,0 +1,107 @@
...
+		AwsCnPartitionID,
+		AwsUsGovPartitionID,
+		AwsIsoPartitionID,
+		AwsIsoBPartitionID,
+		AwsIsoEPartitionID,
+		AwsIsoFPartitionID,
+	}
+
+	AWS_PARTITION_STS_HOSTS = []string{
+		"sts.amazonaws.com",
```
yue9944882, line 28: i think we shouldn't maintain these hardcoded variables in our repo. because it's hard to automate and refresh the list. how about we instead use the endpoint resolver construct from sdk v2 here?
>https://github.com/aws/aws-sdk-go-v2/blob/main/service/sts/endpoints.go#L333

gargipanatula: Oh I think I misread what you meant by resolving the endpoint - I see what you mean now. Let me try making an implementation that uses [ResolveEndpoint](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/endpoints#Resolver.ResolveEndpoint) to validate a hostname.

pkg/token/token.go
```diff
@@ -504,16 +477,103 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
...
+			return nil
+		}
+
 		return FormatError{fmt.Sprintf("unexpected hostname %q in pre-signed URL", host)}
 	}
+
+	// The hostname matches one of the hostnames from the resolver, which follows the format of
+	// sts.{region}.{suffix}.{more-suffix}. Verify that the suffix is appropriate for the given partition.
+	// E.g.: sts.us-west-2.amazonaws.com, sts.us-isob-east-1.sc2s.sgov.gov
+	parts := strings.Split(host, ".")
```
kmala, line 548: Why is this needed? resolver will find the right partition for a valid region right? For invalid region which resolver can't find correct partition, the expectation is that the hostname can't be resolved anyway right?

pkg/token/token.go
```diff
@@ -504,16 +477,103 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
...
 	}
 }
 
-// verify a sts host, doc: http://docs.amazonaws.cn/en_us/general/latest/gr/rande.html#sts_region
+// Ensures that a given host is following the appropriate STS hostname format, e.g. sts.{region}.{suffix}.
+// Does not run any validation, so hostnames that contain invalid regions (e.g. sts.not-a-region.amazonaws.com) or
+// regions where STS/FIPS/Dualstack are not supported will be verified.
+// doc: http://docs.amazonaws.cn/en_us/general/latest/gr/rande.html#sts_region
 func (v tokenVerifier) verifyHost(host string) error {
-	if _, ok := v.validSTShostnames[host]; !ok {
```
kmala, line 514: i think we should still store the value in cache such that this check can be quick and doesn't impact performance

pkg/token/token.go
```diff
@@ -504,16 +477,103 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
...
+			// If the resolver cannot find a hostname for the given region, it will fall back to the commerical
+			// format, resulting in something like sts.{region}.amazonaws.com. So, if the hostRegion is invalid
+			// or unsupported by STS, then we expect to see an endpoint that follows the commerical partition format.
+			endpoint, err := stsResolver.ResolveEndpoint(context.TODO(), sts.EndpointParameters{
+				Region:       aws.String(hostRegion),
+				UseFIPS:      aws.Bool(useFIPS),
+				UseDualStack: aws.Bool(useDualStack),
+			})
+			if err != nil {
+				logrus.Infof("Error resolving endpoint for FIPS=%v, DualStack=%v: %v", useFIPS, useDualStack, err)
```
kmala, line 522: We should not be logging at default level unless we are getting the endpoints for a region only once and caching because we already know this will fail in few scenarios like dualstack for iso partitions.

pkg/token/token.go
```diff
@@ -504,16 +477,103 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
...
+	}
+
+	stsResolver := sts.NewDefaultEndpointResolverV2()
+
+	// Generate all possible endpoints given this region
+	var resolvedHosts []string
+	options := []bool{true, false}
+	for _, useFIPS := range options {
+		for _, useDualStack := range options {
+			// If the resolver cannot find a hostname for the given region, it will fall back to the commerical
```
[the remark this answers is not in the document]
gargipanatula: Yep - I initially thought it was checking for valid hostnames but realized I can't actually validate that because all GA regions support STS as of now. Will change the comment to not imply this.

pkg/token/token.go
```diff
@@ -504,16 +477,103 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
...
+
+			resolvedHosts = append(resolvedHosts, endpoint.URI.Host)
+		}
+	}
+
+	if !slices.Contains(resolvedHosts, host) {
+		// If neither of them match the host, check what the hostname should be using the given region and partition.
+		// This is to account for regions that aren't yet supported by the SDK, but are
+		// present in the instance metadata.
+		defaultHostName, err := getDefaultHostNameForRegion(v.partition, v.region, stsServiceID)
```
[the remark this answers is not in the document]
gargipanatula: This is to account for non-GA regions in non-commercial partitions. Say we get the hostname `sts.cn-new-region.amazonaws.com.cn` where `cn-new-region` is currently being built. `resolvedHosts` will contain a list of hosts following the commercial format, since the Resolver would've not found that region and then fallen back to commercial. 

So, we need to add this fallback to allow non-commercial hostnames that follow the correct STS format, but might not be publicly available yet.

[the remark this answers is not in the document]
kmala: Each partition has region regex https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.go#L266 and a newly created region also will resolve to correct partition. If a new region is being created in an existing partition for which the regex has to be updated i think its okay to update sdk dependencies to support them as this is not a common scenario fwik?

pkg/token/token.go
```diff
@@ -418,75 +420,47 @@ type tokenVerifier struct {
...
+	case endpoints.AwsUsGovPartitionID:
+		endpointHost = "amazonaws.com"
+	case endpoints.AwsIsoPartitionID:
+		endpointHost = "c2s.ic.gov"
+	case endpoints.AwsIsoBPartitionID:
+		endpointHost = "sc2s.sgov.gov"
+	case endpoints.AwsIsoEPartitionID:
+		endpointHost = "cloud.adc-e.uk"
+	case endpoints.AwsIsoFPartitionID:
+		endpointHost = "csp.hci.ic.gov"
```
yue9944882, line 452: concentrate these suffix hard-coding to `pkg/endpoints/partitions.go`

pkg/token/token.go
gargipanatula: You're right, but in addition this is also for the case where the hostname is valid, but the partition doesn't match that of the verifier's. 

Say we get `sts.cn-north-1.amazonaws.com.cn/`. Even though the resolver would (correctly) give out `sts.cn-north-1.amazonaws.com.cn` , we still need to make sure that the hostname matches the partition that the verifier was initialized with.

If the verifier was initialized with `aws-cn` then this is fine, however if it was `aws-iso-b` then the token should be rejected.

TLDR the hostname can be valid, but it should still match the partition parameter.

gargipanatula: I'll include the multiple suffixes case in this logic, though, thanks for the callout.

pkg/token/token.go
```diff
@@ -505,15 +479,118 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
...
+		expectedSuffix = "sc2s.sgov.gov"
+	case endpoints.AwsIsoEPartitionID:
+		expectedSuffix = "cloud.adc-e.uk"
+	case endpoints.AwsIsoFPartitionID:
+		expectedSuffix = "csp.hci.ic.gov"
+	default:
+		return fmt.Errorf("Partition %s not valid", v.partition)
+	}
+
+	if strings.Join(parts[2:], ".") != expectedSuffix {
```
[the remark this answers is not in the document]
gargipanatula: True, it can change in the future. For now, STS has documented that this is the format it will use - would this make it safe to make this assumption? 
https://docs.aws.amazon.com/sdkref/latest/guide/feature-sts-regionalized-endpoints.html

pkg/token/token.go
```diff
@@ -418,75 +421,36 @@ type tokenVerifier struct {
...
-		stsHostName, err := getDefaultHostNameForRegion(partition, region, stsServiceID)
-		if err != nil {
-			logrus.WithError(err).Error("Error getting default hostname")
-			return validSTShostnames
-		}
-		validSTShostnames[stsHostName] = true
-	}
+	return []string{
+		fmt.Sprintf("%s.%s.%s", service, region, partitionDomain),          // hostname
+		fmt.Sprintf("%s.%s.%s", service, region, dualStackPartitionDomain), // dual-stack hostname
```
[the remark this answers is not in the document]
gargipanatula: Ah true, I'll fix that and only return the dual stack hostname if its supported by the partition.

pkg/token/token.go
```diff
@@ -497,30 +465,131 @@ func NewVerifier(clusterID, partitionID, region string) Verifier {
...
+		}
+	}
+
+	if !slices.Contains(resolvedHosts, host) {
+		// If neither of them match the host, check what the hostname should be using the given region and partition.
+		// This is to account for regions that aren't yet supported by the SDK, but are
+		// present in the instance metadata.
+		defaultHostnames, err := getDefaultHostNamesForRegion(v.partition, v.region, stsServiceID)
+		if err != nil {
+			logrus.Infof("Error resolving default hostnames for partition %s, region %s, %v", v.partition, v.region, err)
```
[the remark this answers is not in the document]
gargipanatula: True, I can add that in for readability

PR #905, opened 2025-08-28
Update go dependencies with 1.34.0

Ganiredi:

Fixes #


hack/patches/etcd-tcp-compatibility.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/test/integration/framework/etcd.go b/test/integration/framework/etcd.go
+index 1c0f6699..458aca28 100644
+--- a/test/integration/framework/etcd.go
```
kmala, line 3: why do we want to create a patch ? why can't we use unix socket in the integration tests

PR #994, opened 2026-03-04
Support wildcard patterns in RoleARN mappings

AndreKurait:
## What

Extend `RoleMapping.Matches()` to use `ArnLike()` for wildcard matching when the `RoleARN` contains `*` or `?` characters.

## Why

The `ArnLike` matching infrastructure already exists in the codebase (used for SSO role matching via `SSOArnLike()`) but is not wired up for regular `RoleARN` mappings, which only support exact string matching via `strings.EqualFold()`.

This means there is no way to map a pattern of roles (e.g., all roles with a `dev-` prefix, or all roles in an account) without enumerating each one individually. The SSO feature (issue #333) proved the `ArnLike` approach works — this PR generalizes it to regular role mappings.

## Example

```yaml
mapRoles:
- rolearn: "arn:aws:iam::012345678912:role/dev-*"
  username: "dev-{{SessionName}}"
  groups:
  - developers
- rolearn: "arn:aws:iam::012345678912:role/*"
  username: "{{SessionName}}"
  groups:
  - all-roles
```

## Changes

- **`pkg/config/mapper.go`**: `RoleMapping.Matches()` now delegates to `arn.ArnLike()` when the `RoleARN` contains wildcard characters, falling back to exact match for non-wildcard ARNs (no behavior change for existing configs)
- **`pkg/mapper/file/mapper.go`**: `NewFileMapper` skips `arn.Canonicalize()` for wildcard `RoleARN`s (the AWS ARN parser rejects wildcards) and stores them lowercased instead
- **Tests**: Added for both unit (`Matches`) and integration (`FileMapper.Map`) levels covering prefix wildcards, full wildcards, cross-account rejection, and cross-resource-type rejection

## Backward Compatibility

- Non-wildcard `RoleARN` mappings are completely unchanged (exact `strings.EqualFold` path)
- No new feature flags required — wildcard detection is automatic based on the presence of `*` or `?` in the configured `RoleARN`
- The `ArnLike` function is already well-tested in the existing SSO code path

## Testing

All existing tests pass. New tests added:
- `TestWildcardRoleARNMapping` — full wildcard `role/*`, cross-account rejection, cross-resource-type rejection
- `TestWildcardRoleARNPrefixMapping` — prefix wildcard `role/dev-*`
- `TestWildcardRoleMap` — end-to-end through `FileMapper.Map()`

### DIFF
--- a/pkg/config/mapper.go
+++ b/pkg/config/mapper.go
@@ -68,10 +68,23 @@ func (m *RoleMapping) Validate() error {
 	return nil
 }
 
+// hasWildcards returns true if the string contains ArnLike wildcard characters
+func hasWildcards(s string) bool {
+	return strings.ContainsAny(s, "*?")
+}
+
 // Matches returns true if the supplied ARN or SSO settings matches
 // this RoleMapping
 func (m *RoleMapping) Matches(subject string) bool {
 	if m.RoleARN != "" {
+		if hasWildcards(m.RoleARN) {
+			ok, err := arn.ArnLike(subject, strings.ToLower(m.RoleARN))
+			if err != nil {
+				logrus.Error("Could not parse subject ARN for wildcard match: ", err)
+				return false
+			}
+			return ok
+		}
 		return strings.EqualFold(m.RoleARN, subject)
 	}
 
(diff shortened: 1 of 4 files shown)


AndreKurait:
@bryantbiggs did the test finish? Any other changes needed to get this in?

AndreKurait:
@bryantbiggs @micahhausler any thoughts?

PR #1047, opened 2026-05-29
fix: honor reservedPrefixConfig for ConfigMap and CRD backends

CaidenBorrego:

- panatula@ confirmed issue with mhausler@
- ISSUE: `reservedPrefixConfig` was silently ignored for the EKSConfigMap and CRD backends. `ConfigMapMapper.UsernamePrefixReserveList()` and `CRDMapper.UsernamePrefixReserveList()` hardcoded `[]string{} ` and never read `cfg.ReservedPrefixConfig`. The server enforces the reserved-prefix check only when the list is non-empty (server.go 487), so the protection was a no-op for these backends. MountedFile and DynamicFile honor it correctly.
- FIX: Wire `cfg.ReservedPrefixConfig[mapper.ModeEKSConfigMap]` and `cfg.ReservedPrefixConfig[mapper.ModeCRD] ` into the respective mappers, mirroring the existing MountedFile and DynamicFile pattern. Both mappers gain a `usernamePrefixReserveList` field populated in their constructor and returned from `UsernamePrefixReserveList()`.
- Added unit tests to test new functionality for both backends 
- `make test` and `make bin` pass
- IMPACT: clusters that configure `reservedPrefixConfig` for EKSConfigMap or CRD will now reject usernames matching the reserved prefixes. Any mapping that previously rendered to a reserved-prefix username on these backends will start failing authentication. EKS prod is unaffected (config not set). Self-managed deployments relying on the silent bypass will see those mappings break — that is the desired behavior and aligns with DynamicFile and MountedFile backend behavior.

### DIFF
--- a/pkg/mapper/configmap/configmap_test.go
+++ b/pkg/mapper/configmap/configmap_test.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/client-go/kubernetes/typed/core/v1/fake"
 	k8stesting "k8s.io/client-go/testing"
 	"sigs.k8s.io/aws-iam-authenticator/pkg/config"
+	"sigs.k8s.io/aws-iam-authenticator/pkg/mapper"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


version.txt
```diff
@@ -1 +1 @@
-0.7.16
+0.7.17
```
kmala: we usually have new releases done as a separate PR

gargipanatula:
Thanks for writing this up Caiden! Were you able to reproduce the issue on a cluster with and without the fix to confirm that your changes address the scenario? The unit tests are good but I want to make sure we test the flow end to end.

ronaldngounou (approved):

PR #1063, opened 2026-06-17
 feat: add STSPresignClient field to GetTokenOptions for custom credential injection

jitinchekka2:
 
`GetWithOptions` calls config.LoadDefaultConfig internally, which builds its own AWS config from the environment. There is no way to pass a pre-configured STS client, so callers who have already built credentials (e.g. named profiles, cross-account role chains, or custom endpoints) are forced to re-implement the token presigning logic
  themselves rather than using this library.

This pull request  adds an optional STSPresignClient STSPresignAPI field to GetTokenOptions. When set, GetWithOptions uses it directly and skips LoadDefaultConfig entirely. When nil, existing behavior is unchanged.

  The STSPresignAPI interface is narrow — a single PresignGetCallerIdentity method — and is satisfied by *sts.PresignClient without any adapter.

  The presigning logic is extracted from GetWithSTS into a private presignToken helper shared by both paths, so there is no duplication.

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -114,6 +114,12 @@ type Token struct {
 	Expiration time.Time
 }
 
+// STSPresignAPI is satisfied by *sts.PresignClient and allows callers to inject a
+// pre-configured client with custom credentials, region, or endpoint settings.
+type STSPresignAPI interface {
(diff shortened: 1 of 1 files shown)


jitinchekka2:
@hakuna-matatah @ronaldngounou Please review

PR #1066, opened 2026-06-23
Update for Go 1.26.5

SuperQ:
Update `.go-version` to latest patch release.
* Minimize the `go` directive to avoid unecessary minimum version bumping for downstream users.

Fixes #


SuperQ:
:eyeglasses: @kmala

tests/integration/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/aws-iam-authenticator/tests/integration
 
-go 1.26.4
+go 1.26.1
```
kmala: why 1.26.1 instead of 1.26.0?

SuperQ: I couldn't get `go mod tidy` to accept that. Not sure exactly why.

SuperQ: I figured out a solution, will add another commit.

go.work
```diff
@@ -1,4 +1,4 @@
-go 1.26.4
+go 1.26.1
```
kmala: Can we also change this one?

SuperQ: Yes, now that the integration go.mod is fixed we can.

PR #1084, opened 2026-07-15
Set a custom User-Agent on STS token verification requests

liwadman:

This PR has the authenticator server set a custom user agent when communicating to AWS STS, to make it a bit easier at a glance to understand if a given GetCallerIdentity request originates from an EKS cluster control plane.

No issue addressed by this.

### DIFF
--- a/pkg/token/token.go
+++ b/pkg/token/token.go
@@ -106,6 +106,9 @@ const (
 	kindExecCredential = "ExecCredential" //nolint:gosec // G101: false positive, this is a Kubernetes API kind name not a credential
 	execInfoEnvKey     = "KUBERNETES_EXEC_INFO"
 	stsServiceID       = "sts"
+	// userAgentPrefix is used to build the User-Agent header the verifier sends
+	// when validating pre-signed sts:GetCallerIdentity URLs against STS.
+	userAgentPrefix = "aws-iam-authenticator"
 )
 
 // Token is generated and used by Kubernetes client-go to authenticate with a Kubernetes cluster.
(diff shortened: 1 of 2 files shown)


liwadman:
The single finding from govulncheck exists outside of our PR. I believe the should be implemented the fix in a separate PR.

pkg/token/token.go
```diff
@@ -688,6 +691,7 @@ func (v *tokenVerifier) Verify(token string) (*Identity, error) {
 	}
 	req.Header.Set(clusterIDHeader, v.clusterID)
 	req.Header.Set("accept", "application/json")
+	req.Header.Set("User-Agent", fmt.Sprintf("%s/%s", userAgentPrefix, pkg.Version))
```
lyoung-confluent, line 694: Could we add the cluster ID as well so we can identify which specific cluster the token was passed to?
```suggestion
	req.Header.Set("User-Agent", fmt.Sprintf("%s/%s (%s)", userAgentPrefix, pkg.Version, v.clusterID))
```

pkg/token/token_test.go
```diff
@@ -447,6 +448,42 @@ func TestVerifyCanonicalARN(t *testing.T) {
...
+
+func TestVerifySetsUserAgent(t *testing.T) {
+	rt := &capturingRoundTripper{
+		body: jsonResponse("arn:aws:iam::123456789012:user/Alice", "123456789012", "Alice"),
+	}
+	verifier := &tokenVerifier{
+		client:            &http.Client{Transport: rt},
+		validSTShostnames: make(map[string]bool),
+		partition:         "aws",
+	}
```
lyoung-confluent, line 474:
```suggestion
		clusterID:         "test-cluster",
	}
```

pkg/token/token_test.go
```diff
@@ -447,6 +448,42 @@ func TestVerifyCanonicalARN(t *testing.T) {
...
+	if _, err := verifier.Verify(validToken); err != nil {
+		t.Fatalf("received unexpected error: %s", err)
+	}
+	if rt.lastRequest == nil {
+		t.Fatal("expected an STS request to be made but none was captured")
+	}
+	want := fmt.Sprintf("%s/%s", userAgentPrefix, pkg.Version)
```
lyoung-confluent, line 481:
```suggestion
	want := fmt.Sprintf("%s/%s (%s)", userAgentPrefix, pkg.Version, verifier.clusterID)
```

PR #1097, opened 2026-08-25
added ec2 metrics: ec2_responses_total and ec2_connection_failures_total

HanshalDabbiru:
Adds outcome metrics for the authenticator's EC2 DescribeInstances calls (used to resolve {{EC2PrivateDNSName}}). Currently only ec2_describe_instances_calls_total exists, which counts attempts but gives no insight into their results.

This adds two metrics, modeled on the existing STS instrumentation:

ec2_responses_total{ResponseCode, Region} — counts calls that got an HTTP response, labeled by status code (200, 400, 503, ...).
ec2_connection_failures_total{Region} — counts calls that got no response at all (timeout / unreachable / connection error), mirroring sts_connection_failures_total.
Together they let operators distinguish EC2 errors from EC2 being unreachable, spot throttling/regional issues, and compute error rates, matching the visibility already available for the STS path.

Testing: Unit testing. Built the image and deployed it to a beta EKS cluster, then curled the authenticator's /metrics endpoint to confirm the new ec2_responses_total and ec2_connection_failures_total metrics are registered and exported. Verified successful DescribeInstances calls increment ec2_responses_total{ResponseCode="200"} with the expected Region label.

### DIFF
--- a/pkg/ec2provider/ec2provider.go
+++ b/pkg/ec2provider/ec2provider.go
@@ -70,6 +70,21 @@ type ec2ProviderImpl struct {
 	privateDNSCache    *ec2PrivateDNSCache
 	ec2Requests        *ec2Requests
 	instanceIDsChannel chan string
+	region             string
+}
+
+// ec2ResponseCode returns the specific HTTP status code (e.g. "400"/"503") for
+// a non-nil DescribeInstances error that carried an HTTP response, or "" if the
+// error had no HTTP response / a zero status (a transport failure, which is not
+// counted here).
+func ec2ResponseCode(err error) string {
+	var respErr *smithyhttp.ResponseError
+	if errors.As(err, &respErr) {
+		if code := respErr.HTTPStatusCode(); code != 0 {
+			return fmt.Sprint(code)
+		}
+	}
+	return ""
 }
 
 // New creates and starts a new EC2Provider that resolves instance IDs to private DNS names.
@@ -87,6 +102,7 @@ func New(ctx context.Context, roleARN, sourceARN, region string, qps int, burst 
 		privateDNSCache:    dnsCache,
 		ec2Requests:        ec2Requests,
 		instanceIDsChannel: make(chan string, maxChannelSize),
+		region:             region,
 	}
 }
 
@@ -207,9 +223,15 @@ func (p *ec2ProviderImpl) GetPrivateDNSName(ctx context.Context, id string) (str
 		InstanceIds: []string{id},
 	})
 	if err != nil {
+		if code := ec2ResponseCode(err); code != "" {
+			metrics.Get().EC2Responses.WithLabelValues(code, p.region).Inc()
+		} else {
+			metrics.Get().EC2ConnectionFailure.WithLabelValues(p.region).Inc()
+		}
 		p.unsetRequestInFlightForInstanceID(id)
 		return "", fmt.Errorf("failed querying private DNS from EC2 API for node %s: %s ", id, err.Error())
 	}
+	metrics.Get().EC2Responses.WithLabelValues("200", p.region).Inc()
 	for _, reservation := range output.Reservations {
 		for _, instance := range reservation.Instances {
 			if aws.ToString(instance.InstanceId) == id {
@@ -269,7 +291,13 @@ func (p *ec2ProviderImpl) getPrivateDNSAndPublishToCache(ctx context.Context, in
 	})
 	if err != nil {
 		logrus.Errorf("Batch call failed querying private DNS from EC2 API for nodes [%s] : with error = []%s ", instanceIDList, err.Error())
+		if code := ec2ResponseCode(err); code != "" {
+			metrics.Get().EC2Responses.WithLabelValues(code, p.region).Inc()
+		} else {
+			metrics.Get().EC2ConnectionFailure.WithLabelValues(p.region).Inc()
+		}
 	} else {
+		metrics.Get().EC2Responses.WithLabelValues("200", p.region).Inc()
 		if output.NextToken != nil {
 			logrus.Debugf("Successfully got the batch result , output.NextToken = %s ", *output.NextToken)
 		} else {
--- a/pkg/ec2provider/ec2provider.go
+++ b/pkg/ec2provider/ec2provider.go
@@ -73,11 +73,7 @@ type ec2ProviderImpl struct {
 	region             string
 }
 
-// ec2ResponseCode returns the specific HTTP status code (e.g. "400"/"503") for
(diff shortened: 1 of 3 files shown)


HanshalDabbiru:
@CaidenBorrego

pkg/metrics/metrics.go
```diff
@@ -90,6 +92,20 @@ func createMetrics(reg prometheus.Registerer) Metrics {
 				Help:      "Sts responses with error code label",
 			}, []string{"ResponseCode", "StsRegion"},
 		),
+		EC2Responses: factory.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: Namespace,
+				Name:      "ec2_responses_total",
+				Help:      "EC2 responses with error code label",
```
adp-825, line 99: This also consists of 200s right? would it be fair to say "EC2 responses with error code label" ?

pkg/ec2provider/ec2provider.go
```diff
@@ -70,6 +70,21 @@ type ec2ProviderImpl struct {
...
 	ec2Requests        *ec2Requests
 	instanceIDsChannel chan string
+	region             string
+}
+
+// ec2ResponseCode returns the specific HTTP status code (e.g. "400"/"503") for
+// a non-nil DescribeInstances error that carried an HTTP response, or "" if the
+// error had no HTTP response / a zero status (a transport failure, which is not
+// counted here).
+func ec2ResponseCode(err error) string {
```
adp-825: Can we add one of two tests for this change?

pkg/ec2provider/ec2provider.go
```diff
@@ -70,6 +70,21 @@ type ec2ProviderImpl struct {
...
+	var respErr *smithyhttp.ResponseError
```
adp-825, line 77: nit: why are we parsing to smithy http type? I think this is because this is the error type returned by AWS sdk. But please double check.

pkg/metrics/metrics.go
HanshalDabbiru: I am following the same convention as the sts_responses_total which has the same description.

pkg/ec2provider/ec2provider.go
HanshalDabbiru: added.

pkg/ec2provider/ec2provider.go
HanshalDabbiru: double checked. will show images in doc.

pkg/ec2provider/ec2provider.go
```diff
@@ -70,6 +70,21 @@ type ec2ProviderImpl struct {
...
+	if errors.As(err, &respErr) {
+		if code := respErr.HTTPStatusCode(); code != 0 {
```
CaidenBorrego, line 79: What is this checking? FMU HTTP status codes are  >= 200, so a check of != 0 would always be true? Or in Go is this a nil check?

pkg/ec2provider/ec2provider.go
```diff
@@ -207,9 +223,15 @@ func (p *ec2ProviderImpl) GetPrivateDNSName(ctx context.Context, id string) (str
...
 	if err != nil {
+		if code := ec2ResponseCode(err); code != "" {
+			metrics.Get().EC2Responses.WithLabelValues(code, p.region).Inc()
+		} else {
+			metrics.Get().EC2ConnectionFailure.WithLabelValues(p.region).Inc()
+		}
 		p.unsetRequestInFlightForInstanceID(id)
 		return "", fmt.Errorf("failed querying private DNS from EC2 API for node %s: %s ", id, err.Error())
 	}
+	metrics.Get().EC2Responses.WithLabelValues("200", p.region).Inc()
```
CaidenBorrego: Is it okay to assume the response code will always be 200 at this point? Is there a way to dynamically check the response code of a successful message?

pkg/ec2provider/ec2provider.go
```diff
@@ -269,7 +291,13 @@ func (p *ec2ProviderImpl) getPrivateDNSAndPublishToCache(ctx context.Context, in
 	})
 	if err != nil {
 		logrus.Errorf("Batch call failed querying private DNS from EC2 API for nodes [%s] : with error = []%s ", instanceIDList, err.Error())
+		if code := ec2ResponseCode(err); code != "" {
+			metrics.Get().EC2Responses.WithLabelValues(code, p.region).Inc()
+		} else {
+			metrics.Get().EC2ConnectionFailure.WithLabelValues(p.region).Inc()
+		}
 	} else {
+		metrics.Get().EC2Responses.WithLabelValues("200", p.region).Inc()
```
CaidenBorrego: Same comment as above

pkg/ec2provider/ec2provider.go
```diff
@@ -207,9 +223,15 @@ func (p *ec2ProviderImpl) GetPrivateDNSName(ctx context.Context, id string) (str
 		InstanceIds: []string{id},
 	})
 	if err != nil {
+		if code := ec2ResponseCode(err); code != "" {
+			metrics.Get().EC2Responses.WithLabelValues(code, p.region).Inc()
```
CaidenBorrego, line 230: Just making sure we need region as a label? I thought region was automatically resolved by Cloudwatch. But if that's the convention of other metrics in this component then you can ignore this.

pkg/ec2provider/ec2provider.go
HanshalDabbiru: there is a way, but the response codes are gone from the output object by the time it reaches this part of the code. so getting it would need aws sdk middleware. the response code will be a 2xx always.

pkg/ec2provider/ec2provider.go
HanshalDabbiru: sts_total_responses has a region label so i followed that convention.

pkg/ec2provider/ec2provider.go
HanshalDabbiru: if there was no http response, the error is a http.ResponseError with a status code of 0.

pkg/ec2provider/ec2provider.go
CaidenBorrego: Okay, so I guess we are okay with generalizing all success responses to 200 then? If so then all good

pkg/ec2provider/ec2provider.go
CaidenBorrego: Okay then it is a pseudo-nil check. Sounds good

CaidenBorrego:
lgtm, a collaborator will need to add ok-to-test, lgtm, and approve

ronaldngounou (approved):

PR #1108, opened 2026-09-17
Raise maxTokenLenBytes to allow full-size STS session tokens

### LINKED ISSUE #1106: [Bug]: maxTokenLenBytes (4096) rejects valid STS session tokens above ~2.5KB
### What happened?

`Verify()` rejects any `k8s-aws-v1.` token larger than `maxTokenLenBytes` (4096 bytes) before decoding it (token.go L595-L597). The token embeds the caller's STS session token inside the presigned GetCallerIdentity URL (as `X-Amz-Security-Token`), which is then URL-encoded and base64url-encoded. That encoding inflates the session token's contribution by about 1.42x and adds fixed presigned-URL scaffolding, so a session token of only ~2.5KB produces an assembled bearer token that exceeds the 4096-byte cap.

Authentication then fails server-side with:

    level=warning msg="access denied" error="input token was not properly formatted: token is too large" method=POST path=/authenticate

Because the length check runs before identity resolution, the client sees an opaque `401 Unauthorized` with no indication that token size is the cause.

#### Why this is newly reachable

AWS STS recently moved to a single session-token size limit of 4096 bytes ([announcement](https://aws.amazon.com/blogs/security/aws-sts-simplifies-session-token-size-limits-and-adds-session-token-size-monitoring/)). Previously session tokens were small enough that the encoded bearer token stayed under `maxTokenLenBytes`. Now a valid session token can be large enough (via larger session policies, session tags, or growth over time) that the bearer token it produces exceeds the authenticator's 4096-byte cap. So a session token STS issues successfully can be rejected by the authenticator.

There are effectively two independent 4096-byte limits (STS session-token size, and this bearer-token cap). The encoding overhead between them means the usable session-token size is only ~2.5KB.

#### Encoding overhead

The bearer token is `k8s-aws-v1.` + base64url(presigned URL), and the session token sits inside that URL as a query parameter. Two multiplicative factors apply:

1. Percent-encoding of `+`, `/`, `=` in the base64 session token. Those are ~3.1% of a base64 alphabet at 3 bytes each, so ~1.0625x.
2. base64url of the whole URL, 4/3x.

Combined marginal rate ~1.42x, on top of ~380 bytes of fixed scaffolding (`https://sts...`, `Action`, `X-Amz-Algorithm`, `X-Amz-Credential`, `X-Amz-Date`, `X-Amz-Signature`, `X-Amz-SignedHeaders`, and the `k8s-aws-v1.` prefix):

    bearer_len ≈ 11 + (4/3) × (~380 + 1.0625 × session_token_len)

Solving for `bearer_len = 4096` gives a session-token ceiling of ~2.5KB.

#### Upper bound on any fix

`Verify()` replays the presigned URL as a GET request line to STS ([L685-L692](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go#L685-L692)), so the URL is also subject to the STS front end's request-URI limit. Probing `sts.us-west-2.amazonaws.com` with a synthetic oversized `X-Amz-Security-Token`:

    ~16.1 KB request URI -> HTTP 403 (parsed, auth failure as expected)
    ~16.2 KB request URI -> HTTP 400 (rejected before auth)
    20 KB and above      -> HTTP 400

The wall sits at ~16 KB, consistent with a 16384-byte request-line limit. This is measured, not documented. Since the bearer token is 4/3 of the URL, it bounds any useful value of `maxTokenLenBytes` at roughly 21 KB. Beyond that the authenticator would accept tokens that STS then rejects with an opaque 400, which is a worse failure mode than the current one. The usable range for a fix is therefore (4096, ~21500).

#### Steps to reproduce

1. Assume a role with `--minimum-session-token-size 3500` and use those credentials.
2. Run any kubectl command against the cluster.
3. Observe `401 Unauthorized`. Authenticator logs show `token is too large`.

Measured on a vanilla cluster in `us-west-2`: a session token near 2524 bytes is the last success, near 2528 bytes is the first failure. The exact crossover varies by a few bytes per call because STS compresses policies and tags, signature bytes differ, and the region name length affects the URL.

Note that the kubeconfig exec block must not carry `--role-arn`. If it does, `aws eks get-token` re-assumes the role from the injected credentials and mints a fresh, normal-sized token, discarding the inflated one, and the reproduction silently passes at every size. Inject assumed-role credentials directly instead.

#### Relevant code

- `maxTokenLenBytes = 1024 * 4`: [pkg/token/token.go#L101](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go#L101)
- length check in `Verify()`: [pkg/token/token.go#L595-L597](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go#L595-L597)
- `FormatError` prefix: [pkg/token/token.go#L133-L135](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/token/token.go#L133-L135)
- static deny response: [pkg/server/server.go#L52-L63](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/server/server.go#L52-L63)

### What you expected to happen?

A valid STS session token, which STS now issues at sizes up to 4096 bytes, should be usable for authentication. The bearer-token size cap should account for the encoding overhead of wrapping a session token in the presigned URL, so that a full-size session token does not overflow it.

**Suggested fix**: raise `maxTokenLenBytes` from `1024 * 4` (4096) to `1024 * 16` (16384), as a single constant. No other change to the validation path.

Cap to usable session token, using the measured 2524-byte crossover as the calibration anchor:

| cap          | largest session token that authenticates |
| ------------ | ---------------------------------------- |
| 4096 (today) | ~2.5 KB                                  |
| 10240        | ~6.9 KB                                  |
| 12288        | ~8.3 KB                                  |
| **16384**    | **~11.2 KB**                             |

Why not a smaller bump: a full-size 4096-byte session token already encodes to a ~6.3 KB bearer token, so the current cap fails well before the session token reaches its own limit. Per the STS announcement, 4096 is the current maximum and not a permanent ceiling (AWS cites future growth from new context keys, richer audit metadata, and post-quantum signatures, and advises against hard-coding it). If the limit later doubles to 8 KB, that produces a ~12.1 KB bearer token. A 12288 cap sits only ~160 bytes above that, under 1.5% margin, which is inside the per-call variance noted above. 16384 clears a hypothetical 8 KB session token with ~4 KB of margin.

Why not larger: 16384 yields a ~12.3 KB URL, safely inside the ~16 KB STS request-URI bound. 20480 would yield ~15.3 KB, uncomfortably close to it.

16384 is therefore the last clean power of two with real margin on both sides.

The failure mode is easy to misdiagnose because the client only sees `401 Unauthorized`, which looks like an RBAC or identity-mapping problem. The size cause is currently visible only in the authenticator's own logs.

A secondary improvement: populate `TokenReviewStatus.Error` on the deny path. Today `tokenReviewDenyJSON` ([pkg/server/server.go#L52-L63](https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/master/pkg/server/server.go#L52-L63)) is a static `{Authenticated: false}` with no `Error` field, so the reason is dropped at the webhook boundary. Kubernetes' webhook authenticator does convert a non-empty `Status.Error` into a Go error, so setting it would carry `token is too large` into kube-apiserver logs. It will not reach kubectl, since the apiserver returns a generic 401 either way. So this is a diagnosability fix, not a client-facing one. The genuinely client-visible improvement would be for `aws eks get-token` to warn when the token it just minted exceeds the cap, which is a separate change in the AWS CLI.

### Installation tooling

other (please specify in description)

### AWS IAM Authenticator server Version

Amazon EKS managed authenticator (version not exposed to customers)

### Client information

```Markdown
- OS/arch: macOS 26.6.2 (arm64)
- kubernetes client & version: Client Version: v1.37.0 and Kustomize Version: v5.8.1
- authenticator client & version: aws-cli/2.36.46 Python/3.14.6 Darwin/25.6.0 exe/arm64
```

### Kubernetes API Version

Client Version: v1.37.0 and Kustomize Version: v5.8.1

### aws-iam-authenticator YAML manifest

```YAML
N/A. Managed EKS control plane, manifests not customer-accessible.
```

### kube-apiserver YAML manifest

```YAML
N/A. Managed EKS control plane, manifests not customer-accessible.
```

### aws-iam-authenticator logs

```JSON
time="..." level=warning msg="access denied" client="127.0.0.1:xxxxx" error="input token was not properly formatted: token is too large" method=POST path=/authenticate
```

tanmaybaid:

`Verify()` rejects any token larger than `maxTokenLenBytes` (4096) before decoding it. The
caller's STS session token rides in the pre-signed `GetCallerIdentity` URL as
`X-Amz-Security-Token`, where percent-encoding of `+`, `/` and `=` (~1.0625x) and base64url
encoding (4/3) inflate it by roughly 1.42x on top of ~400 bytes of fixed pre-signed URL
scaffolding. That leaves only ~2.5KB of usable session token, so STS can issue a session
token well within its own 4096 byte limit that the authenticator then refuses.

Because the size check runs before identity resolution, the server logs `input token was not
properly formatted: token is too large` while the client sees only an opaque 401, which is
easy to misdiagnose as an RBAC or identity-mapping problem.

This raises the limit to 16KB, which admits ~11.2KB of session token. That leaves room for
STS raising its 4096 byte limit, while staying under the ~16KB request line STS itself
accepts for the pre-signed URL (probing `sts.us-west-2.amazonaws.com`, URIs up to ~16.1KB
reach authentication and return 403, while ~16.2KB and above return 400). The check remains
what it has always been: a cheap guard against decoding oversized input, rather than a
mirror of STS's current limit.

**Which issue(s) this PR fixes**:

Fixes #1106

**Testing**:

- `TestVerifyLargeSessionTokens` generates tokens through `generator.GetWithSTS`, the real
  presign and encode path, for a 4096 byte session token (STS's current limit) and an 8192
  byte one, asserting each fits under `maxTokenLenBytes` and verifies end to end. The
  resulting tokens are 6303 and 12106 bytes, and both cases fail against the old 4096 byte
  limit.
- `TestVerifyTokenPreSTSValidations` now also covers a token of exactly `maxTokenLenBytes`,
  which is not too large and so reaches the prefix check. The existing `maxTokenLenBytes+1`
  case still returns `token is too large`.
- `go test -race ./pkg/... ./cmd/...`, `golangci-lint run` and `make verify` pass locally.

/kind bug

```release-note
Raise the maximum accepted token size so full-size STS session tokens authenticate successfully instead of being rejected with a 401.
```


tanmaybaid:
This duplicates #1107, which was opened about an hour before it and fixes the same issue. I hadn't seen it when I pushed this.

Deferring to #1107 on the grounds that it came first. I've flagged the extra coverage from here over there (an 8192-byte session token case, which is what actually defends 16KiB over 12KiB, and a boundary case at exactly `maxTokenLenBytes`).

Happy to close this in favor of #1107, or to keep it open as an alternative if maintainers would rather take this diff — the production change here is a one-line constant bump, and the tests drive the token through `generator.GetWithSTS` rather than assembling the pre-signed URL by hand. Just say which and I'll act accordingly.

micahhausler (approved):
LGTM

